PR #30, opened 2020-04-17
Use shared_ptr in &Engine::AddWorld() fn

claireyywang:
I think it might be a better practice to create `shared_ptr` instead of creating local object.


iche033:
The changes look fine. Just curious, was this causing a problem? I think the original intention was that the `insert` call should make a copy of `world` so therefor it's safe for the local variable to go out of scope.

claireyywang:
> The changes look fine. Just curious, was this causing a problem? I think the original intention was that the `insert` call should make a copy of `world` so therefore it's safe for the local variable to go out of scope.

There wasn't any problem. I've just been reading about C++ and wondered if using shared_ptr would prevent any future issues. If it's good as is, I'd be fine closing this PR.

iche033:
ok sounds good. In that case, we could make `worlds` just store shared ptr? I saw that Entity's `children` is now using map of shared ptrs now so we could be consistent here

claireyywang:
@iche033 friendly ping for visibility. I addressed the previous comment. Ready for another round of review whenever works the best for you.

PR #31, opened 2020-04-17
Fix collision issue with detachable joints

azeey:
The issue occurs if a model is spawned and is attached to a parent model without a call to `step()` in between. This should be fixed in DART, but in the meantime, we can just call `incrementVersion` after `moveTo` is called on a DART BodyNode.

### DIFF
--- a/dartsim/src/JointFeatures.cc
+++ b/dartsim/src/JointFeatures.cc
@@ -161,6 +161,8 @@ void JointFeatures::DetachJoint(const Identity &_jointId)
   freeJoint->setSpatialVelocity(spatialVelocity,
           dart::dynamics::Frame::World(),
           dart::dynamics::Frame::World());
+  // TODO(addisu) Remove incrementVersion once DART has been updated to
+  // internally increment the BodyNode's version after moveTo.
   child->incrementVersion();
 }
 
@@ -201,6 +203,8 @@ Identity JointFeatures::AttachFixedJoint(
 
   const std::size_t jointID = this->AddJoint(
       bn->moveTo<dart::dynamics::WeldJoint>(parentBn, properties));
+  // TODO(addisu) Remove incrementVersion once DART has been updated to
+  // internally increment the BodyNode's version after moveTo.
   bn->incrementVersion();
   return this->GenerateIdentity(jointID, this->joints.at(jointID));
 }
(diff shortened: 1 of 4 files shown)


mxgrey:
Just to clarify, what's the issue that occurs? Does one of the models fall through the floor?

azeey:
Yes, the child model falls through the floor. It appears that the moved BodyNode and its descendants are not added to a CollisionGroup in DART which prevents any collision between the moved BodyNodes and other BodyNodes from being detected. Sorry, I should described the issue better.

dartsim/src/JointFeatures.cc
```diff
@@ -161,6 +161,7 @@ void JointFeatures::DetachJoint(const Identity &_jointId)
   freeJoint->setSpatialVelocity(spatialVelocity,
           dart::dynamics::Frame::World(),
           dart::dynamics::Frame::World());
+  child->incrementVersion();
```
nkoenig, line 164: Do you want to leave these calls in after DART has been patched? If not, maybe add a `/todo`?

dartsim/src/JointFeatures_TEST.cc
```diff
@@ -317,8 +321,17 @@ TEST_F(JointFeaturesFixture, JointAttachDetach)
...
   }
 
+  const auto poseParent = dartBody1->getTransform();
+  const auto poseChild = dartBody2->getTransform();
+  auto poseParentChild = poseParent.inverse() * poseChild;
+
   auto fixedJoint = model2Body->AttachFixedJoint(model1Body);
 
+  // AttachFixedJoint snaps the child body to the origin of the parent, so we
+  // set a transform ont he joint to keep the transform between the two bodies
```
nkoenig, line 331:
```suggestion
  // set a transform on the joint to keep the transform between the two bodies
```

dartsim/src/JointFeatures_TEST.cc
```diff
@@ -397,6 +418,134 @@ TEST_F(JointFeaturesFixture, LinkCountsInJointAttachDetach)
...
+  auto *dartBody1 = skeleton1->getBodyNode(bodyName1);
+  auto *dartBody2 = skeleton2->getBodyNode(bodyName2);
+
+  ASSERT_NE(nullptr, dartBody1);
+  ASSERT_NE(nullptr, dartBody2);
+
+  const auto poseParent = dartBody1->getTransform();
+  const auto poseChild = dartBody2->getTransform();
+
+  // Commenting out the following `step` call makes this test fail
```
nkoenig, line 516: Is this comment accurate? If so, should the next line be uncommented?

azeey: What I meant is that before this PR, commenting it out make it fail. Uncommenting it will make the test pass. I'll add a sentence to clarify.

azeey: 4458614

dartsim/src/JointFeatures.cc
azeey: I added a todo in 4458614

PR #32, opened 2020-04-17
[Citadel] Add Base and EntityManagement to tpeplugin

claireyywang:
This PR adds `EntityManagementFeatures` and `Base.hh` to lay the ground work of tpeplugin features. It should be merged after #45 which add functions to `tpelib` and before #46 which adds all the other features to `tpeplugin`.


tpe/plugin/src/Base.hh
```diff
@@ -0,0 +1,133 @@
...
+/// \brief The structs tpelib::ModelInfo,
+/// tpelib::LinkInfo, JointInfo, and ShapeInfo are used
+/// for two reasons:
+/// 1) Holding extra information such as the name or offset
+///    that will be different from the underlying engine
+/// 2) Wrap shared pointers to DART entities. Since these shared pointers (eg.
+///    dart::dynamics::BodyNodePtr) are different from std::shared_ptr, we
+///    cannot use them directly as parameters to GenerateIdentity. Instead we
+///    create a std::shared_ptr of the struct that wraps the corresponding DART
+///    shared pointer.
```
iche033, line 47: update doc

tpe/plugin/src/EntityManagementFeatures.cc
```diff
@@ -0,0 +1,410 @@
...
+    }
+  }
+  return this->GenerateInvalidId();
+}
+
+/////////////////////////////////////////////////
+const std::string &EntityManagementFeatures::GetLinkName(
+  const Identity &_linkID) const
+{
+  static std::string linkName = this->links.at(_linkID)->link->GetName();
```
iche033, line 224: does this need to be static?

tpe/plugin/src/EntityManagementFeatures.hh
```diff
@@ -0,0 +1,147 @@
...
+  // ----- Get entities -----
+  public: const std::string &GetEngineName(const Identity &) const override;
+
+  public: std::size_t GetEngineIndex(const Identity &) const override;
+
+  public: std::size_t GetWorldCount(const Identity &) const override;
+
+  public: Identity GetWorld(
+    const Identity &, std::size_t /*_worldIndex*/) const override;
+    // { return this->GenerateInvalidId(); };
```
iche033, line 59: remove commented out code?

tpe/plugin/src/EntityManagementFeatures.hh
```diff
@@ -0,0 +1,147 @@
...
+
+  public: Identity GetWorld(
+    const Identity &, const std::string &/*_worldName*/) const override;
+    // { return this->GenerateInvalidId(); };
```
iche033, line 63: remove?

tpe/plugin/src/EntityManagementFeatures.hh
```diff
@@ -0,0 +1,147 @@
...
+
+  public: const std::string &GetWorldName(
+    const Identity &_worldID) const override;
+
+  public: std::size_t GetWorldIndex(
+    const Identity &/*_worldID*/) const override;
+    // { return 0; };
```
iche033, line 70: remove?

tpe/plugin/src/EntityManagementFeatures.cc
```diff
@@ -0,0 +1,497 @@
...
+const std::string &EntityManagementFeatures::GetModelName(
+  const Identity &_modelID) const
+{
+  static std::string modelName {};
+  auto it = this->models.find(_modelID);
+  if (it != this->models.end() && it->second != nullptr)
+  {
+    modelName = it->second->model->GetName();
+  }
+  return modelName;
```
iche033, line 174: I think we could use `ReferenceInterface` in this function to avoid lookups?

```
static std::string modelName =
    this->ReferenceInterface<ModelInfo>(_modelID)->model->GetName();
return modelName;
```

I now understand that `static std::string` is used because of the `const &` return type.  I think it's designed to match the return type of dart's [Skeleton::getName](https://dartsim.github.io/api/v5.1.4/d3/d19/classdart_1_1dynamics_1_1Skeleton.html#a583867ed1a17d3f2397562d68b860a7b) function

tpe/plugin/src/EntityManagementFeatures.cc
```diff
@@ -0,0 +1,497 @@
...
+std::size_t EntityManagementFeatures::GetLinkCount(
+  const Identity &_modelID) const
+{
+  auto it = this->models.find(_modelID);
+  if (it != this->models.end() && it->second != nullptr)
+  {
+    return it->second->model->GetChildCount();
+  }
+  // return invalid count if model not found
+  return -1;
```
iche033, line 217: this also seems to work:

```
return this->ReferenceInterface<ModelInfo>(_modelID)->model->GetChildCount();
```

claireyywang:
> thanks for adding the extra tests!
> 
> I noticed in the dartsim implementation, it makes use of `ReferenceInterface` function for fast retrieval of entities. I gave that a try in a couple of places and they seem to work fine. Should we update other functions?

Thanks for the review, @iche033 ! I have updated functions to use `ReferenceInterface` and that simplifies the code a lot. I have also cleaned up the code a bit more, so ready for another round whenever you are :smiley:

claireyywang:
> looks good to me!
> 
> A side note, I noticed that code check is not running on `tpe` src code. We can add the check [here](https://github.com/ignitionrobotics/ign-physics/blob/master/tools/code_check.sh#L48). After that, running `sh tools/code_check.sh` should pick up style issues. We can do that and fix the errors in a separate PR

Awesome, thanks for catching the code check! I didn't know it needs to be manually added. I'll follow up with another PR fixing styles as suggested.

PR #33, opened 2020-04-17
[ign-physics1] Update BitBucket links

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,35 +20,35 @@ designed to support simulation and rapid development of robot applications.
 
 # Table of Contents
 
-[Motivation](#markdown-header-motivation)
+[Motivation](#motivation)
 
-[Features](#markdown-header-features)
+[Features](#features)
 
-[Install](#markdown-header-install)
+[Install](#install)
 
-* [Binary Install](#markdown-header-binary-install)
+* [Binary Install](#binary-install)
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -157,7 +157,7 @@ sudo update-alternatives --install \
 
 # Usage
 
-Please refer to the [examples directory](https://bitbucket.org/ignitionrobotics/ign-physics/raw/default/examples/?at=default).
+Please refer to the [examples directory](https://github.com/ignitionrobotics/ign-physics/raw/master/examples/).
```
scpeters, line 160: this link is broken because the examples folder only exists on `ign-physics2`; we need to merge it forward to master

scpeters:
> some table of contents readme links are broken due to case-sensitivity

it wasn't actually case-sensitivity, but it's fixed in 154b18e

README.md
chapulina: I removed the usage section for `ign-physics1` because it doesn't have any examples. I think we should add it if we ever add examples to this version. This was probably here only because it's part of the template we used for all libs. Let me know what you think.

7573e5eab805a3d0bff52072733e8b073f93ac5d

PR #41, opened 2020-04-20
Reduce the symbol load caused by feature templates

### LINKED ISSUE #28: Reopen pull request to reduce symbol load
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/issues/28)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
Reopen [pull request #136](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/pull-requests/136) once we're migrated to GitHub.

chapulina:
Reopening the pull request with @mxgrey 's work.

Resolves #28.

---

### Original description

This PR should greatly reduce the size and quantity of symbols that get generated by templates, likely by some orders of magnitude. It should also help make the size and quantity of symbols scale much much better (should be nearly constant [size] and linear [quantity] respectively now instead of linear [size] and squared [quantity]).

I believe this has been the leading cause of slow compilation times, especially for GCC.

I made a simple program [here](https://github.com/mxgrey/test_ign_physics) that instantiates every entity type with every existing feature in ign-physics, and using all headers provided by ign-physics. Before this PR, using gcc-8 it spent over an hour trying to compile on my XPS15 with no success. After this PR, using gcc-8 it compiles in just under 30 seconds. Clang is significantly better, compiling in 11 seconds after this PR.

There is still some cleanup to do, but I wanted to open this PR so the compilation times of ign-gazebo can be tested against it.

### DIFF
--- a/include/ignition/physics/detail/FeatureList.hh
+++ b/include/ignition/physics/detail/FeatureList.hh
@@ -264,12 +264,21 @@ namespace ignition
       template <typename InputTuple>
       struct RemoveTupleRedundancies;
 
+      template <typename T>
+      struct wrap { };
+
+      template <typename Tuple>
+      struct unwrap;
+
+      template <typename... T>
+      struct unwrap<std::tuple<wrap<T>...>>
+      {
+        using type = std::tuple<T...>;
+      };
+
       template <typename... InputTupleArgs>
       struct RemoveTupleRedundancies<std::tuple<InputTupleArgs...>>
       {
-        template <typename T>
-        struct wrap { };
-
         template <typename PartialResultInput, typename...>
         struct Impl;
 
@@ -295,15 +304,6 @@ namespace ignition
           using type = decltype(std::tuple_cat(
               PartialResult(),
               typename Impl<AggregateResult, Others...>::type()));
-        };
-
-        template <typename Tuple>
-        struct unwrap;
-
-        template <typename... T>
-        struct unwrap<std::tuple<wrap<T>...>>
-        {
-          using type = std::tuple<T...>;
         };
 
         using type =
--- a/include/ignition/physics/detail/FeatureList.hh
+++ b/include/ignition/physics/detail/FeatureList.hh
@@ -326,7 +326,8 @@ namespace ignition
 
       /////////////////////////////////////////////////
       template <typename FeatureListT>
-      struct FlattenFeatures<FeatureListT, void_t<typename FeatureListT::FeatureTuple>>
+      struct FlattenFeatures<
+          FeatureListT, void_t<typename FeatureListT::FeatureTuple>>
       {
         using type =
             typename FlattenFeatures<typename FeatureListT::FeatureTuple>::type;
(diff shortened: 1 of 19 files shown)


scpeters:
I don't understand the template stuff. I don't know if @azeey has a better handle on that.

It would be nice to add the example program from https://github.com/mxgrey/test_ign_physics to either the `examples` folder or as a performance test

mxgrey:
We could add that as a test to this repo, although the point of it is to make the most excruciating compilation time possible. I don't think it would be a good idea to add it to the `all` target or even a `tests` target. If it gets added, it should be its own standalone target that nothing else depends on.

include/ignition/physics/detail/FeatureList.hh
```diff
@@ -367,9 +502,50 @@ namespace ignition
       template <template<typename> class Selector, typename FeatureListT>
       struct ExtractAPI
       {
-        public: template<typename... T>
-        class type : public virtual
-            Aggregate<Selector, FeatureListT>::template type<T...> { };
+        template <typename FeatureTuple, typename... T>
```
azeey, line 505: Can the `Aggregate` template be removed now that `ExtractAPI` doesn't use it?

chapulina: Thanks for the review, @azeey ! I'll go ahead and merge this PR to unblock `ign-gazebo`, but @mxgrey , feel free to follow up in a subsequent PR if needed.

azeey:
Can this be backported to ign-physics1? It looks like the compile times there are also getting worse.

scpeters:
> Can this be backported to ign-physics1? It looks like the compile times there are also getting worse.

I'm willing to review a pull request; I'm not sure if it will cause a problem with API/ABI

include/ignition/physics/detail/FeatureList.hh
mxgrey: > From what I can tell, this still checks for duplicate Features, but I remember you mentioned that in order to reduce compile times, we have to assume that there are no duplicate Features in a FeatureList. Is that still the case?

That is not the case, fortunately! There should be no problem at all with having duplicate features.

> Can ComposePlugin be updated to work like the new ExtractAPI? Would that save any more compile time?

I considered that, but I didn't have enough time to really explore it. When I inspect the symbols that get generated, there are relatively few symbols related to `ComposePlugin`, most likely because all entity types share the same plugin type, so cutting down on `ComposePlugin` probably won't offer much. It's certainly worth trying if someone has cycles to spare.

> Can the Aggregate template be removed now that ExtractAPI doesn't use it?

For some reason I thought `Aggregate` was still being used somewhere, but I can no longer find wherever I thought it was being used. So sure, I guess we can throw it out now.

PR #42, opened 2020-04-27
Restore detached BodyNodes to original skeleton

azeey:
The is was originally opened by @scpeters before the Github migration ([archived PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/pull-requests/104)). I have modified it to store the original names of links and using those names when moving BodyNodes between skeletons.

### DIFF
--- a/dartsim/src/Base.hh
+++ b/dartsim/src/Base.hh
@@ -56,6 +56,10 @@ struct ModelInfo
 struct LinkInfo
 {
   dart::dynamics::BodyNodePtr link;
+  /// \brief It may be necessary for dartsim to rename a BodyNode (eg. when
+  /// moving the BodyNode to a new skeleton), so we store the Gazebo-specified
+  /// name of the Link here.
+  std::string name;
 };
 
 struct JointInfo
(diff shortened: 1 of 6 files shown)


dartsim/src/Base.hh
```diff
@@ -56,6 +56,10 @@ struct ModelInfo
 struct LinkInfo
 {
   dart::dynamics::BodyNodePtr link;
+  /// \brief It maybe necessary for dartsim to rename a BodyNode (eg. when
```
scpeters, line 59: nit: `maybe` -> `may be`

dartsim/src/JointFeatures.cc
```diff
@@ -156,7 +156,56 @@ void JointFeatures::DetachJoint(const Identity &_jointId)
...
+
+  auto childLinkInfo = this->links.at(child);
+
+  dart::dynamics::SkeletonPtr skeleton;
+  {
+    // Find the original skeleton the child BodyNode belonged to
+    std::string oldName = child->getName();
+    if (oldName != childLinkInfo->name)
+    {
+      std::size_t originalNameIndex = oldName.find(childLinkInfo->name);
```
scpeters, line 173: I think we should use an `rfind` so we don't accidentally find the skeleton's name?

azeey: Good point. I added a failing test that shows what happens when a model and link have the same name in 53314df and applied your suggestion in 7a9e269, which fixes the test.

dartsim/src/Base.hh
azeey: 551c18a

PR #45, opened 2020-04-29
Add tpelib::Shape::GetSize() and tpelib::Model::GetCanonicalLink()

claireyywang:
`tpelib::BoxShape::GetSize()` was declared but not implemented in previous PR and caused windows CI failure in `tpe_plugin` PR

`tpelib::Model::GetCanonicalLink()` assumes the first link of each model is the canonical link, thus enabling ign-gazebo to find the link entity https://github.com/ignitionrobotics/ign-gazebo/blob/2259f027cd3a244f2c841b62c968088eace1b338/src/systems/physics/Physics.cc#L1233 and update world pose accordingly. 

This PR needs to be merged before #32

### DIFF
--- a/tpe/lib/src/Model.hh
+++ b/tpe/lib/src/Model.hh
@@ -47,6 +47,8 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Model : public Entity
   /// \return Newly created Link
   public: Entity &AddLink();
 
+  /// \brief Get the canonical link of model
+  /// \return Entity the canonical (first) link
   public: Entity &GetCanonicalLink();
 
   /// \brief Set the linear velocity of model
--- a/tpe/lib/src/Model.hh
+++ b/tpe/lib/src/Model.hh
@@ -47,6 +47,8 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Model : public Entity
   /// \return Newly created Link
   public: Entity &AddLink();
 
+  public: Entity &GetCanonicalLink();
+
   /// \brief Set the linear velocity of model
   /// \param[in] _velocity linear velocity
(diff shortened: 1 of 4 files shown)


tpe/lib/src/Model.hh
```diff
@@ -47,6 +47,8 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Model : public Entity
   /// \return Newly created Link
   public: Entity &AddLink();
 
+  public: Entity &GetCanonicalLink();
```
iche033, line 50: can you add doxygen comment for this function?

tpe/lib/src/Shape.cc
```diff
@@ -84,6 +84,12 @@ void BoxShape::SetSize(const math::Vector3d &_size)
   this->dirty = true;
 }
 
+//////////////////////////////////////////////////
+math::Vector3d BoxShape::GetSize()
```
iche033, line 88: can you add a simple check [here](https://github.com/ignitionrobotics/ign-physics/blob/claire/tpelib_add_fns/tpe/lib/src/Shape_TEST.cc#L39) in Shape_TEST.cc to test this function?

tpe/lib/src/Model.hh
claireyywang: added https://github.com/ignitionrobotics/ign-physics/pull/45/files#diff-a2e564da4da991a87f531e647b213a9cR50-R52

tpe/lib/src/Shape.cc
claireyywang: added https://github.com/ignitionrobotics/ign-physics/pull/45/files#diff-b6ac34fb28f6fcdad5799d1f97124481R40

PR #59, opened 2020-05-29
[Citadel] [TPE] Add function to get an Entity's bounding box

iche033:
Add helper function that returns an entity's bounding box by recursively merging it's children's bounding boxes.

### DIFF
--- a/tpe/lib/src/Utils.cc
+++ b/tpe/lib/src/Utils.cc
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


tpe/lib/src/Utils.cc
```diff
@@ -0,0 +1,90 @@
...
+*/
+
+#include "Utils.hh"
+
+namespace ignition {
+namespace physics {
+namespace tpelib {
+
+//////////////////////////////////////////////////
+math::AxisAlignedBox transformAxisAlignedBox(
```
claireyywang, line 25: Could we add more documentation on the math behind this function? I find it a bit hard to understand by just reading the code.

iche033: added more doc and testing linker fix in 558cbb7

iche033:
all checks passed, merging!

PR #60, opened 2020-05-29
[Citadel] [TPE] Add collision detector

iche033:
depends on pull request #59 

Add collision detector in TPE that checks collisions between pairs of models by doing axis aligned bounding box intersection tests.

If intersection occurs, the collision detector can either return 
  * 8 contact points associated with the corners of the intersecting region, or
  * 1 contact point at the center of intersecting region

This can be specified by setting the `_singleContact` bool variable when calling `CollisionDetector::CheckCollisions` function


iche033:
suppressed dll warnings in b5e4700. Waiting for one more round of CI

iche033:
all checks passed, merging

PR #61, opened 2020-05-29
[Citadel] [TPE] Add GetContactsFromLastStepFeature

iche033:
depends on pull request #59  and pull request #60 

Add TPE implementation of `GetContactsFromLastStepFeature`. It returns contact points generated by the CollisionDetector introduced in pull request #60 

Since contacts are associated with collisions in ign-physics but TPE checks contacts on the model level, a workaround is implemented to return the first collision in the canonical link of the model. 

The test added caught a bug in EntityManagementFeatures in which incorrect entity names were returned so I had to remove the static variables.

### DIFF
--- a/tools/code_check.sh
+++ b/tools/code_check.sh
@@ -45,7 +45,7 @@ then
   CPPLINT_FILES="$CHECK_FILES"
   QUICK_TMP=`mktemp -t asdfXXXXXXXXXX`
 else
-  CHECK_DIRS="./src ./include ./test/integration ./test/regression ./test/performance"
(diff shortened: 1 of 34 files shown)


iche033:
I just rebuilt ign-physics and ign-gazebo in two colcon worksapces and I can't reproduce the unresolved symbol error. I do see the component warning msgs. 

Could you try clean the build dir and rebuild to see if the problem still occurs?

claireyywang:
> I just rebuilt ign-physics and ign-gazebo in two colcon worksapces and I can't reproduce the unresolved symbol error. I do see the component warning msgs.
> 
> Could you try clean the build dir and rebuild to see if the problem still occurs?

I did a clean build and the error went away. Thanks for double checking!

PR #69, opened 2020-06-23
[Citadel][TPE] Implement collision filtering using collide bitmasks and add CollisionFilterMaskFeature

iche033:
Added collide bitmask to Collisions in tpelib for doing collision filtering. 

Since collision detection in TPE is done between models (not collisions), I also added a function to Entity to get a single combined (bitwise OR'd) collide bitmask from all child collisions. e.g. if one collision within model A can collide with any one of the collisions in model B then the two models should collide. 

Also added the CollisionFilterMaskFeature similar to the way it was done for dart in this [pull reqeuest](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/pull-requests/116/page/1)

### DIFF
--- a/tpe/lib/src/CollisionDetector.cc
+++ b/tpe/lib/src/CollisionDetector.cc
@@ -15,6 +15,8 @@
  *
 */
 
+#include <unordered_map>
+
 #include "CollisionDetector.hh"
 #include "Utils.hh"
 
@@ -31,23 +33,15 @@ std::vector<Contact> CollisionDetector::CheckCollisions(
   // contacts to be filled and returned
   std::vector<Contact> contacts;
 
-  // cache of collide bitmasks
-  std::map<std::size_t, uint16_t> collideBitmasks;
-
   // cache of axis aligned box in world frame
-  std::map<std::size_t, math::AxisAlignedBox> worldAabb;
+  std::unordered_map<std::size_t, math::AxisAlignedBox> worldAabb;
 
   for (auto it = _entities.begin(); it != _entities.end(); ++it)
   {
     std::shared_ptr<Entity> e1 = it->second;
 
     // Get collide bitmask for enitty 1
-    uint16_t cb1 = 0xFF;
-    auto cb1It = collideBitmasks.find(e1->GetId());
-    if (cb1It == collideBitmasks.end())
-      collideBitmasks[e1->GetId()] = e1->GetCollideBitmask();
-    else
-      cb1 = cb1It->second;
+    uint16_t cb1 = e1->GetCollideBitmask();
 
     // Get world axis aligned box for entity 1
     math::AxisAlignedBox wb1;
@@ -71,12 +65,7 @@ std::vector<Contact> CollisionDetector::CheckCollisions(
       std::shared_ptr<Entity> e2 = it2->second;
 
       // Get collide bitmask for entity 2
-      uint16_t cb2 = 0xFF;
-      auto cb2It = collideBitmasks.find(e2->GetId());
-      if (cb2It == collideBitmasks.end())
-        collideBitmasks[e2->GetId()] = e2->GetCollideBitmask();
-      else
-        cb2 = cb2It->second;
+      uint16_t cb2 = e2->GetCollideBitmask();
 
       // collision filtering using collide bitmask
       if ((cb1 & cb2) == 0)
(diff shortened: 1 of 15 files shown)


tpe/lib/src/CollisionDetector.cc
```diff
@@ -31,14 +31,26 @@ std::vector<Contact> CollisionDetector::CheckCollisions(
   // contacts to be filled and returned
   std::vector<Contact> contacts;
 
+  // cache of collide bitmasks
+  std::map<std::size_t, uint16_t> collideBitmasks;
+
```
luca-della-vedova, line 36: A few comments about this.
First of all you can probably see coming my "usual" feedback about using hash maps, since we are also doing quite a lot of operations on this map (number of models in the simulation squared).
But I am also not too sure about using a cache in the first place, from what I can see we are trying to avoid the overhead of iterating over the children in the `Entity::GetCollideBitmask()` function, what do you think about calculating / caching the entity bitmask in the Entity object in the first place?

Also small comment in the worldAabb map, if it's not out of the scope of the PR I would still suggest changing to a hash map to save a logn binary search in every find() and insert operation. All the function calls should be unchanged since the API is quite similar.

tpe/plugin/src/SDFFeatures.cc
```diff
@@ -168,5 +168,27 @@ Identity SDFFeatures::ConstructSdfCollision(
...
+      {
+        elem = elem->GetElement("contact");
+        if (elem->HasElement("collide_bitmask"))
+        {
+          collideBitmask = elem->Get<unsigned int>("collide_bitmask");
+          this->SetCollisionFilterMask(collisionIdentity, collideBitmask);
+        }
+      }
+    }
+  }
```
luca-della-vedova, line 191: Just a note on this (you are probably aware of it already), the original implementation in the dart SDFFeatures was done before the availability of optional features and the ign-gazebo PR that takes care of this logic using the [SDF Surface DOM](https://github.com/ignitionrobotics/ign-gazebo/pull/160/files). Admittedly when I did that PR I didn't double check that it could be compatible with Citadel, if that's the case we could probably backport it and get rid of this logic?

tpe/lib/src/CollisionDetector.cc
iche033: ok yeah I think the cache for collide bitmask is probably premature optimization so I just removed it. The `GetCollideBitmask` function is just doing bitwise OR operations which should be fast. Updated `worldAabb` to use `unordered_map`. 1555caa

tpe/plugin/src/SDFFeatures.cc
iche033: I backported the collide bitmask changes to citadel in ignitionrobotics/ign-gazebo#221

I just noticed that if the logic here is removed, ign-physics alone won't be able to consume the collide bitmasks flags from sdf as it will now rely on ign-gazebo to set these flags, This causes ` SimulationFeature_TEST.CollideBitmasks` to fail unless we write extra logic in the integration test to set them. What do you think about leaving the logic here for ign-physics users that do not use ign-gazebo? I think it'll just be overriden by ign-gazebo as it parses SDF surface DOMs?

tpe/lib/src/CollisionDetector.cc
luca-della-vedova: I'm still quite interested in the idea of caching the collide bitmask at the entity level (i.e. whenever a "child" is added / removed recalculate the bitmask and make the Get function be a simple property getter), do you think that would be a good approach? Otherwise this looks good to me!

tpe/plugin/src/SDFFeatures.cc
luca-della-vedova: Admittedly I don't know much about use cases for ignition libraries (such as physics, or rendering) without ignition gazebo, but no strong feelings here we can leave it if it helps with testing.

tpe/lib/src/CollisionDetector.cc
iche033: I tested the caching approach and ended up adding more code than I had planned so I created a separate pull request for it, #73. Feedback welcome.

tpe/plugin/src/SDFFeatures.cc
iche033: ok yeah let's leave it in there for testing purposes.

tpe/lib/src/CollisionDetector.cc
iche033: ok merged #73 into this PR

iche033:
fixed windows build in 82578e9

PR #73, opened 2020-06-25
[Citadel][TPE] Cache Entity data

iche033:
Added flags to keep track of entity changes so that we can return cached data like collide bitmasks and bounding box if no changes are detected.

While making these changes, I noticed there was already a flag to keep track of bounding box changes (`bboxDirty`) but it was never set to true so bounding boxes were never recomputed.

I added some logic and public functions to to help propagate these flags up the tree (`Set|GetParent`, `ChildrenChanged`). Not ideal so feedback welcome. 

Signed-off-by: Ian Chen <ichen@osrfoundation.org>

### DIFF
--- a/tpe/lib/src/Model_TEST.cc
+++ b/tpe/lib/src/Model_TEST.cc
@@ -133,11 +133,11 @@ TEST(Model, BoundingBox)
 
   math::AxisAlignedBox expectedBoxLinkFrame(
       math::Vector3d(-2, -2, -2), math::Vector3d(2, 2, 2));
-  EXPECT_EQ(expectedBoxLinkFrame, linkEnt.GetBoundingBox(true));
+  EXPECT_EQ(expectedBoxLinkFrame, linkEnt.GetBoundingBox());
 
   math::AxisAlignedBox expectedBoxModelFrame(
       math::Vector3d(-2, -2, -1), math::Vector3d(2, 2, 3));
-  EXPECT_EQ(expectedBoxModelFrame, model.GetBoundingBox(true));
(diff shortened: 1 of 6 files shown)


tpe/lib/src/Model_TEST.cc
```diff
@@ -133,11 +133,11 @@ TEST(Model, BoundingBox)
 
   math::AxisAlignedBox expectedBoxLinkFrame(
       math::Vector3d(-2, -2, -2), math::Vector3d(2, 2, 2));
-  EXPECT_EQ(expectedBoxLinkFrame, linkEnt.GetBoundingBox(true));
+  EXPECT_EQ(expectedBoxLinkFrame, linkEnt.GetBoundingBox());
```
luca-della-vedova, line 136: Oh interesting, so basically since `_force` was being set to true before the dirty flag was actually not being tested?
Can you add a few lines to set bitmasks for the Collision objects here and test the OR-ing / `collideBitmaskDirty` logic as well?

iche033: yeah that's correct, removing the force arg revealed that bounding box caching was not working. 

there is already a CollideBitmask test added in pull request #69 ([here](https://github.com/ignitionrobotics/ign-physics/pull/69/files#diff-8ab09be8e30cbdbf937d238f22285466R165)) and I've been using that to test this PR. It should cover `collideBitmaskDirty` logic.

luca-della-vedova: Ok saw it, looks great!

PR #40, opened 2020-04-18
Extend contact data with force, normal, and penetration depth

diegoferigo:
This PR provides the support of exporting optional contact data that the physics engine might provide. In particular:

1. Contact force
1. Contact normal
1. Penetration depth

This PR also extends the DART implementation to fill the new data structure containing the additional contact data.

I'm not yet very familiar with all the templates of ign-physics. This is the first working implementation I managed to obtain, I'm not sure if there's a better way to handle the data exchanged between the public APIs and the implementation method.

In the current form, ABI is likely broken. Given the field `CompositeData extraData`, I'm not sure if there's any other way to get that data without the need to define a new public struct.

Closes https://github.com/ignitionrobotics/ign-physics/issues/24.

PR #77, opened 2020-07-01
Fix copying SpecifyData objects

azeey:
This fixes a crash when copying `SpecifyData` objects and implements a copy constructor for `ExpectData` so that copied objects don't share the same underlying iterator to `DataMap`. 

This was causing unexpected test failures in #40.

- https://github.com/ignitionrobotics/ign-physics/commit/3b480dda6e21b63cf33fa52b8aeef2c231c5aa1b adds a test that segfaults because a `SpecifyData` is copied. 
- https://github.com/ignitionrobotics/ign-physics/commit/51cfb62388e959b9d7926e753c2cf21f9f3262cb fixes the segfault, but the test fails because the copied objects share an iterator.
- https://github.com/ignitionrobotics/ign-physics/commit/aa2c61de4f9be9e8374e500724719689606a2614 Fixes the test by implementing a copy constructor and copy assignment operator for `ExpectData`. Note that implementing these for `RequiredData` is not necessary because `RequiredData` does not have any members that need special handing.

I left a TODO for implementing a move constructor for `SpecifyData` because right now, trying to move a `SpecifyData` object results in a copy. I looked into adding the constructor and move assignment operator, but since `SpecifyData` inherits from all of its template parameters, there is potential to move from an already moved from object. Here's the compiler warning I got while I was attempting to add defaulted move constructors and assignment operators to `SpecifyData`.

```
warning: defaulted move assignment for ‘ignition::physics::SpecifyData<ignition::physics::SpecifyData<ignition::physics::RequireData<StringData, BoolData, CharData>, ignition::physics::ExpectData<IntData, FloatData> >, ignition::physics::RequireData<StringData> >’ calls a non-trivial move assignment operator for virtual base ‘ignition::physics::SpecifyData<ignition::physics::RequireData<StringData, BoolData, CharData>, ignition::physics::ExpectData<IntData, FloatData> >’ [-Wvirtual-move-assign]
```
/cc @mxgrey, @diegoferigo

### DIFF
--- a/include/ignition/physics/detail/SpecifyData.hh
+++ b/include/ignition/physics/detail/SpecifyData.hh
@@ -41,6 +41,23 @@ namespace ignition
 
     /////////////////////////////////////////////////
     template <typename Expected>
+    ExpectData<Expected>::ExpectData(const ExpectData<Expected> &_other)
+        : ExpectData()
+    {
+      this->Copy(_other);
+    }
+
+    /////////////////////////////////////////////////
+    template <typename Expected>
+    ExpectData<Expected> &ExpectData<Expected>::operator=(
+        const ExpectData<Expected> &_other)
+    {
+      this->Copy(_other);
+      return *this;
+    }
+
+    /////////////////////////////////////////////////
+    template <typename Expected>
     template <typename Data>
     Data &ExpectData<Expected>::Get()
     {
--- a/include/ignition/physics/detail/SpecifyData.hh
+++ b/include/ignition/physics/detail/SpecifyData.hh
@@ -41,19 +41,10 @@ namespace ignition
 
     /////////////////////////////////////////////////
     template <typename Expected>
-    ExpectData<Expected>::ExpectData(const ExpectData<Expected> &_other)
+    ExpectData<Expected>::ExpectData(const ExpectData<Expected> &)
         : ExpectData()
     {
-      this->Copy(_other);
-    }
-
-    /////////////////////////////////////////////////
-    template <typename Expected>
-    ExpectData<Expected> &ExpectData<Expected>::operator=(
-        const ExpectData<Expected> &_other)
-    {
-      this->Copy(_other);
-      return *this;
+      // Do nothing
     }
 
     /////////////////////////////////////////////////
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ### Ignition Physics 2.x.x (20XX-XX-XX)
 
+1. Fix copying of SpecifyData objects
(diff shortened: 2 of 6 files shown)


include/ignition/physics/detail/SpecifyData.hh
```diff
@@ -39,6 +39,23 @@ namespace ignition
       // Do nothing
     }
 
+    /////////////////////////////////////////////////
+    template <typename Expected>
+    ExpectData<Expected>::ExpectData(const ExpectData<Expected> &_other)
+        : ExpectData()
+    {
+      this->Copy(_other);
```
mxgrey, line 47: It's been a long time since I've looked at this code, but I don't think this is the "right" thing to do here, even if it gives the correct outcome. If I remember how this all works, this will cause all of the data in `CompositeData` to be re-copied over and over again, once for each data type `T` that was given to an `ExpectData<T>`.

I think we can just leave out this line. I believe delegating to the default constructor should be enough to fix the problem described by the PR.

include/ignition/physics/detail/SpecifyData.hh
```diff
@@ -39,6 +39,23 @@ namespace ignition
...
+    }
+
+    /////////////////////////////////////////////////
+    template <typename Expected>
+    ExpectData<Expected> &ExpectData<Expected>::operator=(
+        const ExpectData<Expected> &_other)
+    {
+      this->Copy(_other);
```
mxgrey, line 55: Similar to the previous comment, I think doing this will re-copy all the data of the `CompositeData` many times.

As far as I can figure, an explicit copy assignment operator should not be needed. If the compiler doesn't produce one automatically (I can't remember the rules about that), then we should be able to simply do

```
ExpectedData &operator=(const ExpectedData &_other) = default;
```

in the class definition of `ExpectedData`.

include/ignition/physics/detail/SpecifyData.hh
azeey: Yeah, you're right. The copy constructor of `CompositeData` is called before this and copies the `MapData` object so all that is left to do here is point the `PrivateSpecifyData::expectedIterator` to the new map. https://github.com/ignitionrobotics/ign-physics/pull/77/commits/ec3947b318120951011ec1982c1c81f8abd63ffd

include/ignition/physics/detail/SpecifyData.hh
azeey: Agreed. I had to add an empty copy assignment operator to `PrivateSpecifyData` for this to work because the compiler doesn't generate one due to `PrivateSpecifyData::expectedIterator` being a const. https://github.com/ignitionrobotics/ign-physics/pull/77/commits/ec3947b318120951011ec1982c1c81f8abd63ffd

Issue #57, opened 2020-05-28
Add link offset to tpeplugin::KinematicsFeatures


claireyywang:
Current implementation does not consider link position.  Setting model pos also sets every entity within  the model. We will need to consider the offset when there is a <pose> tag in <link>, and adjust framedata accordingly. More info refer to discussion here https://github.com/ignitionrobotics/ign-physics/pull/46#discussion_r429395878

PR #79, opened 2020-07-09
Add offset to link and collision pose

claireyywang:
Use `pose` represent relative transformation to parent entity. `GetWorldPose` represents pose relative to world. This PR takes account of link and collision relative pose to compute their worldPose.  

This is the first step to implement #57

### DIFF
--- a/tpe/lib/src/Collision.hh
+++ b/tpe/lib/src/Collision.hh
@@ -72,18 +72,6 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Collision : public Entity
   // Documentation inherited
   public: math::AxisAlignedBox GetBoundingBox(bool _force) override;
 
-  /// \brief Update the pose of the entity
-  /// \param[in] _linkPose parent link pose
-  public: void UpdatePose(const math::Pose3d _linkPose);
-
-  /// \brief Get the relative transform to its parent link
-  /// \return relative transform
-  public: const Eigen::Isometry3d GetTf();
-
-  /// \brief Set the relative transform to its parent link
-  /// \param[in] _tf relative transform
-  public: void SetTf(const Eigen::Isometry3d _tf);
-
   /// \brief Private data pointer class
   private: CollisionPrivate *dataPtr = nullptr;
 };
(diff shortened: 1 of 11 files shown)


tpe/lib/src/Link.hh
```diff
@@ -42,6 +44,22 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Link : public Entity
...
+  /// \brief Get the relative transform to its parent model
+  /// \return relative transform
+  public: const Eigen::Isometry3d GetTf();
+
+  /// \brief Set the relative transform to its parent model
+  /// \param[in] _tf relative transform
+  public: void SetTf(const Eigen::Isometry3d _tf);
+
+  /// \brief Private relative transform to its parent entity
+  public: Eigen::Isometry3d tf =
```
scpeters, line 61: according to the comment, should this be `private:`?

iche033:
The changes look good to me. Is this PR still a draft?

claireyywang:
@iche033 awesome, thanks for looking into it. I think this PR is ready to go. I have updates in tpeplugin but that is on a separate branch and doesn't interfere with this one.

tpe/lib/src/Collision.hh
```diff
@@ -18,7 +18,7 @@
 #ifndef IGNITION_PHYSICS_TPE_LIB_SRC_COLLISION_HH_
 #define IGNITION_PHYSICS_TPE_LIB_SRC_COLLISION_HH_
 
-#include <ignition/math/AxisAlignedBox.hh>
+#include <ignition/math/eigen3/Conversions.hh>
```
iche033, line 21: I think we don't need this include now right?

tpe/lib/src/Link.hh
```diff
@@ -18,6 +18,8 @@
 #ifndef IGNITION_PHYSICS_TPE_LIB_SRC_LINK_HH_
 #define IGNITION_PHYSICS_TPE_LIB_SRC_LINK_HH_
 
+#include <ignition/math/eigen3/Conversions.hh>
```
iche033, line 21: no longer needed?

PR #81, opened 2020-07-14
Tweaks to support physics plugin loading in gazebo

mjcarroll:

### DIFF
--- a/dartsim/BUILD.bazel
+++ b/dartsim/BUILD.bazel
@@ -47,12 +47,11 @@ public_headers = [
 ]
 
 cc_binary(
-    name = "libignition-physics2-dartsim.so",
+    name = "libignition-physics-dartsim-plugin.so",
     srcs = sources + public_headers,
     includes = ["include", "src"],
-    linkopts = ["-Wl,-soname,libignition-physics2-dartsim.so"],
     linkshared =  True,
-    visibility = [],
+    visibility = ["//visibility:public"],
     deps = [
         "//ign_bazel:utilities",
         "//ign_plugin/core:ign_plugin",
@@ -74,7 +73,6 @@ cc_binary(
 
 cc_library(
     name = "dartsim",
-    srcs = ["libignition-physics2-dartsim.so"],
     hdrs = public_headers,
     includes = ["include", "src"],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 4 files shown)


dartsim/BUILD.bazel
```diff
@@ -47,12 +47,11 @@ public_headers = [
 ]
 
 cc_binary(
-    name = "libignition-physics2-dartsim.so",
```
JShep1, line 50: Was there any reason this was renamed?  There are tests that seem to expect this naming that begin to fail.

dartsim/BUILD.bazel
```diff
@@ -47,12 +47,11 @@ public_headers = [
...
+    name = "libignition-physics-dartsim-plugin.so",
     srcs = sources + public_headers,
     includes = ["include", "src"],
-    linkopts = ["-Wl,-soname,libignition-physics2-dartsim.so"],
```
JShep1, line 53: Tests fail if this line isn't included

dartsim/BUILD.bazel
```diff
@@ -74,7 +73,6 @@ cc_binary(
 
 cc_library(
     name = "dartsim",
-    srcs = ["libignition-physics2-dartsim.so"],
```
JShep1, line 77: I think this line also needs to be included

dartsim/BUILD.bazel
JShep1: I see that some tests in `ign-gazebo` look for a `.so` named `libignition-physics-dartsim-plugin.so` but some tests within `ign-physics` look for a `.so` named `libignition-physics2-dartsim.so`, perhaps we just generate them both?

mjcarroll: Yes, that was the issue.

dartsim/BUILD.bazel
mjcarroll: I don't think that should be the case.  Since this is a dynamically loaded plugin, it shouldn't be linked into the downstream consumers.

dartsim/BUILD.bazel
mjcarroll: I just revisited this, it's because I injected the `libignition-physics2` prefix in the test fixture.  I corrected this back to the non-versioned name, as Gazebo expects in 92e5427

PR #83, opened 2020-07-15
Add link offset to simulation

### LINKED ISSUE #57: Add link offset to tpeplugin::KinematicsFeatures

claireyywang:
Closes #57 

Link offset is added to the simulation after some update in tpeplugin. The following shows distinct offsetted pose of wheels and caster.

Screenshot of demo with `velocity_control.sdf`
successful link offset


claireyywang:
@iche033 Fortunately, I was able to figure out a fix for the problem. I think this is ready for a first pass of review :smiley:

tpe/plugin/src/SimulationFeatures_TEST.cc
```diff
@@ -230,10 +230,6 @@ TEST_P(SimulationFeatures_TEST, FreeGroup)
     auto frameData = model->GetLink(0)->FrameDataRelativeToWorld();
     EXPECT_EQ(ignition::math::Pose3d(0, 0, 2, 0, 0, 0),
               ignition::math::eigen3::convert(frameData.pose));
-    EXPECT_EQ(ignition::math::Vector3d(0.5, 0, 0.1),
-              ignition::math::eigen3::convert(frameData.linearVelocity));
-    EXPECT_EQ(ignition::math::Vector3d(0.1, 0.2, 0),
-              ignition::math::eigen3::convert(frameData.angularVelocity));
```
iche033, line 236: do we need to do something to fix the linear and angular vel?

claireyywang: My understanding is that linear and angular velocity are tied to `model` instead of `link`, and this test case started checking the velocity of `link` after I corrected the implementation of `framedata`, which is why I removed it. Before it worked because when calling `link->FrameDataRelativeToWorld()`, it was getting the velocity of the link's parent model. 

If we do want to add velocity for links, then that would be a different feature which allows links to move by itself.

iche033: oh I see, that makes sense. Just checking, if we assume the link does not move within a model, should it inherit the parent model's velocities in world frame?

In any case, I'm not sure if there's a use case for querying a link's world velocities. We can address this when the need comes.

claireyywang: If the link doesn't move within model, its `WorldPose` is then updated as model's `WorldPose` gets updated (since its offset will remain the same), so we probably don't have to keep track of link velocities in world frame. 

However, if we want the link to move within model, then we'll have to compute the link velocities with the offset. And yes, I agree that we can add this if the need comes.

PR #86, opened 2020-07-21
Add support for nested models in TPE

iche033:
Depends on osrf/sdformat#316

This PR adds support for nested models in TPE plugin (dartsim plugin does not support nested models yet). 

Example sdf description of a nested model:

```
  <model>
      <link>...  </link>
      <model>
      ...
      </model>
  <model>
```

The main changes are:
  * Added functions to allow Models to construct child models.
  * Added support for getting canonical links from a nested model tree
  * Updated SetWorldPose in FreeGroup to work with nested models


tpe/plugin/src/FreeGroupFeatures.cc
```diff
@@ -77,24 +77,56 @@ void FreeGroupFeatures::SetFreeGroupWorldPose(
...
+  }
+  if (!model)
+  {
+    ignerr << "No model for free group with [" << _groupID.id << "] found."
+      << std::endl;
+    return;
+  }
+
+  // set the pose
+  model->SetPose(tfChange * model->GetPose());
```
azeey, line 129: Comparing with the existing code, I see that there was a code path where `link->SetPose` is called, but that's no longer the case. Is that intentional?

iche033: yes that's intentional. TPE does not support moving a model by its canonical link so the idea here is to find the difference between the current pose of link and target pose and apply that transform to the model containing the link.

scpeters:
@azeey so now we want to declare that `ignition::physics::sdf::ConstructSdfModel` does not support nested models? If we require SDFormat 9.3, then we can use the new DOM API's to check if an `sdf::Model` passed to this API has any nested models. Should we add this as a check in our shim function (the one usually in a detail folder)?

* https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics2_2.2.0/sdf/include/ignition/physics/sdf/ConstructModel.hh#L48-L55

At a minimum, we should document the new restriction on the `ignition::physics::sdf::ConstructSdfModel` feature. We can do this in a separate PR, as long as this doesn't get released before the follow-up is merged

iche033:
> At a minimum, we should document the new restriction on the ignition::physics::sdf::ConstructSdfModel feature. We can do this in a separate PR, as long as this doesn't get released before the follow-up is merged

done. added comment in ef6503e

scpeters:
@iche033 I've made a prerelease of sdformat 9.3.0; to use prereleases in CI for this PR, we need the following: https://github.com/osrf/gzdev/pull/16

scpeters:

scpeters:

scpeters:
CI looks good with the prerelease, except for the slight decrease in code coverage

tpe/lib/src/Model.cc
```diff
@@ -51,14 +51,42 @@ Entity &Model::AddLink()
...
+  this->ChildrenChanged();
+  return *it->second.get();
+}
+
 //////////////////////////////////////////////////
 Entity &Model::GetCanonicalLink()
 {
   // return first link as canonical link
-  return *this->GetChildren().begin()->second;
+  for (auto &it : this->GetChildren())
```
scpeters, line 70: this logic doesn't match the approach in https://github.com/osrf/sdformat/pull/341 (which is still under review), which chooses the first link if a model has links, and otherwise gets the canonical link of the first nested model

* https://github.com/osrf/sdformat/pull/341/files#diff-8d3009e318ee9ebbc70c49f879cf8b57R630-R632

iche033: thanks for pointing that out. I've updated the logic and added test in 855b671

tpe/lib/src/Model.cc
```diff
@@ -52,12 +53,50 @@ Entity &Model::AddLink()
...
+
+  it->second->SetParent(this);
+  this->ChildrenChanged();
+  return *it->second.get();
 }
 
+//////////////////////////////////////////////////
+Entity &Model::GetCanonicalLink()
+{
+  std::set<Model *> models;
```
scpeters, line 70: will a `std::set` maintain the same order of nested models?

iche033: I don't think so. Similarly, all the children (links and nested mdoels) in the a TPE Model are stored in an `std::map` which also does not guarantee order. Will that be an issue?

scpeters: the spec calls for the first link to be the canonical link if the attribute `//model/@canonical_link` is unspecified, so if this ignores the xml order then the choice of canonical link may not be consistent

it's a bit pedantic I guess. we can make an issue for it ("TPE canonical link choice may be inconsistent") instead of spending more time on it

iche033: ok I added some logic to mark the first link added to the model as the canonical link, and also added a todo note about handling (or preventing) removal of this link. The order in which the links are added should be consistent with the xml order so the canonical link should also be consistent now. d2a9481

scpeters:
I think we can ignore the abichecker failure, since it was not configured to use prereleases. It should be fixed by https://github.com/ignition-tooling/release-tools/pull/273, but we don't need to wait for that.

tpe/lib/src/Model.cc
scpeters: the handling of links looks good now. for a model with no links and multiple nested models, I'm not sure that it will always pick the same model since it is storing them in a `std::set`, but I'm fine with adding a `todo` comment and opening an issue instead of iterating further

scpeters (commented):
looks ok to me, though I haven't tested it

scpeters (approved):

tpe/lib/src/Model.cc
iche033: sounds good. Added note in 359ca9a4a842ab17e8096ca6a1041e00638b7205 and ticketed issue #100

iche033:

PR #92, opened 2020-08-06
1 ➡️ 2

chapulina:
Also added Focal CI.


dartsim/src/JointFeatures.cc
```diff
@@ -89,7 +89,12 @@ void JointFeatures::SetJointAcceleration(
 void JointFeatures::SetJointForce(
     const Identity &_id, const std::size_t _dof, const double _value)
 {
-  this->ReferenceInterface<JointInfo>(_id)->joint->setForce(_dof, _value);
+  auto joint = this->ReferenceInterface<JointInfo>(_id)->joint;
+  if (joint->getActuatorType() != dart::dynamics::Joint::FORCE)
+  {
+    joint->setActuatorType(dart::dynamics::Joint::FORCE);
```
scpeters, line 95: @azeey I guess codecov didn't run for #74, but it's now complaining that the patch coverage is too low. This is the only line without coverage. Did you expect this line to be covered by the current tests?

azeey: Ah, no, I don't think we have a test that uses both SERVO and FORCE actuator types. I can add one to ign-physics1, but I'm not sure I can get to it this week.

scpeters: it's ok; I was just checking

scpeters:

scpeters:
I'm going to manually squash the commits that are exclusive to this branch before merging

scpeters:
I couldn't figure out how to squash into a merge commit (3fb287d), so I just squashed the commits on top of that

scpeters:
windows build had a flaky failure; I'm ignoring it and will merge

PR #93, opened 2020-08-12
Add profiler code to TPE

claireyywang:
To enable and test with profiler
```colcon build --merge-install --packages-select ignition-physics2 --cmake-args -DENABLE_PROFILER=1```

and simulation worlds in https://github.com/ignitionrobotics/ign-gazebo/pull/283


claireyywang:
cc @iche033

PR #97, opened 2020-08-20
Integrate AABB tree for collision checking in TPE

iche033:
Initial performance tests revealed a bottleneck in TPE's naive collision detection algorithm. This PR integrates an AABB tree from a third party library to help with collision detection by checking for overlaps between objects' axis aligned bounding boxes.

I tested 3 AABB tree libraries and [the one](https://github.com/lohedges/aabbcc) chosen has the best performance while providing more features like tree balancing.

Here're graphs comparing the `World::Step` performance in ign-gazebo's [shapes_population.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/examples/worlds/shapes_population.sdf) world that consists of 3000 models.

TPE before (~779ms) :

tpe_step_before

TPE after using AABB tree (~17ms)

tpe_step_aabb_tree

Dart (~155ms):

dart_step

The performance of TPE with AABB tree is roughly 45x faster than before and 9x faster than Dart

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "tpe/lib/src/aabb_tree"
--- a/tools/code_check.sh
+++ b/tools/code_check.sh
@@ -46,14 +46,15 @@ then
   QUICK_TMP=`mktemp -t asdfXXXXXXXXXX`
 else
   CHECK_DIRS="./src ./include ./test/integration ./test/regression ./test/performance ./tpe"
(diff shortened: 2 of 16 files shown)


scpeters:
@iche033 to exclude some code from coverage testing before merging

iche033:
added `codecov.yml` in 9da47cc. Code coverage is now back up.

iche033:
CI looks good. Merging

PR #107, opened 2020-09-29
Bumped examples to ign-physics3

ahcorde:


chapulina (approved):
Nice catch, we should be running these examples through CI, I ticketed an issue: https://github.com/ignitionrobotics/ign-physics/issues/109

PR #110, opened 2020-09-29
Update readme

JShep1:
Do we also want to add the source install section?

Signed-off-by: John Shepherd <john@openrobotics.org>


chapulina (approved):
I see source install instructions here, so I think this is enough

https://ignitionrobotics.org/api/physics/3.0/installation.html

PR #113, opened 2020-09-30
Bump to 3.0.0 and update changelog

scpeters:
Signed-off-by: Steve Peters <scpeters@openrobotics.org>


chapulina (approved):
A major release just to bump a dependency :grimacing:

azeey (approved):

scpeters:
> A major release just to bump a dependency

pretty much. we don't have many times when we have to break ABI because we're using plugins. the main change from 1 -> 2 was the thing that sped up compilation

PR #127, opened 2020-10-11
Fix getting model bounding box in world frame in TPE

iche033:
By default the `GetAxisAlignedBoundingBox` feature requests a model's bounding box in world frame, and computes it by merging all child links and shapes' bounding boxes in the requested frame. The problem was that we did not have support for getting frame data for shapes in the KinematicsFeatures so the bounding box returned was incorrect. The PR adds an extra check in the collision (shape) list when searching for frame data.

Signed-off-by: Ian Chen <ichen@osrfoundation.org>

### DIFF
--- a/tpe/plugin/src/KinematicsFeatures.cc
+++ b/tpe/plugin/src/KinematicsFeatures.cc
@@ -39,9 +39,8 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
     return data;
   }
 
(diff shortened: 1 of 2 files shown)


luca-della-vedova (approved):
Works for me! Good to see the more comprehensive tests as well!

scpeters (approved):

scpeters:

claireyywang (approved):

PR #136, opened 2020-10-20
[Bullet] FreeGroupFeatures + convert functions

Lobotuerk:
FreeGroupFeatures properly detected by ignition.
Convert from eigen to bullet functions were moved to base so any file can reuse them instead of having to declare them again

### DIFF
--- a/bullet/src/FreeGroupFeatures.cc
+++ b/bullet/src/FreeGroupFeatures.cc
@@ -4,6 +4,7 @@ namespace ignition {
 namespace physics {
 namespace bullet {
 
+/////////////////////////////////////////////////
 Identity FreeGroupFeatures::FindFreeGroupForModel(
     const Identity &_modelID) const
 {
@@ -19,6 +20,32 @@ Identity FreeGroupFeatures::FindFreeGroupForModel(
     return this->GenerateInvalidId();
 
   return _modelID;
+}
+
+/////////////////////////////////////////////////
+Identity FreeGroupFeatures::FindFreeGroupForLink(
+    const Identity &_linkID) const
+{
+  const auto &model = this->models.at(this->links.at(_linkID)->model)->model;
+
+  // If there are no links at all in this model, then the FreeGroup functions
+  // will not work properly, so we'll just reject these cases.
+  if (model->getNumLinks() == 0)
+    return this->GenerateInvalidId();
+
+  // Reject also if the model has fixed base
+  if (model->hasFixedBase())
+    return this->GenerateInvalidId();
+
+  return _linkID;
+}
+
+/////////////////////////////////////////////////
+Identity FreeGroupFeatures::GetFreeGroupCanonicalLink(
+    const Identity &_groupID) const
+{
+  (void) _groupID;
+  return this->GenerateInvalidId();
 }
 
 /////////////////////////////////////////////////
--- a/bullet/src/FreeGroupFeatures.cc
+++ b/bullet/src/FreeGroupFeatures.cc
@@ -1,7 +1,5 @@
 #include "FreeGroupFeatures.hh"
 
-#include <ignition/math/eigen3/Conversions.hh>
-
 namespace ignition {
 namespace physics {
 namespace bullet {
--- a/bullet/src/FreeGroupFeatures.cc
+++ b/bullet/src/FreeGroupFeatures.cc
@@ -37,26 +37,27 @@ Identity FreeGroupFeatures::FindFreeGroupForLink(
 Identity FreeGroupFeatures::GetFreeGroupCanonicalLink(
     const Identity &_groupID) const
 {
+  (void) _groupID;
+  //  TODO(lobotuerk) when joints are supported, fix canonical getter
   //  Verify that link exists
-  const auto &link_it = this->links.find(_groupID);
-  if (link_it != this->links.end() && link_it->second != nullptr)
-  {
-    auto &model = this->models.at(link_it->second->model)->model;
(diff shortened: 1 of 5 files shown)


bullet/src/FreeGroupFeatures.hh
```diff
@@ -0,0 +1,41 @@
...
+>;
+
+class FreeGroupFeatures
+    : public virtual Base,
+      public virtual Implements3d<FreeGroupFeatureList>
+{
+  // ----- FindFreeGroupFeature -----
+  Identity FindFreeGroupForModel(const Identity &_modelID) const override;
+
+  Identity FindFreeGroupForLink(const Identity &/* _linkID */) const override
```
Blast545, line 24: This and `GetFreeGroupCanonicalLink` seem to be a dummy implementation, did you check if we can add those before merging?

bullet/src/FreeGroupFeatures.cc
```diff
@@ -0,0 +1,45 @@
...
+{
+  // Convert pose
+  const auto poseTranslation = _pose.translation();
+  const auto poseLinear = _pose.linear();
+  btTransform baseTransform;
+  baseTransform.setOrigin(convertVec(poseTranslation));
+  baseTransform.setBasis(convertMat(poseLinear));
+
+  // Set base transform
+  const auto &model = this->models.at(_groupID)->model;
```
Blast545, line 39: There seems to be something odd here, the function receives a group_ID, but operates on the model IDs, have any clue why is this happening like this?

Blast545 (commented):
Pls check my comments

bullet/src/FreeGroupFeatures.cc
Lobotuerk: A FreeGroupWorld has a model inside of it and I think this function sets the pose of that model so the Group has that pose

bullet/src/FreeGroupFeatures.hh
Lobotuerk: Im not even sure if those 2 functions are used at all

Lobotuerk: Update: Seem to be necessary for ignition, even if dummy, I'll look into implementing them

bullet/src/FreeGroupFeatures.cc
Blast545: Correct me if I'm wrong, the model queried `this->models.at(_groupID)` gets the `models` unordered map from the base class. The code seems to imply with the naming `_groupID` that there is a list of groups, and I think at the moment we don't have groups defined anywhere. 

I think you are saying that a Model can be a FreeWorldGroup?

Lobotuerk: I think it can

bullet/src/FreeGroupFeatures.hh
Lobotuerk: 9dafb2d Updates Canonical, but i need to finish that template (does not compile yet). Link has been added on c1e145c

Lobotuerk:
@Blast545 After analysing GetFreeGroupCanonicalLink function, I left the template there, but did not finish it as the whole system does not support it, check https://github.com/ignitionrobotics/ign-physics/blob/devel_bullet/bullet/src/SDFFeatures.cc#L132, usually base AKA canonical links have `-1` parent, but we are using that on everyone as we dont support joints ergo all links are cannonical

bullet/src/FreeGroupFeatures.cc
```diff
@@ -0,0 +1,82 @@
...
+{
+  //  Verify that link exists
+  const auto &link_it = this->links.find(_groupID);
+  if (link_it != this->links.end() && link_it->second != nullptr)
+  {
+    auto &model = this->models.at(link_it->second->model)->model;
+    //  Search on that link's parents for the base
+    for (int link_num = 0; link_num < model->getNumLinks(); link_num++)
+    {
+      if (model->getParent(link_num) == -1)
```
Blast545, line 48: This portion of the code is browsing through the links but at the end does nothing? You can leave the GenerateInvalidId and leave the rest under the To-Do for the moment if you'd like

bullet/src/FreeGroupFeatures.cc
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/////////////////////////////////////////////////
+Identity FreeGroupFeatures::FindFreeGroupForLink(
+    const Identity &_linkID) const
+{
+  const auto &link_it = this->links.find(_linkID);
+
+  if (link_it != this->links.end() && link_it->second != nullptr)
+    return this->GenerateIdentity(_linkID.id, link_it->second);
```
Blast545, line 32: In this case, isn't it the same to just return the `link_it` object?

Lobotuerk: link_it is an iterator, but if you mean returning the same ID, i think FindFreeGroup returns a new link, like in dart https://github.com/ignitionrobotics/ign-physics/blob/main/dartsim/src/FreeGroupFeatures.cc#L85

bullet/src/FreeGroupFeatures.cc
Lobotuerk: Ok

Lobotuerk: 6394afc changes it into dummy

Blast545 (approved):

Lobotuerk:
Note: Merging this with `GetFreeGroupCanonicalLink` as a dummy, will iterate further once we get into joint features

PR #137, opened 2020-10-20
Ignore invalid joint commands

azeey:
Robots falling through the floor were reported in https://github.com/osrf/subt/issues/675. This was caused by an input joint velocity command of `nan`  being passed to `JointFeatures::SetJointVelocityCommand` via ign-gazebo's DiffDrive system. In DART, Joint velocity commands are handled by the constraint solver, which also handles collisions, so it makes sense that the robot falls through when the solver fails. The solution here is to ignore `nan` values in `JointFeatures::SetJointVelocityCommand`.

I have taken the opportunity to ignore invalid values in other `JointFeatures::SetJoint*` commands as well.

### DIFF
--- a/dartsim/src/JointFeatures.cc
+++ b/dartsim/src/JointFeatures.cc
@@ -67,22 +67,57 @@ Pose3d JointFeatures::GetJointTransform(const Identity &_id) const
 void JointFeatures::SetJointPosition(
     const Identity &_id, const std::size_t _dof, const double _value)
 {
-  this->ReferenceInterface<JointInfo>(_id)->joint->setPosition(_dof, _value);
+  auto joint = this->ReferenceInterface<JointInfo>(_id)->joint;
+
+  // Take extra care that the value is finite. A nan can cause the DART
+  // constraint solver to fail, which will in turn either cause a crash or
+  // collisions to fail
+  if (!std::isfinite(_value))
+  {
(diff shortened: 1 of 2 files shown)


chapulina (approved):
Wow some bugs are just amazing. The fix looks good :+1:

nkoenig:
Thanks for finding this!

mxgrey (approved):
This seems like a reasonable safety measure, although I'm alarmed that the DiffDrive system would be outputting NaNs. I hope that issue can be addressed as well.

azeey:
> This seems like a reasonable safety measure, although I'm alarmed that the DiffDrive system would be outputting NaNs. I hope that issue can be addressed as well.

Yeah, we should probably do a better job of validating user input in ign-gazebo.

azeey:
Jenkins CI looks good, but the GitHub actions didn't run. @chapulina, do you know why? is there a way to start them manually?

chapulina:
> the GitHub actions didn't run. @chapulina, do you know why?

I've only been fixing the actions for forks from Citadel onwards because Blueprint will go EOL soon and I wasn't expecting many fork PRs for it.

> is there a way to start them manually?

I don't think so. You could open a PR from a branch... But I think this is safe to merge just with Jenkins CI, so I'll go ahead and do that.

PR #138, opened 2020-10-20
[bullet] Add RemoveEntities features

Blast545:
Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/bullet/src/EntityManagementFeatures.cc
+++ b/bullet/src/EntityManagementFeatures.cc
@@ -29,6 +29,108 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
       {world, _name, collisionConfiguration, dispatcher, broadphase, solver});
 }
 
+/////////////////////////////////////////////////
+bool EntityManagementFeatures::RemoveModel(const Identity &_modelID)
+{
+  // Check if the model exists
+  if (this->models.find(_modelID) == this->models.end()){
+    return false;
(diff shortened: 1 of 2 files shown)


bullet/src/EntityManagementFeatures.cc
```diff
@@ -29,6 +29,108 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
...
+}
+
+bool EntityManagementFeatures::ModelRemoved(
+  const Identity &_modelID) const
+{
+  return this->models.find(_modelID) == this->models.end();
+}
+
+bool EntityManagementFeatures::RemoveModelByIndex(
+  const Identity &/* _worldID */, std::size_t _modelIndex)
```
Lobotuerk, line 73: Identity of the model and Index of the model seem to be the same right? Cant you just use `RemoveModel(_modelIndex)` ?

Lobotuerk (approved):

bullet/src/EntityManagementFeatures.cc
Blast545: `RemoveModelByIndex` receives a `size_t` index, while the other version receives an Identity object, I could rewrite the one receiving the identity to call this one instead, but I'm not sure in which contexts these functions are used. So we can reconsider this if needed later after we test the features

PR #139, opened 2020-10-21
Add restitution coefficient support for bouncing

diegoferigo:
This PR adds the support of simulating bouncing using [`//collision/surface/bounce/restitution_coefficient`](http://sdformat.org/spec?ver=1.7&elem=collision#surface_bounce). It calls the methods introduced in https://github.com/dartsim/dart/pull/1369.

This is the resulting behaviour simulating 11 spheres with a different restitution coefficient and setting the coefficient of the ground to `1.0`:

out

<details>

<summary>Script (requires third party libraries)</summary>

```py
import numpy as np
from dataclasses import dataclass
from gym_ignition.utils import misc
from scenario import core as scenario_core
from scenario import gazebo as scenario_gazebo
from gym_ignition.utils.scenario import init_gazebo_sim


@dataclass
class SphereURDF:

    mass: float = 5.0
    radius: float = 0.1
    restitution: float = 0

    def urdf(self) -> str:

        i = 2.0 / 5 * self.mass * self.radius * self.radius
        urdf = f"""
            <robot name="sphere_robot" xmlns:xacro="http://www.ros.org/wiki/xacro">
                <link name="sphere">
                    <inertial>
                      <origin rpy="0 0 0" xyz="0 0 0"/>
                      <mass value="{self.mass}"/>
                      <inertia ixx="{i}" ixy="0" ixz="0" iyy="{i}" iyz="0" izz="{i}"/>
                    </inertial>
                    <visual>
                      <geometry>
                        <sphere radius="{self.radius}"/>
                      </geometry>
                      <origin rpy="0 0 0" xyz="0 0 0"/>
                    </visual>
                    <collision>
                      <geometry>
                        <sphere radius="{self.radius}"/>
                      </geometry>
                      <origin rpy="0 0 0" xyz="0 0 0"/>
                    </collision>
                </link>
                <gazebo reference="sphere">
                  <collision><surface><bounce>
                    <restitution_coefficient>{self.restitution}</restitution_coefficient>
                  </bounce></surface></collision>
                </gazebo>
            </robot>
            """

        return urdf


# Set the verbosity
scenario_gazebo.set_verbosity(scenario_gazebo.Verbosity_debug)

# Get the default simulator and the default empty world
gazebo, world = init_gazebo_sim()

# Show the GUI
import time
gazebo.gui()
gazebo.run(paused=True)
time.sleep(3)

for restitution in np.linspace(start=0, stop=1.0, num=11):

    # Create a tmp file with the urdf
    sphere_urdf = SphereURDF(restitution=restitution).urdf()
    sphere_urdf_file = misc.string_to_file(sphere_urdf)

    # Insert the sphere
    pose = scenario_core.Pose_identity()
    pose.position = [0, restitution * 3, 1.0]
    assert world.to_gazebo().insert_model(sphere_urdf_file, pose, f"sphere_{restitution}")

gazebo.run(paused=True)

for _ in range(10_000):
    gazebo.run()

time.sleep(5)
gazebo.close()

```

</details>

This PR fills one of the gaps described in https://www.allisonthackston.com/articles/ignition_vs_gazebo.html.

### DIFF
--- a/dartsim/src/SDFFeatures_TEST.cc
+++ b/dartsim/src/SDFFeatures_TEST.cc
@@ -158,20 +158,26 @@ TEST(SDFFeatures_TEST, CheckDartsimData)
     EXPECT_DOUBLE_EQ(0.0, axis[2]);
   }
 
-  const dart::dynamics::SkeletonPtr freeBody =
+  {
+    const dart::dynamics::SkeletonPtr freeBody =
       dartWorld->getSkeleton("free_body");
-  ASSERT_NE(nullptr, freeBody);
-  ASSERT_EQ(1u, freeBody->getNumBodyNodes());
-  const dart::dynamics::BodyNode *bn = freeBody->getBodyNode(0);
-  ASSERT_NE(nullptr, bn);
+    ASSERT_NE(nullptr, freeBody);
+    ASSERT_EQ(1u, freeBody->getNumBodyNodes());
+    const dart::dynamics::BodyNode *bn = freeBody->getBodyNode(0);
+    ASSERT_NE(nullptr, bn);
 
-  EXPECT_TRUE(dynamic_cast<const dart::dynamics::FreeJoint*>(
-                bn->getParentJoint()));
+    EXPECT_TRUE(dynamic_cast<const dart::dynamics::FreeJoint*>(
+          bn->getParentJoint()));
 
-  const Eigen::Vector3d translation = bn->getTransform().translation();
-  EXPECT_DOUBLE_EQ(0.0, translation[0]);
-  EXPECT_DOUBLE_EQ(10.0, translation[1]);
-  EXPECT_DOUBLE_EQ(10.0, translation[2]);
+    const Eigen::Vector3d translation = bn->getTransform().translation();
+    EXPECT_DOUBLE_EQ(0.0, translation[0]);
+    EXPECT_DOUBLE_EQ(10.0, translation[1]);
+    EXPECT_DOUBLE_EQ(10.0, translation[2]);
+
+    const dart::dynamics::ShapeNode *collision1 = bn->getShapeNode(0);
+    auto aspect = collision1->getDynamicsAspect();
+    EXPECT_EQ(0.8, aspect->getRestitutionCoeff());
+  }
 
   const dart::dynamics::SkeletonPtr screwJointTest =
       dartWorld->getSkeleton("screw_joint_test");
--- a/dartsim/src/SDFFeatures_TEST.cc
+++ b/dartsim/src/SDFFeatures_TEST.cc
@@ -174,9 +174,13 @@ TEST(SDFFeatures_TEST, CheckDartsimData)
     EXPECT_DOUBLE_EQ(10.0, translation[1]);
     EXPECT_DOUBLE_EQ(10.0, translation[2]);
 
+#if DART_VERSION_AT_LEAST(6, 10, 0)
     const dart::dynamics::ShapeNode *collision1 = bn->getShapeNode(0);
     auto aspect = collision1->getDynamicsAspect();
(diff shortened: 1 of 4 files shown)


diegoferigo:
Closes https://github.com/ignitionrobotics/ign-physics/issues/84

azeey (changes requested):
Thanks for the contribution @diegoferigo. The PR looks good. It does demonstrate the limitation of the current API that uses SDFormat objects for construction because it make it possible for two physics engine plugins to implement the `Construct*` feature, but have different capabilities in reality. i.e, in this case, if there is another physics engine plugin that implements the `ConstructCollision` feature, but doesn't support restitution, we now have a discrepancy. 

For now though, I think we should go forward with this PR. Would it be possible for you to add a test?

diegoferigo:
@azeey Thanks for the feedback. I get back to this issue to say that lately I'm running a bit low with time, not sure if I manage to add the test in the next days.

> It does demonstrate the limitation of the current API that uses SDFormat objects for construction

Yes, I think this is the trade off between API complexity and flexibility. I'm not sure what it the best solution here, I just say that often the flexibility is good to have, especially in this development phase, and it can be compensated with a proper documentation so that downstream users know what they can and cannot expect.

azeey (approved):
LGTM! I took the liberty to add a small test.

dartsim/src/SDFFeatures_TEST.cc
```diff
@@ -158,20 +158,26 @@ TEST(SDFFeatures_TEST, CheckDartsimData)
...
+          bn->getParentJoint()));
+
+    const Eigen::Vector3d translation = bn->getTransform().translation();
+    EXPECT_DOUBLE_EQ(0.0, translation[0]);
+    EXPECT_DOUBLE_EQ(10.0, translation[1]);
+    EXPECT_DOUBLE_EQ(10.0, translation[2]);
+
+    const dart::dynamics::ShapeNode *collision1 = bn->getShapeNode(0);
+    auto aspect = collision1->getDynamicsAspect();
+    EXPECT_DOUBLE_EQ(0.8, aspect->getRestitutionCoeff());
```
chapulina, line 179: This test failed on GitHub actions Focal CI:

```
  [ RUN      ] SDFFeatures_TEST.CheckDartsimData
  [Err] [SDFFeatures.cc:814] Asked to construct a joint of sdf::JointType [7], but that is not supported yet. Creating a FIXED joint instead
  [Err] [SDFFeatures.cc:814] Asked to construct a joint of sdf::JointType [4], but that is not supported yet. Creating a FIXED joint instead
  [Err] [SDFFeatures.cc:814] Asked to construct a joint of sdf::JointType [2], but that is not supported yet. Creating a FIXED joint instead
  /github/workspace/dartsim/src/SDFFeatures_TEST.cc:179: Failure
  Expected equality of these values:
    0.8
      Which is: 0.80000000000000004
    aspect->getRestitutionCoeff()
      Which is: 0
  [  FAILED  ] SDFFeatures_TEST.CheckDartsimData (128 ms)
```

Is this related to the DART version?

azeey: Yes, the coefficient was only being set for DART>=6.10.0. 3e6310a fixes it, but note that for older versions of DART, the restitution coefficient has to be set on the BodyNode instead of the shape. See https://github.com/dartsim/dart/issues/200

azeey:

diegoferigo:
> LGTM! I took the liberty to add a small test.

Thanks @azeey, few weeks ago I tried to add a test but I couldn't find anything simple enough (and delayed again). I was thinking something about computing the average velocity of the spheres and checking that those with bigger restitution coefficients were higher. Though, it was a bit convoluted and in any case it was not checking against a ground truth.

azeey:
CI looks good. Merging!

PR #140, opened 2020-10-21
[Bullet] Remove dummy engine feature

Blast545:
This was added in the initial PR to have a dummy feature to be able of compiling the plugin. However, is not expected to be there for the basic tests we are targetting.


Lobotuerk (approved):
So, basically, we don't need those functions at all for a first run?

Blast545:
@Lobotuerk Yes, those are not needed! I'll proceed merging, and later on we might need to add them as part of the entity management features. (the ones being removed with this PR are dummy versions in any case)

PR #142, opened 2020-10-22
Support setting canonical link 

### LINKED ISSUE #100: [TPE] Return first link as canonical link
The SDF spec calls for the first link to be the canonical link if the attribute `//model/@canonical_link` is unspecified. However, TPE stores entities in an `std::map` which does not preserve order in which the entities are added.

pull request #86 attempted to workaround this by marking the first link added as the canonical link but the logic does not work when the model has multiple nested models and no top level links.

see this comment for more info: https://github.com/ignitionrobotics/ign-physics/pull/86#discussion_r483307451

claireyywang:
Set canonical link if defined in sdformat file, otherwise set first link of model as canonical. Closes #100

### DIFF
--- a/tpe/lib/src/Model.cc
+++ b/tpe/lib/src/Model.cc
@@ -88,7 +88,7 @@ Entity &Model::AddModel()
 }
 
 //////////////////////////////////////////////////
-void Model::SetCanonicalLink(const std::size_t linkId)
+void Model::SetCanonicalLink(std::size_t linkId)
 {
   if (linkId != kNullEntityId)
   {
@@ -109,21 +109,6 @@ Entity &Model::GetCanonicalLink()
   if (linkEnt.GetId() != kNullEntityId)
   {
     return linkEnt;
-  }
-
-  // todo(anyone) the code below does not guarantee that the link returned is
-  // the first link defined in SDF since GetChildren returns a std::map, and
-  // likewise the use of std::set, do not preserve order.
-  std::set<Model *> models;
-  for (auto &it : this->GetChildren())
-  {
-    // if child is nested model, store it first and only return nested
-    // links if there are no links in this model
-    Model *model = dynamic_cast<Model *>(it.second.get());
-    if (model)
-    {
-      models.insert(model);
-    }
   }
 
   for (auto m : models)
--- a/tpe/lib/src/Model.cc
+++ b/tpe/lib/src/Model.cc
@@ -90,11 +90,8 @@ Entity &Model::AddModel()
 //////////////////////////////////////////////////
 void Model::SetCanonicalLink(std::size_t linkId)
 {
-  if (linkId != kNullEntityId)
-  {
-    this->dataPtr->canonicalLinkId = linkId;
-  }
-  else if (this->dataPtr->firstLinkId != kNullEntityId)
+  this->dataPtr->canonicalLinkId = linkId;
+  if (this->dataPtr->canonicalLinkId == kNullEntityId)
   {
     this->dataPtr->canonicalLinkId = this->dataPtr->firstLinkId;
   }
@@ -110,14 +107,22 @@ Entity &Model::GetCanonicalLink()
   {
     return linkEnt;
   }
-
-  for (auto m : models)
+  else
   {
-    auto &link = m->GetCanonicalLink();
-    if (link.GetId() != kNullEntity.GetId())
-      return link;
+    for (auto &child : this->GetChildren())
+    {
+      Entity childEnt = *(child.second);
+      Model *nestedModel = static_cast<Model *>(&childEnt);
+      if (nestedModel != nullptr)
+      {
+        Entity &ent = nestedModel->GetChildById(this->dataPtr->canonicalLinkId);
+        if (ent.GetId() != kNullEntityId)
+        {        
+          return ent;
+        }
+      }
+    }
   }
-
   return kNullEntity;
 }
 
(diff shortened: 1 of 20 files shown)


claireyywang:
I have some questions related to nested model @iche033 @scpeters 
1. Do each model (nested or not) only have one canonical link or each nested model can have its own canonical link? 
2. If a model doesn’t have a top level link, does it take the nested model’s first link as the canonical link?

So far I only added a generic function in `Model.cc` and let `SdfFeatures.cc` set the canonical link since that's when we can determine whether `CanonicalLink` is defined or not for a model. I'm not sure if canonical link canbe set at `Model::AddLink()` due to limited information about a link (whether I should set first link or defined `<canonical_link>`).

tpe/lib/src/Model_TEST.cc
```diff
@@ -307,19 +312,4 @@ TEST(Model, NestedModel)
...
-  // add link in top level model and verify the new link becomes the canonical
-  // link in the tree as preference is given to links in higher level
-  Entity &link = model.AddLink();
-  EXPECT_EQ(link.GetId(), model.GetCanonicalLink().GetId());
-
-  // add a few more links in top level model and verify canonical link stays
-  // the same
-  model.AddLink();
-  model.AddLink();
-  EXPECT_EQ(link.GetId(), model.GetCanonicalLink().GetId());
```
iche033, line 324: can we add back a test with a nested model and make sure we get the correct canonical link? I am thinking of a model structure like the one below:

```
<model name="m0">
  <model name="m1">
    <link name="x" />
  </model>
  <model name="m2">
    <link name="y" />
  </model>
</model>
```

If we set link `y` to be the canonical link and call `m0->GetCanonicalLink()`, would we be able to get back link `y`?

tpe/lib/src/Model.cc
```diff
@@ -85,35 +87,42 @@ Entity &Model::AddModel()
   return *it->second.get();
 }
 
+//////////////////////////////////////////////////
+void Model::SetCanonicalLink(const std::size_t linkId)
```
iche033, line 91: just to be consistent with other function, we stopped using `const` for primitive arg types

tpe/lib/src/Model.cc
```diff
@@ -85,35 +87,42 @@ Entity &Model::AddModel()
...
-    else
+    Model *model = dynamic_cast<Model *>(it.second.get());
+    if (model)
     {
-      Model *model = dynamic_cast<Model *>(it.second.get());
-      if (model)
-      {
-        models.insert(model);
-      }
+      models.insert(model);
```
iche033, line 125: I think we shouldn't have to insert into a set any more. The logic was previously added because we want to process all the links first before moving onto models but we're no longer checking for links here in this for loop.

iche033 (commented):
thanks for the changes. I have a couple minor comments, and one about adding a test to verify canonical link in a nested model

tpe/lib/src/Model.cc
```diff
@@ -85,45 +87,42 @@ Entity &Model::AddModel()
...
-      Model *model = dynamic_cast<Model *>(it.second.get());
-      if (model)
+      Entity childEnt = *(child.second);
+      Model *nestedModel = static_cast<Model *>(&childEnt);
+      if (nestedModel != nullptr)
       {
-        models.insert(model);
+        Entity &ent = nestedModel->GetChildById(this->dataPtr->canonicalLinkId);
+        if (ent.GetId() != kNullEntityId)
+        {
```
iche033, line 120: can you remove extra spaces here?

tpe/plugin/src/SDFFeatures.cc
```diff
@@ -171,6 +183,18 @@ Identity SDFFeatures::ConstructSdfNestedModel(
...
+  {
+    std::string canonicalLinkName = _sdfModel.CanonicalLinkName();
+    tpelib::Entity &canonicalLink = model->GetChildByName(canonicalLinkName);
+    model->SetCanonicalLink(canonicalLink.GetId());
+  }
+  else
+  {
+    model->SetCanonicalLink();
+  }
+
```
iche033, line 197: I think there are extra spaces here

iche033 (commented):
Just a couple of minor code check issues. I added some checks in SDFFeatures_TEST and tests pass. Feel free to incorporate these changes

```
diff --git a/tpe/plugin/src/SDFFeatures_TEST.cc b/tpe/plugin/src/SDFFeatures_TEST.cc
index bbef33f..78cd972 100644
--- a/tpe/plugin/src/SDFFeatures_TEST.cc
+++ b/tpe/plugin/src/SDFFeatures_TEST.cc
@@ -37,6 +37,7 @@
 #include <ignition/physics/sdf/ConstructWorld.hh>
 
 #include "lib/src/Entity.hh"
+#include "lib/src/Model.hh"
 #include "lib/src/World.hh"
 #include "World.hh"
 
@@ -413,6 +414,12 @@ TEST(SDFFeatures_TEST, NestedModel)
       nestedCollision.GetId());
   EXPECT_EQ("nested_collision", nestedCollision.GetName());
   EXPECT_EQ(ignition::math::Pose3d::Zero, nestedCollision.GetPose());
+
+  // canonical link
+  ignition::physics::tpelib::Model *m =
+      static_cast<ignition::physics::tpelib::Model *>(&model);
+  ignition::physics::tpelib::Entity canLink = m->GetCanonicalLink();
+  EXPECT_EQ(link.GetId(), canLink.GetId());
 }
```

tpe/lib/src/Model.hh
```diff
@@ -52,6 +52,10 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Model : public Entity
   /// \return Newly created nested model
   public: Entity &AddModel();
 
+  /// \brief Set the canonical link of model
+  public: void SetCanonicalLink(
+    std::size_t linkId=kNullEntityId);
```
iche033, line 57: add space around `=`

iche033 (approved):

PR #146, opened 2020-10-23
[bullet] Test simulation output

Blast545:


Blast545:
Merging this, to close temporary the work being done on top of devel_bullet to change to the classic implementation of bullet instead of featherstone

Lobotuerk (approved):
LGTM

PR #148, opened 2020-10-28
dartsim-plugin windows build fixes

scpeters:
These are some fixes needed to compile against dartsim on Windows, as mentioned in #87. Though dartsim version 6.9.4 is [now available in vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/dartsim/portfile.cmake), this version is missing some needed fixes (https://github.com/dartsim/dart/pull/1510, https://github.com/dartsim/dart/pull/1513), so I recommend building dart's `master` branch from source in a colcon workspace alongside this branch. Unfortunately, dart still seems to have some runtime errors (https://github.com/dartsim/dart/issues/1522), so all the tests that use the `dartsim-plugin` currently fail, but at least the build succeeds.

The fixes include:

* creating the unversioned plugin by copy instead of symlink, following the example of TPE
* setting some compiler flags to fix build errors (https://github.com/dartsim/dart/issues/753) and disable some warnings
* remove features related to QTest and UNIX from the `check_test_ran.py` script (based on ign-cmake's [copy of this file](https://github.com/ignitionrobotics/ign-cmake/blob/ignition-cmake2_2.5.0/tools/check_test_ran.py))

### DIFF
--- a/dartsim/CMakeLists.txt
+++ b/dartsim/CMakeLists.txt
@@ -7,6 +7,10 @@ ign_add_component(dartsim INTERFACE
 
 target_link_libraries(${features} INTERFACE ${DART_LIBRARIES})
 target_include_directories(${features} SYSTEM INTERFACE ${DART_INCLUDE_DIRS})
+if (MSVC)
+  # needed by DART, see https://github.com/dartsim/dart/issues/753
+  target_compile_options(${features} INTERFACE "/permissive-")
+endif()
 
 install(
   DIRECTORY include/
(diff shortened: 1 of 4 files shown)


JShep1 (approved):
Thanks, these fixes allow me to build against dart.  Although, it should be important to note that `ign-physics` now won't compile since it uses @azeey / Ignition's fork (https://github.com/ignition-forks/dart) and not the standard dart `master` branch.

PR #151, opened 2020-10-29
[Bullet] Erase Entity - Memory Issue

Lobotuerk:
Way to reproduce:
After compiling and sourcing, running `ign gazebo -s -v4 --physics-engine ignition-physics3-bullet-plugin shapes.sdf`, pressing play and deleting the sphere should trigger the `free` related segfault

### DIFF
--- a/bullet/src/Base.hh
+++ b/bullet/src/Base.hh
@@ -108,11 +108,14 @@ class Base : public Implements3d<FeatureList<Feature>>
 
   public: inline std::size_t GetNextEntity()
   {
+    ignerr << "GetNextEntity" << std::endl;
     return entityCount++;
   }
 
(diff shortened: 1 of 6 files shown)


chapulina:
Before we get the Bullet work into a release branch, we should make sure it is being tested on CI. For Homebrew, it looks like the bullet component is already being successfully built. It's missing here:

* Linux (actions and Jenkins): https://github.com/ignitionrobotics/ign-physics/blob/main/.github/ci/packages.apt
* Windows, I think it goes here: https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/ign_physics-default-devel-windows-amd64.bat

Lobotuerk:
Closing PR as we are moving bullet from its Featherstone implementation to Dynamics

PR #152, opened 2020-10-29
[Bullet] Fixed RemoveModel

Lobotuerk:
Signed-off-by: lobotuerk <jtlorente@ekumenlabs.com>

### DIFF
--- a/bullet/src/EntityManagementFeatures.cc
+++ b/bullet/src/EntityManagementFeatures.cc
@@ -33,7 +33,7 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
 bool EntityManagementFeatures::RemoveModel(const Identity &_modelID)
 {
   // Check if the model exists
-  if (this->models.find(_modelID) == this->models.end()){
+  if (this->models.find(_modelID.id) == this->models.end()){
     return false;
   }
   // Current implementation does not include collisions nor joints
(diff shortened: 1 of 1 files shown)


bullet/src/EntityManagementFeatures.cc
```diff
@@ -33,31 +33,28 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
...
+    it++;
   }
 
-  // (To-DO blast545): test if its required to make an extra call to
-  // world->removeMultiBody(model) to remove it from the bullet system
-  // or is it enough just by deleting its pointed memory
-
   // Clean up model
-  delete modelInfo->model;
+  this->worlds.at(this->models.at(_modelID)->world)->world->removeMultiBody(this->models.at(_modelID)->model);
```
Blast545, line 56: In line 36, the get to the model is `this->models.find(_modelID.id)`, it seems to me the call here should be replaced to match that behavior as well (using id instead of identity)
```suggestion
  this->worlds.at(this->models.at(_modelID.id)->world).id->world->removeMultiBody(this->models.at(_modelID.id)->model);
```

Blast545: Although the key access of the unordered map, are `size_t` and internally the compiler is making the conversion. Maybe just remove the `id` in line 36 to have them using Identity

bullet/src/EntityManagementFeatures.cc
```diff
@@ -33,31 +33,28 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
 bool EntityManagementFeatures::RemoveModel(const Identity &_modelID)
 {
   // Check if the model exists
-  if (this->models.find(_modelID) == this->models.end()){
+  if (this->models.find(_modelID.id) == this->models.end()){
```
Blast545, line 36: Just leave as it was
```suggestion
  if (this->models.find(_modelID) == this->models.end()){
```

Blast545 (approved):

bullet/src/EntityManagementFeatures.cc
Blast545: (disregard this comment in favor to the suggestion for line 36)

bullet/src/EntityManagementFeatures.cc
Lobotuerk: Cant, Identity has no operator ==

bullet/src/EntityManagementFeatures.cc
Lobotuerk: Updated on 205fcc8, great catch

Blast545 (approved):
LGTM

PR #162, opened 2020-11-14
API documentation common Get* feature

anindex:
From issue #9


include/ignition/physics/GetBoundingBox.hh
```diff
@@ -62,6 +66,10 @@ namespace ignition
       GetLinkBoundingBox,
       GetLinkFromModel,
       ModelFrameSemantics>;
+
+    /////////////////////////////////////////////////
+    /// \brief This feature retrieves the axis aligned bounding box for the
+    /// links attached to this Model in the requested frame.
```
claireyywang, line 72:
```suggestion
    /// links attached to this model in the requested frame.
```

include/ignition/physics/GetEntities.hh
```diff
@@ -51,6 +53,9 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature retrieves the index-specified World pointer from the
```
claireyywang, line 57: I think we can just use lower case here instead of referring to the class name. Same for all the documentation below.
```suggestion
    /// \brief This feature retrieves the index-specified world pointer from the
```

include/ignition/physics/GetEntities.hh
```diff
@@ -128,6 +133,9 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature retrieves the index-specified Model pointer from the
+    /// index-specified simulation World.
```
claireyywang, line 138: `index-specified` sounds awkward. How about 
```suggestion
    /// \brief This feature retrieves model pointer from simulation World using index
```
?

include/ignition/physics/Joint.hh
```diff
@@ -26,6 +26,10 @@ namespace ignition
   namespace physics
   {
     /////////////////////////////////////////////////
+    /// \brief This feature retrieves the generalized Joint states such as
+    /// Position, Velocity, Acceleration of the Joint, applied Force to the
+    /// Joint and the transformation matrix from the Joint's parent Link to this
+    /// Joint's child Link based on its current position.
```
claireyywang, line 32: We can just use lower case to refer to the class names to keep documentation consistent.

claireyywang (changes requested):
Looks good in general. I just had some minor comments about format and wording. I noticed you borrowed some definition and wording from existing API documentation, which is helpful but not great reference. It'd be better for our new API documentation to improve based on that effort rather than keeping it.

include/ignition/physics/GetEntities.hh
scpeters, line 57: this feature has API's for getting a world pointer by index or by name, but this feature-level description only says "index-specified"

claireyywang (approved):

PR #163, opened 2020-11-14
API documentation common Attach* features

anindex:
Signed-off-by: anindex <an.thai.le97@gmail.com>

From issue #9

### DIFF
--- a/include/ignition/physics/BoxShape.hh
+++ b/include/ignition/physics/BoxShape.hh
@@ -81,8 +81,9 @@ namespace ignition
     };
 
     /////////////////////////////////////////////////
-    /// \brief This feature rigidly attachs a BoxShape to the specified
-    /// Link.
+    /// \brief This feature constructs a new box shape and attaches the
+    /// desired pose in the link frame. The pose could be defined to be the box
+    /// center point or any corner in actual implementation.
     class IGNITION_PHYSICS_VISIBLE AttachBoxShapeFeature
         : public virtual FeatureWithRequirements<BoxShapeCast>
     {
--- a/include/ignition/physics/BoxShape.hh
+++ b/include/ignition/physics/BoxShape.hh
@@ -80,6 +80,9 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature rigidly attachs a BoxShape to the specified
+    /// Link.
     class IGNITION_PHYSICS_VISIBLE AttachBoxShapeFeature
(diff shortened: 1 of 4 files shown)


include/ignition/physics/BoxShape.hh
```diff
@@ -80,6 +80,9 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature rigidly attachs a BoxShape to the specified
+    /// Link.
```
claireyywang, line 85: Maybe we can elaborate on "rigidly attach a BoxShape" here, something like "This feature constructs a new box shape, and attaches the pose of the box shape relative to the link frame."

include/ignition/physics/CylinderShape.hh
```diff
@@ -92,6 +92,9 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature rigidly attachs a CylinderShape to the specified
```
claireyywang, line 96: same as above

include/ignition/physics/PlaneShape.hh
```diff
@@ -115,6 +115,8 @@ namespace physics
   };
 
   /////////////////////////////////////////////////
+  /// \brief This feature rigidly attachs a PlaneShape to the specified
```
claireyywang, line 118: same

include/ignition/physics/PlaneShape.hh
```diff
@@ -132,8 +134,8 @@ namespace physics
       ///   Name to give to the PlaneShape
       /// \param[in] _normal
       ///   Normal vector for the plane
-      /// \param[in] _offset
-      ///   Offset of the plane
+      /// \param[in] _point
+      ///   The point that the plane passes through (hence defining the plane)
```
claireyywang, line 138: Is this point default to the origin `Point::Zero`?

claireyywang (changes requested):
Some definitions need more clarification, but otherwise look good to me

include/ignition/physics/PlaneShape.hh
anindex: Yea, I think the plane will pass through point zero as default.

claireyywang:
```suggestion
      ///   The point that the plane passes through (the origin by default)
```

claireyywang (commented):
LGTM with a tad bit more detail

claireyywang (approved):

PR #167, opened 2020-11-20
[Bullet] Bullet classic, Add sdf entities

Blast545:
Changing internal base representation, and adding features to add/remove links


Blast545:
@Lobotuerk With 37ab82d I added the auxiliar vector we discussed as a nive to have, to save IDs of links in a model, but didn't update the removeModel and similar methods to use it yet.

Lobotuerk (approved):
LGTM

PR #169, opened 2020-11-24
Kinematics features

Lobotuerk:
Signed-off-by: Lobotuerk <jtlorente@ekumenlabs.com>

### DIFF
--- a/bullet/src/KinematicsFeatures.cc
+++ b/bullet/src/KinematicsFeatures.cc
@@ -1,0 +1,98 @@
+#include <ignition/common/Console.hh>
+#include "KinematicsFeatures.hh"
+
+namespace ignition {
+namespace physics {
+namespace bullet {
+
+/////////////////////////////////////////////////
+FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
+    const FrameID &_id) const
+{
+  FrameData3d data;
+
+  // The feature system should never send us the world ID.
+  if (_id.IsWorld())
+  {
+    ignerr << "Given a FrameID belonging to the world. This should not be "
+           << "possible! Please report this bug!\n";
+    assert(false);
+    return data;
+  }
+
+  const auto linkID = _id.ID();
+  const auto &linkInfo = this->links.at(linkID);
+  const auto modelIdentity = linkInfo->model;
+  // auto linkIndex = linkInfo->linkIndex;
+  const auto &modelInfo = this->models.at(modelIdentity);
+  const auto &model = modelInfo->model;
+
+  const btVector3 pos = model->getCenterOfMassTransform().getOrigin();
+  const btMatrix3x3 mat = model->getCenterOfMassTransform().getBasis();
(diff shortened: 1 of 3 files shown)


bullet/src/KinematicsFeatures.cc
```diff
@@ -0,0 +1,98 @@
...
+    assert(false);
+    return data;
+  }
+
+  const auto linkID = _id.ID();
+  const auto &linkInfo = this->links.at(linkID);
+  const auto modelIdentity = linkInfo->model;
+  // auto linkIndex = linkInfo->linkIndex;
+  const auto &modelInfo = this->models.at(modelIdentity);
+  const auto &model = modelInfo->model;
```
Blast545, line 28: `modelInfo->model` does not exist anymore, probably you meant `linkInfo->model`.

bullet/src/KinematicsFeatures.cc
```diff
@@ -0,0 +1,98 @@
...
+  const btVector3 pos = model->getCenterOfMassTransform().getOrigin();
+  const btMatrix3x3 mat = model->getCenterOfMassTransform().getBasis();
+
+  auto eigenMat = convert(mat);
+  auto eigenVec = convert(pos);
+
+  data.pose.linear() = eigenMat;
+  data.pose.translation() = eigenVec;
+
+  // Get frame velocities
```
Blast545, line 39: Those comments can be removed, as those are already in the previous implementation in the case we ever need them.

bullet/src/KinematicsFeatures.cc
```diff
@@ -0,0 +1,98 @@
...
+  if (_id.IsWorld())
+  {
+    ignerr << "Given a FrameID belonging to the world. This should not be "
+           << "possible! Please report this bug!\n";
+    assert(false);
+    return data;
+  }
+
+  const auto linkID = _id.ID();
+  const auto &linkInfo = this->links.at(linkID);
```
Blast545, line 24: If we will the _id, which is a FrameID as a replacement to query the linkIDs I think we should something to exit the function cleanly in the case this function receives a not valid linkID.

I won't block this PR on that, but you can add a comment about it here so we remember later.

Blast545 (changes requested):
Minor changes

bullet/src/KinematicsFeatures.cc
Lobotuerk: Added on 25225aa

bullet/src/KinematicsFeatures.cc
Lobotuerk: Removed on 25225aa

bullet/src/KinematicsFeatures.cc
Lobotuerk: Fixed on 25225aa

Blast545 (approved):

Lobotuerk:
Merging

PR #171, opened 2020-11-25
Free Group Feature

Lobotuerk:
Signed-off-by: Lobotuerk <jtlorente@ekumenlabs.com>

### DIFF
--- a/bullet/src/Base.hh
+++ b/bullet/src/Base.hh
@@ -56,7 +56,8 @@ struct ModelInfo
 {
   std::string name;
   Identity world;
-  std::vector<std::size_t> links = {};
+  bool fixed;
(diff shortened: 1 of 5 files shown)


bullet/src/Base.hh
```diff
@@ -56,7 +56,8 @@ struct ModelInfo
 {
   std::string name;
   Identity world;
-  std::vector<std::size_t> links = {};
+  bool fixed;
```
Blast545, line 59: If the models are entities on its own, shouldn't be a boolean like this be on top of the links instead? Can models be composed of fixed links + not fixed ones?

Blast545 (approved):
Minor question, otherwise LGTM

PR #172, opened 2020-12-01
[bullet] Basic Features

Lobotuerk:
Signed-off-by: Tomas Lorente <jtlorente@ekumenlabs.com>
-Fixed position by adding the base model position into the equiation.
-Fixed gravity by adding an empty collisionshape into the bodies.
-Added a motion state to get the position out of bullet.
-Changed some autos into their proper variable.

### DIFF
--- a/bullet/src/SDFFeatures.cc
+++ b/bullet/src/SDFFeatures.cc
@@ -59,7 +59,6 @@ Identity SDFFeatures::ConstructSdfLink(
   const btVector3 linkInertiaDiag =
       convertVec(ignition::math::eigen3::convert(diagonalMoments));
 
-  // (TO-DO: do those calculation need to take into account in some way the base model?)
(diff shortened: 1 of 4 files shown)


bullet/src/SDFFeatures.cc
```diff
@@ -60,17 +60,20 @@ Identity SDFFeatures::ConstructSdfLink(
       convertVec(ignition::math::eigen3::convert(diagonalMoments));
 
   // (TO-DO: do those calculation need to take into account in some way the base model?)
```
Blast545, line 62: Can you remove this comment? I think you addressed this out here

Blast545 (approved):
LGTM, minor comment

bullet/src/SDFFeatures.cc
Lobotuerk: Done on a4c0c69

Lobotuerk:
Merging after approval

PR #175, opened 2020-12-09
[bullet] Add find or construct link

Blast545:
This is to create links in a way similar to dartsim. I think this probably may avoid issues with duplicated links. As bullet doesn't save the names of the links, the names are searched in the `links` unordered map.

### DIFF
--- a/bullet/src/SDFFeatures.cc
+++ b/bullet/src/SDFFeatures.cc
@@ -173,7 +173,7 @@ Identity SDFFeatures::ConstructSdfCollision(
 }
 
 /////////////////////////////////////////////////
-Identity SDFFeatures::FindOrConstructSdfLink(
+std::size_t SDFFeatures::FindOrConstructSdfLink(
   const Identity &_modelID,
   const ::sdf::Link &_sdfLink)
 {
@@ -186,8 +186,8 @@ Identity SDFFeatures::FindOrConstructSdfLink(
     if (linkInfo->name == linkName)
     {
       // A link was previously created with that name,
-      // Return an identity with it
-      return Identity(link.first, link.second);
+      // Return its entity value
+      return link.first;
     }
   }
   return this->ConstructSdfLink(_modelID, _sdfLink);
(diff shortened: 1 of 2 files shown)


bullet/src/SDFFeatures.cc
```diff
@@ -42,7 +42,7 @@ Identity SDFFeatures::ConstructSdfModel(
   // Build links
   for (std::size_t i = 0; i < _sdfModel.LinkCount(); ++i)
   {
-    this->ConstructSdfLink(modelIdentity, *_sdfModel.LinkByIndex(i));
+    this->FindOrConstructSdfLink(modelIdentity, *_sdfModel.LinkByIndex(i));
```
Lobotuerk, line 45: I have confirmed this loop is never entered, as models get constructed without links. We should delete it to save misunderstandings

bullet/src/SDFFeatures.cc
```diff
@@ -172,6 +172,27 @@ Identity SDFFeatures::ConstructSdfCollision(
   return this->GenerateInvalidId();
 }
 
+/////////////////////////////////////////////////
+std::size_t SDFFeatures::FindOrConstructSdfLink(
```
Lobotuerk, line 176: Is this a physics plugin function, or a custom one for the engine? Who consumes this one?

Lobotuerk (changes requested):

bullet/src/SDFFeatures.cc
Blast545: I tested removing it and `shapes.sdf` still works. So I removed it with b8758a8. 

This looks odd to me, at least tthrough the whole `dartsim` implementation they seemed to create the links there.

bullet/src/SDFFeatures.cc
Blast545: It is part of the plugin, but not part of the default functions consumed by ignition. It's consumed internally to create links avoiding duplicates if the link is already created. I had it inside the createModel function (The one you suggested me to erase) and I plan to use it inside createJoint to get the entities of the links involved in the joint creation.

Lobotuerk (approved):

PR #178, opened 2020-12-14
[Bullet] Mesh features

Lobotuerk:
Mesh collision seems to be working, at least with the duck.dae model.
ezgif-3-f39e68b5f641

### DIFF
--- a/bullet/src/ShapeFeatures.cc
+++ b/bullet/src/ShapeFeatures.cc
@@ -100,7 +100,7 @@ Identity ShapeFeatures::AttachMeshShape(
 Identity ShapeFeatures::CastToMeshShape(
     const Identity &_shapeID) const
 {
-  ignerr << "CASTTOMESHSHAPE" << std::endl;
+  ignwarn << "Dummy implementation CastToMeshShape" << std::endl;
   (void) _shapeID;
   return this->GenerateInvalidId();
 }
--- a/bullet/src/ShapeFeatures.cc
+++ b/bullet/src/ShapeFeatures.cc
@@ -1,0 +1,102 @@
+#include "ShapeFeatures.hh"
+#include <BulletCollision/Gimpact/btGImpactShape.h>
+#include <BulletCollision/Gimpact/btGImpactCollisionAlgorithm.h>
+
+namespace ignition {
+namespace physics {
+namespace bullet {
+
+/////////////////////////////////////////////////
+Identity ShapeFeatures::AttachMeshShape(
+    const Identity &_linkID,
+    const std::string &_name,
+    const ignition::common::Mesh &_mesh,
+    const Pose3d &_pose,
+    const LinearVector3d &_scale)
+{
+  (void) _pose;
+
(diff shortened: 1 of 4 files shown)


bullet/src/ShapeFeatures.cc
```diff
@@ -0,0 +1,110 @@
...
+                  vertices[indices[j+2]*3+2]);
+
+    mTriMesh->addTriangle(bv0, bv1, bv2);
+  }
+
+  btGImpactMeshShape *gimpactMeshShape =
+    new btGImpactMeshShape(mTriMesh);
+  gimpactMeshShape->updateBound();
+
+  // TODO(lobotuerk) Save collision if needed
```
Blast545, line 75: Isnt it neeeded to remove collision at the exit of the code?

bullet/src/ShapeFeatures.cc
```diff
@@ -0,0 +1,110 @@
...
+  const auto &modelInfo = this->models.at(modelID);
+  const auto &link = linkInfo->link;
+  const auto &worldInfo = this->worlds.at(modelInfo->world);
+  const auto &world = worldInfo->world;
+  delete link->getCollisionShape();
+  link->setCollisionShape(gimpactMeshShape);
+
+  btGImpactCollisionAlgorithm::registerAlgorithm(worldInfo->dispatcher);
+
+  world->addRigidBody(link);
```
Blast545, line 92: Isn't it needed to remove the link once the collision is updated? I saw you removed its collision, I imagine you shouldn't need to add the link again (or if needed, remove it first)

bullet/src/ShapeFeatures.cc
```diff
@@ -0,0 +1,110 @@
...
+  return this->AddCollision({_name, gimpactMeshShape, _linkID,
+                             modelID});
+
+}
+
+/////////////////////////////////////////////////
+Identity ShapeFeatures::CastToMeshShape(
+    const Identity &_shapeID) const
+{
+  ignerr << "CASTTOMESHSHAPE" << std::endl;
```
Blast545, line 103:
```suggestion
  ignwarn << "Dummy implementation CastToMeshShape" << std::endl;
```

Blast545 (approved):
Minor comments

bullet/src/ShapeFeatures.cc
Lobotuerk: You can get the collision pointer from the rigid body

bullet/src/ShapeFeatures.cc
Lobotuerk: Its only added here or in collision if the object has no mesh. So its only added on the last step

Blast545: This: `world->addRigidBody(link);` doesnt add a link already in the world for a second time?, not sure if I'm missing something

Lobotuerk: If the model has a mesh, it does not call sdfConstructCollision

PR #179, opened 2020-12-17
[TPE] Update link pose

claireyywang:
Support updating individual link pose, related to https://github.com/ignitionrobotics/ign-gazebo/pull/427

### DIFF
--- a/tpe/lib/src/Link.cc
+++ b/tpe/lib/src/Link.cc
@@ -43,19 +43,19 @@ Entity &Link::AddCollision()
   return *it->second.get();
 }
 
-//////////////////////////////////////////////////
-void Link::UpdatePose(
-  const double _timeStep,
-  const math::Vector3d _linearVelocity,
-  const math::Vector3d _angularVelocity)
-{
-  if (_linearVelocity == math::Vector3d::Zero &&
-      _angularVelocity == math::Vector3d::Zero)
-    return;
+// //////////////////////////////////////////////////
+// void Link::UpdatePose(
+//   const double _timeStep,
+//   const math::Vector3d _linearVelocity,
+//   const math::Vector3d _angularVelocity)
+// {
+//   if (_linearVelocity == math::Vector3d::Zero &&
+//       _angularVelocity == math::Vector3d::Zero)
+//     return;
 
-  math::Pose3d currentPose = this->GetPose();
-  math::Pose3d nextPose(
-    currentPose.Pos() + _linearVelocity * _timeStep,
-    currentPose.Rot().Integrate(_angularVelocity, _timeStep));
-  this->SetPose(nextPose);
-}
+//   math::Pose3d currentPose = this->GetPose();
+//   math::Pose3d nextPose(
+//     currentPose.Pos() + _linearVelocity * _timeStep,
+//     currentPose.Rot().Integrate(_angularVelocity, _timeStep));
+//   this->SetPose(nextPose);
+// }
--- a/tpe/lib/src/Link.cc
+++ b/tpe/lib/src/Link.cc
@@ -44,7 +44,7 @@ Entity &Link::AddCollision()
 }
 
 //////////////////////////////////////////////////
-void Link::SetLinearVelocity(const &math::Vector3d _velocity)
+void Link::SetLinearVelocity(const math::Vector3d &_velocity)
 {
   this->linearVelocity = _velocity;
 }
@@ -56,7 +56,7 @@ math::Vector3d Link::GetLinearVelocity() const
 }
 
 //////////////////////////////////////////////////
-void Link::SetAngularVelocity(const &math::Vector3d _velocity)
+void Link::SetAngularVelocity(const math::Vector3d &_velocity)
 {
   this->angularVelocity = _velocity;
 }
@@ -70,8 +70,8 @@ math::Vector3d Link::GetAngularVelocity() const
 //////////////////////////////////////////////////
 void Link::UpdatePose(
   const double _timeStep,
-  const math::Vector3d _linearVelocity,
-  const math::Vector3d _angularVelocity)
+  const math::Vector3d &_linearVelocity,
+  const math::Vector3d &_angularVelocity)
 {
   if (_linearVelocity == math::Vector3d::Zero &&
       _angularVelocity == math::Vector3d::Zero)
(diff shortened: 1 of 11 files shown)


tpe/lib/src/Link.cc
```diff
@@ -42,3 +42,44 @@ Entity &Link::AddCollision()
   this->ChildrenChanged();
   return *it->second.get();
 }
+
+//////////////////////////////////////////////////
+void Link::SetLinearVelocity(const math::Vector3d _velocity)
```
iche033, line 47: let's pass by reference. I noticed that it's the same in Model.hh but having `const &` is more efficient.

Same for `SetAngularVelocity`

iche033 (commented):
looks good. Can you add some unit tests for the new functions in Link?

tpe/lib/src/Link.cc
mxgrey: Weirdly, I've read that it isn't always the case that a const reference is more efficient for small POD statically sized data structures because of cache behaviors. Copying the three `double` values in the `Vector3d` into the stack frame may have less overhead than repeatedly dereferencing the `const Vector3d&` which is allocated elsewhere, perhaps in the heap.

That's not say we shouldn't use const-reference here. In the absence of profiling, it's fair to assume that `const T&` will be more efficient than `const T` for non-primitive types. I just thought it's interesting that the common wisdom doesn't always hold.

claireyywang:
@iche033  I think this one is ready for another look as well :D

luca-della-vedova:
We tried the PR with TPE and thought I'd check with you how you think we should model transformation when acting on links and what frame or reference commands should use.
We tried a model with two links and noticed that when we apply a link velocity command the behavior changes depending on the link it is applied upon. Specifically int the example below we are applying a constant velocity and when clicking the button we apply an angular velocity (to simulate opening the door).

https://user-images.githubusercontent.com/9111558/103064036-ea92ff80-45ed-11eb-93a0-19b9a0f23d3c.mp4

When applying the angular velocity on one of the links the model coordinates are also rotated, hence the two links start moving in the opposite direction (since they both rotate by 90 degrees in opposite directions).
I am not sure what behavior would make more sense for link level commands, if the current one is OK (and we should add a "dummy" link that represents a static dummy base) or if we should change it so that any Link level command has no effect on the model coordinates itself.
What do you think?

iche033:
fixed velocity cmd frame of reference in ba45184

iche033:
> We tried a model with two links and noticed that when we apply a link velocity command the behavior changes depending on the link it is applied upon.

Gazebo has a concept of a canonical link which is always fixed to the model. It's usually the first link in the model if not specified through SDF.  So I think what's happening is that in your example, one of the links is the canonical link and setting vel of that link may have caused the pose changes to affect the whole model.  I think the workaround is to create a base link that is always fixed (e.g. door frame?), and use the velocity control system on two other door links. 

As for the reference frame, I've made some fixes in ba45184, the velocity should now be in link frame (it was in world frame before).

@luca-della-vedova  Can you give https://github.com/ignitionrobotics/ign-gazebo/pull/427 and this PR a try and see if it'll produce the correct behavior now?

tpe/lib/src/Link.hh
```diff
@@ -42,6 +44,39 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Link : public Entity
...
+  /// \brief Set the angular velocity of link
+  /// \param[in] _velocity angular velocity
+  public: void SetAngularVelocity(const math::Vector3d &_velocity);
+
+  /// \brief Get the angular velocity of link
+  /// \return angular velocity
+  public: math::Vector3d GetAngularVelocity() const;
+
+  /// \brief Update the pose of the entity
+  /// \param[in] _timeStep current world timestep
```
chapulina, line 65: It would be nice to document the unit. Alternatively, we could use chrono so there's no ambiguity.

tpe/lib/src/Link_TEST.cc
```diff
@@ -48,6 +48,12 @@ TEST(Link, BasicAPI)
   EXPECT_EQ(math::Pose3d(10, -0.312675, 2.43845, 1.23686, 0.471978, 0.918989),
             linkEnt.GetWorldPose());
 
+  link.SetLinearVelocity(math::Vector3d(0, 0.1, 0));
+  EXPECT_EQ(math::Vector3d(0, 0.1, 0), link.GetLinearVelocity());
+
+  link.SetAngularVelocity(math::Vector3d(0.2, 0, 1));
+  EXPECT_EQ(math::Vector3d(0.2, 0, 1), link.GetAngularVelocity());
+
```
chapulina, line 56: Can we add some tests for `UpdatePose`?

tpe/lib/src/Link.cc
```diff
@@ -42,3 +42,44 @@ Entity &Link::AddCollision()
...
+math::Vector3d Link::GetAngularVelocity() const
+{
+  return this->angularVelocity;
+}
+
+//////////////////////////////////////////////////
+void Link::UpdatePose(
+  const double _timeStep,
+  const math::Vector3d &_linearVelocity,
+  const math::Vector3d &_angularVelocity)
```
chapulina, line 74: Is there a use case for calling this function with velocities different from the ones internal to the link? Shouldn't we be using `this->linearVelocity` and `this->angularVelocity` here instead of passing them in? Otherwise these values are never used by the link itself.

tpe/lib/src/Link.hh
```diff
@@ -42,6 +44,39 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Link : public Entity
   /// \brief Add a collision
   /// \return Newly created Collision
   public: Entity &AddCollision();
+
+  /// \brief Set the linear velocity of link
+  /// \param[in] _velocity linear velocity
+  public: void SetLinearVelocity(const math::Vector3d &_velocity);
```
chapulina, line 50: It would be nice to document the unit and frame of these velocities.

iche033: updated doc in 58f3e0b and ab04c90

tpe/lib/src/Link.cc
iche033: don't think there is a use case for it so removed `_linearVelocity` and `_angularVelocity` args. 58f3e0b

tpe/lib/src/Link_TEST.cc
iche033: added test. 58f3e0b

tpe/lib/src/Link.hh
iche033: added unit. 58f3e0b

chapulina (approved):

PR #181, opened 2020-12-18
TPE: Skip computing collisions for static objects

luca-della-vedova:
This PR introduces skipping collision checking when going through the AABB tree for static objects.
This change allows large performance gains in worlds with a lot of static objects but very few non static ones.

Apart from some complex test scenarios we are working on (where I observed about 300% performance gain) I went for a pathological case and generated a [shapes_population](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/examples/worlds/shapes_population.sdf.erb) with a total of 1500 shapes (500 per type), all set to static except one and I had about double the performance (12% RTF before, 24% RTF after).

I'm considering adding caching to the CollisionDetector where, if an object didn't move since the previous time step, its collision detection step will be skipped. This would work for most scenarios where a second non static object is moving but would fail if a static object was placed (i.e. "teleported") to collide with the static but not moving object (hope this makes sense!).
This should allow further speedups in worlds with a lot of performers but only a few of them moving at the same time but I'll probably address it in a future PR if the performance gain is significant.

### DIFF
--- a/tpe/lib/src/CollisionDetector.cc
+++ b/tpe/lib/src/CollisionDetector.cc
@@ -125,6 +125,9 @@ std::vector<Contact> CollisionDetector::CheckCollisions(
   for (auto it = _entities.begin(); it != _entities.end(); ++it)
   {
     std::shared_ptr<Entity> e = it->second;
+    // Skip if the entity is static
+    if (e->GetStatic() == true)
+      continue;
 
     math::AxisAlignedBox b = e->GetBoundingBox();
     if (b == math::AxisAlignedBox())
--- a/tpe/lib/src/CollisionDetector.cc
+++ b/tpe/lib/src/CollisionDetector.cc
@@ -126,7 +126,7 @@ std::vector<Contact> CollisionDetector::CheckCollisions(
   {
     std::shared_ptr<Entity> e = it->second;
     // Skip if the entity is static
-    if (e->GetStatic() == true)
+    if (e->GetStatic())
       continue;
 
     math::AxisAlignedBox b = e->GetBoundingBox();
--- a/tpe/lib/src/CollisionDetector_TEST.cc
+++ b/tpe/lib/src/CollisionDetector_TEST.cc
@@ -241,3 +241,44 @@ TEST(CollisionDetector, CheckCollisions)
   contacts = cd.CheckCollisions(entities, true);
   EXPECT_EQ(1u, contacts.size());
 }
+
+/////////////////////////////////////////////////
+TEST(CollisionDetector, CheckStaticCollisionFiltering)
+{
+  // set up entities for testing collision filtering between static objects
+
+  // model A
+  std::shared_ptr<Model> modelA(new Model);
+  modelA->SetStatic(true);
+  Entity &linkAEnt = modelA->AddLink();
+  Link *linkA = static_cast<Link *>(&linkAEnt);
+  Entity &collisionAEnt = linkA->AddCollision();
+  Collision *collisionA = static_cast<Collision *>(&collisionAEnt);
+  BoxShape boxShapeA;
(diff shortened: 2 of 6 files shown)


iche033 (commented):
nice! Just have small minor comments. I think we can also backport this to ign-physics2 for Citadel.

iche033:
> I'm considering adding caching to the CollisionDetector where, if an object didn't move since the previous time step, its collision detection step will be skipped. 

yeah that makes sense. For caching, you mean caching the contact points? You could check if either of the entity's pose is dirty or not like [here](https://github.com/ignitionrobotics/ign-physics/blob/6736e0debb6f1b667d49c0c7f61cf1b8c632cc6d/tpe/lib/src/CollisionDetector.cc#L109). So something like skipping collision check if both entities are static and neither of their poses are dirty?

luca-della-vedova:
> nice! Just have small minor comments. I think we can also backport this to ign-physics2 for Citadel.

Ah I always forget the camelCase for variable names :sweat_smile:, addressed all in [1513ba6](https://github.com/ignitionrobotics/ign-physics/pull/181/commits/1513ba6318c009d12ba33842a1d3e02383efaefb)
For the porting sure, I have at least a small fix and the big one below lef to do. Maybe I'll just get everything into Dome and open a PR cherry-picking all of them at once? Otherwise I can target Citadel moving forward and we can do the usual forward merge.

> yeah that makes sense. For caching, you mean caching the contact points? You could check if either of the entity's pose is dirty or not like here. So something like skipping collision check if both entities are static and neither of their poses are dirty?

If both entities are static the collision check would already be skipped by this PR. If this gets merged we would only be iterating over non static objects (from now on performers). I was thinking of caching the contact points for non static objects and return them straight away if the performer object didn't have its pose updated in the latest iteration. If I understand correctly the following behavior would change:
* If the performer has a static object spawned or teleported to collide, it will now _not_ be reported until the performer tries to move and updates its pose. This is probably not super ideal but it is probably an obscure enough case?
* If a moving performer collides with a non moving performer, the non moving performer will _not_ report a collision, but from what I understand in the normal case this would generate a duplicate collision that is filtered by [CheckDuplicateCollisionPair](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics3/tpe/lib/src/CollisionDetector.cc#L245) so it might be even a desirable side effect.

But again, I'll be looking at the performance gains and make sure they are relevant enough for this to make sense.

iche033:
> Maybe I'll just get everything into Dome and open a PR cherry-picking all of them at once? Otherwise I can target Citadel moving forward and we can do the usual forward merge.

So far we've been trying to target the earliest ignition version as much as possible. Backporting by cherry picking also works.

> If both entities are static the collision check would already be skipped by this PR.

I see.  I just checked gazebo-classic and noticed that we don't report contacts for static-static objects.. If we want, we can do something like changing the check to :

```
if (e->GetStatic() && !e->PoseDirty())
```
The downside is that it'll only report contacts for only one iteration, which may be enough for some applications. In any case, since this would be a behavior change from a released version, we should document this in Migration.md.

> I was thinking of caching the contact points for non static objects and return them straight away if the performer object didn't have its pose updated in the latest iteration.

sounds good.

> If the performer has a static object spawned or teleported to collide, it will now not be reported until the performer tries to move and updates its pose. 

yep makes sense.
 
> If a moving performer collides with a non moving performer, the non moving performer will not report a collision,

yep that sounds fine to me. Duplicate contacts are filtered anyway.

luca-della-vedova:
OK I changed the base branch to Citadel since it's a backwards compatible change.

iche033:
I added a small test for the new `SetStatic` API in 5308b6f. The changes look good to me!

iche033 (approved):

PR #182, opened 2020-12-21
Fix TPE poseDirty getter

luca-della-vedova:
Classic `=` vs `==` typo that wasn't caught at compile time because of the pimpl's raw pointer not enforcing const correctness.
It disabled the optimization that skipped updating the AABB tree for entities that didn't move, so the PR should help improve performance.

### DIFF
--- a/tpe/lib/src/Entity.cc
+++ b/tpe/lib/src/Entity.cc
@@ -319,7 +319,7 @@ Entity *Entity::GetParent() const
 //////////////////////////////////////////////////
 bool Entity::PoseDirty() const
 {
-  return this->dataPtr->poseDirty == true;
+  return this->dataPtr->poseDirty;
 }
 
 //////////////////////////////////////////////////
--- a/tpe/lib/src/Entity.cc
+++ b/tpe/lib/src/Entity.cc
@@ -319,7 +319,7 @@ Entity *Entity::GetParent() const
 //////////////////////////////////////////////////
 bool Entity::PoseDirty() const
 {
-  return this->dataPtr->poseDirty = true;
+  return this->dataPtr->poseDirty == true;
 }
 
 //////////////////////////////////////////////////
--- a/dartsim/src/EntityManagementFeatures.cc
+++ b/dartsim/src/EntityManagementFeatures.cc
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <string>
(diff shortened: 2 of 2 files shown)


mxgrey:
I think it's worth mentioning that there's no use for PIMPL in an `ignition-physics` physics engine plugin anyway.

We already get an API+ABI firewall for this library by loading it using `ignition-plugin`, so PIMPL doesn't accomplish anything.

iche033:
I think we did pimpl at the beginning because there was an idea that the TPE library (in `tpe/lib`) would live outside of ign-physics or could be migrated later if needed. But that's no longer the case as we're now committing to having the library inside ign-phyics so we could just stop using pimpl in future classes

tpe/lib/src/Entity.cc
```diff
@@ -319,7 +319,7 @@ Entity *Entity::GetParent() const
 //////////////////////////////////////////////////
 bool Entity::PoseDirty() const
 {
-  return this->dataPtr->poseDirty = true;
+  return this->dataPtr->poseDirty == true;
```
iche033, line 322: how about just:

```
  return this->dataPtr->poseDirty;
```

iche033 (commented):
bad mistake, thanks for catching this!

chapulina:
> there was an idea that the TPE library (in tpe/lib) would live outside of ign-physics or could be migrated later if needed. But that's no longer the case as we're now committing to having the library inside ign-phyics

Is anything is TPE so specific to Ignition Physics that it couldn't be used on its own?

iche033:
> Is anything is TPE so specific to Ignition Physics that it couldn't be used on its own?

nope, TPE is not specific to ign-physics. It's designed to be a standalone library.

scpeters:
fix for homebrew CI problems in https://github.com/ignition-tooling/release-tools/pull/366

tpe/lib/src/Entity.cc
luca-della-vedova: Done

chapulina (approved):
I took the liberty of addressing the codechecker warning on a4faa4b50679833880aaa0018237775fef258248 so we don't need a new PR, then a merge into this PR, etc.

LGTM with happy CI.

iche033 (approved):

PR #186, opened 2021-01-07
[Dome] Ign-physics tutorials and API doc back port

claireyywang:
Back port tutorials and API doc to citadel to show on website

### DIFF
--- a/include/ignition/physics/SphereShape.hh
+++ b/include/ignition/physics/SphereShape.hh
@@ -79,6 +79,9 @@ namespace ignition
     };
 
     /////////////////////////////////////////////////
+    /// \brief This feature constructs a new sphere shape and attaches the
+    /// desired pose in the link frame. The pose is defined as the
+    /// sphere center point in actual implementation.
     class IGNITION_PHYSICS_VISIBLE AttachSphereShapeFeature
         : public virtual FeatureWithRequirements<SphereShapeCast>
     {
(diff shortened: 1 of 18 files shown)


include/ignition/physics/SphereShape.hh
```diff
@@ -92,6 +95,12 @@ namespace ignition
 
         public: using ShapePtrType = SphereShapePtr<PolicyT, FeaturesT>;
 
+        /// \brief Rigidly attach a SphereShape to this link.
+        /// \param[in] _radius
```
chapulina, line 99: Just noticed `_name` is missing, but I think this is not introduced by the backport

tutorials/01_intro.md
```diff
@@ -2,18 +2,85 @@
...
+
+**Ignition Physics 2.x**
+- Support sdformat 1.7 frame semantics.
+- Support compiling against dart 6.9.
+- Trivial Physics Engine (TPE)- partial implementation
+- Add features for TPE physics engines (more detail in \ref physicsplugin).
+- Extend contact data with force, normal, and penetration depth.
+- Add Base and EntityManagement to `tpeplugin`
+
+**Ignition Physics 3.x**
```
chapulina, line 81: Maybe we could at least mention that it uses libSDFormat 10

tutorials/04-switching-physics-engines.md
```diff
@@ -0,0 +1,129 @@
+\page switchphysicsengines Switching physics engines
```
chapulina, line 1: This tutorial looks very similar to this one: https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/tutorials/physics.md

Since it's `ign-gazebo` specific, maybe we should just keep that one.

chapulina (approved):
LGTM! I just have some comments mainly because it's my first time looking at some of these tutorials. But I think the backport PR is probably not the best place to address those, so feel free to ignore them in this PR.

PR #188, opened 2021-01-07
[BULLET] Compound Collisions

Lobotuerk:


bullet/src/Base.hh
```diff
@@ -67,7 +67,7 @@ struct LinkInfo
   btRigidBody* link;
   Identity model;
   math::Pose3d pose;
-  btScalar mass;
+  double mass;
```
Blast545, line 70: Is this change needed?

Lobotuerk: It fixes some warnings at compilation

Blast545 (approved):
LGTM

PR #191, opened 2021-01-12
Bump in edifice: sdformat11

chapulina:
Addresses https://github.com/ignition-tooling/release-tools/issues/360

### DIFF
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -117,6 +117,20 @@ static void CopyStandardJointAxisProperties(
     const int _index, Properties &_properties,
     const ::sdf::JointAxis *_sdfAxis)
 {
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#endif
+  // TODO(anyone): stop using this deprecated function
+  _properties.mInitialPositions[_index] = _sdfAxis->InitialPosition();
+#ifndef _WIN32
+# pragma GCC diagnostic pop
+#else
+# pragma warning(pop)
+#endif
   _properties.mDampingCoefficients[_index] = _sdfAxis->Damping();
   _properties.mFrictions[_index] = _sdfAxis->Friction();
   _properties.mRestPositions[_index] = _sdfAxis->SpringReference();
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -117,7 +117,6 @@ static void CopyStandardJointAxisProperties(
     const int _index, Properties &_properties,
     const ::sdf::JointAxis *_sdfAxis)
 {
-  _properties.mInitialPositions[_index] = _sdfAxis->InitialPosition();
   _properties.mDampingCoefficients[_index] = _sdfAxis->Damping();
   _properties.mFrictions[_index] = _sdfAxis->Friction();
   _properties.mRestPositions[_index] = _sdfAxis->SpringReference();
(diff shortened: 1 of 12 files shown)


j-rivero (approved):

chapulina:
Trying a build with https://github.com/ignition-tooling/release-tools/pull/375 to see if all tests pass:

[Build Status](https://build.osrfoundation.org/job/ignition_physics-ci-pr_any-ubuntu_auto-amd64/1066/)

tutorials/06-physics-simulation-concepts.md
```diff
@@ -70,9 +70,9 @@ second in Z-axis.
 
 Note that the mechanism to move the car is different depending on the used physics
 engine. Using [dartsim](https://github.com/ignitionrobotics/ign-physics/tree/master/dartsim),
-the car is moved by applying force on the joints, please see [DiffDrive.cc](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/diff_drive/DiffDrive.cc#L333) code.
+the car is moved by applying force on the joints, please see [DiffDrive.cc](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/systems/diff_drive/DiffDrive.cc#L333) code.
 Using [TPE](https://github.com/ignitionrobotics/ign-physics/tree/main/tpe),
-TPE directly sets model velocity in [VelocityControl.cc](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/velocity_control/VelocityControl.cc#L117) code.
```
scpeters, line 75: note to self: check that these line numbers are still accurate

scpeters (commented):
I think we just need a prerelease or nightly of libsdformat11-dev for Ubuntu CI to work

dartsim/src/SDFFeatures.cc
```diff
@@ -117,7 +117,6 @@ static void CopyStandardJointAxisProperties(
     const int _index, Properties &_properties,
     const ::sdf::JointAxis *_sdfAxis)
 {
-  _properties.mInitialPositions[_index] = _sdfAxis->InitialPosition();
```
scpeters, line 120: this is deprecated but still functional; so I would keep this line and suppress its deprecation warnings (currently there are test failures because this was removed)

chapulina: The test has been fixed. I followed the approach being taken on https://github.com/ignitionrobotics/ign-physics/pull/189/.

I can suppress the warning instead if that's preferred. I'm not sure if there's a new approach to use or if this feature will just disappear when we use `SDF 12`. SDFormat's migration guide didn't list any replacements. Are we sure we want to remove the feature?

CC @azeey

scpeters: I think the replacement may be using the `<state>` element, but until we've articulated that I'd like to keep the feature working, since it is only deprecated at this point, not disabled. I'll add a comment to that effect on #189 as well.

scpeters: > I'll add a comment to that effect on #189 as well.

https://github.com/ignitionrobotics/ign-physics/pull/189/files#r558691120

scpeters: macro to suppress deprecation warnings in https://github.com/ignitionrobotics/ign-utils/pull/5

chapulina: Suppressing the warning in a more traditional style while this library doesn't depend on `ign-utils` yet. I'd like to merge this ASAP to unblock upstream.

35ce9275e3a60077dab35a5a61f8050478abac72

tutorials/06-physics-simulation-concepts.md
scpeters: they aren't, but I'll open an issue for that

scpeters (approved):

PR #192, opened 2021-01-13
[Bullet] Parameter tuning

Lobotuerk:


Lobotuerk:
Selection_124
After fixing the axis and setting properly the update world function, wobbling seems to be done. I was able to go all the way on that tunnel with just the go-forward command on a diff drive. Also added friction between joints that we were missing

Blast545 (approved):
LGTM

PR #197, opened 2021-01-15
[bullet] Getters for revolute joints

Blast545:
This PR addresses three things:

* Solves a problem with the axis of revolute joints.
* Adds support for fixed joints
* Add some basic getters for revolute joints

### DIFF
--- a/bullet/src/JointFeatures.cc
+++ b/bullet/src/JointFeatures.cc
@@ -27,20 +27,29 @@ namespace bullet {
 double JointFeatures::GetJointPosition(
     const Identity &_id, const std::size_t _dof) const
 {
-  (void) _id;
   (void) _dof;
-  double result = 0;
+  double result = ignition::math::NAN_D;
   if (this->joints.find(_id.id) != this->joints.end())
   {
-    btHingeAccumulatedAngleConstraint* hinge =
-      static_cast<btHingeAccumulatedAngleConstraint*>(this->joints.at(_id.id)->joint);
-    if (hinge)
-    {
-      result = hinge->getAccumulatedHingeAngle();
+    const JointInfoPtr &jointInfo = this->joints.at(_id.id);
+    const int jointType = jointInfo->constraintType;
+    // Check the type of joint and act accordignly
+    if (jointInfo->constraintType == static_cast<int>(::sdf::JointType::REVOLUTE)) {
+      btHingeAccumulatedAngleConstraint* hinge =
+	static_cast<btHingeAccumulatedAngleConstraint*>(jointInfo->joint);
+      if (hinge) {
+	result = hinge->getAccumulatedHingeAngle();
+	// result -= this->angleOffset;
+      }
+      else {
+	ignerr << "Corrupted joint at index:" << _id.id << "\n";
+      }
     }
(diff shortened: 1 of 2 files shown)


nkoenig:
I updated the formatting in 0b0b869187fcc135476bf62deb12491d13dc012d

Blast545:
@nkoenig Thanks for the commit! Did you use a particular linter or similar tool? At some point @Lobotuerk and I have to reformat most of the code in the `bullet_classic_dev` dev branch

bullet/src/SDFFeatures.cc
```diff
@@ -359,6 +359,13 @@ Identity SDFFeatures::ConstructSdfJoint(
     convertVec(ignition::math::eigen3::convert(axisChild)),
     convertVec(ignition::math::eigen3::convert(axisParent)));
 
+  // Limit movement for fixed joints
+  if(type == ::sdf::JointType::FIXED) {
+    btHingeConstraint *hingeJoint = dynamic_cast<btHingeConstraint *>(joint);
```
Lobotuerk, line 364: You don't need the dynamic cast, **joint** is a `btHingeContraint` at that point, just after retrieving it from the joints map you need the dynamic_cast

Blast545: You are right, I removed that cast with 72b656f

bullet/src/JointFeatures.cc
```diff
@@ -27,51 +27,206 @@ namespace bullet {
...
+      {
+        result = 0.0;
+        // Get the axis of the joint
+        btVector3 vec =
+          hinge->getRigidBodyA().getCenterOfMassTransform().getBasis() *
+          hinge->getFrameOffsetA().getBasis().getColumn(2);
+
+        math::Vector3 globalAxis(vec[0], vec[1], vec[2]);
+
+        if (this->links.find(jointInfo->childLinkId) != this->links.end())
```
Lobotuerk, line 86: I'm not sure about these verifications, if jointInfo has an invalid childLink, we should explode ASAP, as probably something went really wrong

Lobotuerk (approved):
Also, 3 getters look really alike. Are they a little different or are they in fact the same?

bullet/src/JointFeatures.cc
Blast545: These verifications check for joints with a parent/child link that doesn't have an extra child/parent link (attached to the world). The final angular velocity is the sum of the angular velocities of both links, the extra check is to see if it's attached to the world (although that's not really supported yet in constructionSdf)

Blast545:
@Lobotuerk The wrapping code is the same, but the specific getter functions are different. In particular, the force and acceleration is the same, it remains to divide that one over the link inertia and I marked it as a TO-DO in that section of the code.

bullet/src/JointFeatures.cc
Lobotuerk: Checking if childLinkId is in this->links means checking if the link exist as an entity. Joints with the world as a parent/child would have the world identity as the id. I think that checking if the id is the same as the world id would be the proper check, but I'm also fine with the answer

PR #202, opened 2021-01-22
Added ellipsoid and capsule to dartsim

ahcorde:
Added ellipsoid and capsule to dartsim

Related with https://github.com/ignitionrobotics/ign-physics/issues/196 and https://github.com/ignitionrobotics/ign-physics/issues/195


scpeters:
Ellipsoids are not supported natively by ODE or DART (they are in simbody)

In bullet, dart supports the EllipsoidShape by constructing a mesh:

* https://github.com/dartsim/dart/blob/master/dart/collision/bullet/BulletCollisionDetector.cpp#L529-L530

but not for ODE:

* https://github.com/dartsim/dart/blob/master/dart/collision/ode/OdeCollisionObject.cpp#L244-L253

we could copy the approach for bullet by creating a mesh based on the following code for creating meshes:

* https://github.com/dartsim/dart/blob/master/dart/collision/ode/detail/OdeMesh.cpp

should we consider requiring both the ODE and bullet components of dartsim?

cc @azeey

ahcorde:
I have prototyped a `CustomMeshShape` for the ellipsoid and the capsule using the generated mesh from the `common::MeshManager`

capsule_ellipsoid

Thoughts @azeey @scpeters ?

dartsim/src/ShapeFeatures.cc
```diff
@@ -154,6 +220,64 @@ Identity ShapeFeatures::AttachCylinderShape(
   return this->GenerateIdentity(shapeID, this->shapes.at(shapeID));
 }
 
+/////////////////////////////////////////////////
+Identity ShapeFeatures::CastToEllipsoidShape(const Identity &_shapeID) const
+{
+  std::cerr << "CastToEllipsoidShape" << '\n';
```
scpeters, line 226: nit: remove this line

dartsim/src/ShapeFeatures.cc
```diff
@@ -154,6 +220,64 @@ Identity ShapeFeatures::AttachCylinderShape(
...
+}
+
+/////////////////////////////////////////////////
+Identity ShapeFeatures::AttachEllipsoidShape(
+    const Identity &_linkID,
+    const std::string &_name,
+    const Vector3d _radii,
+    const Pose3d &_pose)
+{
+  ignerr << "AttachEllipsoidShape" << '\n';
```
scpeters, line 257: nit: remove this line

scpeters (commented):
I just took a quick look and made some very minor comments; I haven't tested it yet

dartsim/src/SDFFeatures.cc
```diff
@@ -315,8 +323,46 @@ static ShapeAndTransform ConstructGeometry(
...
+
+    auto mesh = std::make_shared<CustomMeshShape>(*_mesh, Vector3d(1, 1, 1));
+    auto mesh2 = std::dynamic_pointer_cast<dart::dynamics::MeshShape>(mesh);
+    return {mesh2};
+  }
   else if (_geometry.CylinderShape())
     return ConstructCylinder(*_geometry.CylinderShape());
+  else if (_geometry.EllipsoidShape())
+  {
+    // TODO(anyone): Replace this code when Ellipsoid is supported by DART
```
adlarkin, line 349: Similar to @azeey's comment about capsule support in DART, are ellipsoids supported in DART? https://dartsim.github.io/dart/v6.9.5/db/da3/classdart_1_1dynamics_1_1EllipsoidShape.html

ahcorde:
Using the `dart::dynamic::Capsule::Shape` I have this issue:

Selección_116


The semicircular part of the capsule is not colliding with the floor.

ahcorde:
with the ellipsoid I'm using meshes but the collisions are wrong too

Selección_117

@azeey or @adlarkin any thoughts ?

adlarkin:
> The semicircular part of the capsule is not colliding with the floor.
> 
> with the ellipsoid I'm using meshes but the collisions are wrong too

Can you share the SDF file(s) you are using for testing so that I can take a look?

ahcorde:
@adlarkin If you want to reproduce this issue. These PRs are related

 - https://github.com/ignitionrobotics/ign-rendering/pull/200
 - https://github.com/ignitionrobotics/ign-gazebo/pull/581/files

In the Gazebo draft PR you will find a world with the capsule and ellipsoid `shapes.sdf`.

chapulina (commented):
They're both working for me! I think the rendering just needs to be rotated:

rolling_shapes_new2

I'm wondering if there are more tests we can add here for this.

Also, I think we should make these features optional on `ign-gazebo`, so that we can support physics engines which don't have capsule or ellipsoid.

adlarkin:
> They're both working for me! I think the rendering just needs to be rotated for the capsule

I just took a look and agree with @chapulina; once I used https://github.com/ignitionrobotics/ign-common/pull/186, the capsule issue went away.

chapulina (approved):
Pushed some changes in 0162d54:

* Added a test
* Changed more `height` to `length`
* Better handled when a physics engine doesn't support a given geometry

Good to me with green CI!

dartsim/src/SDFFeatures.cc
```diff
@@ -273,6 +281,14 @@ static ShapeAndTransform ConstructSphere(
   return {std::make_shared<dart::dynamics::SphereShape>(_sphere.Radius())};
 }
 
+/////////////////////////////////////////////////
+static ShapeAndTransform ConstructCapsule(
+    const ::sdf::Capsule &_capsule)
+{
+  return {std::make_shared<dart::dynamics::CapsuleShape>(
+        _capsule.Radius(), _capsule.Length())};
```
chapulina, line 289: After playing with this a bit, I noticed that collisions between the capsule shape and a plane get unstable quickly. With 0.008 step size all other shapes collide reliably with a plane, but the capsule sinks :confused: The largest step size I could use without it sinking was 0.002.

adlarkin: Should we ticket an issue for this and come back to it after the Edifice release?

chapulina: I suspect this is a DART internal issue and the only thing we could do would be to use a mesh instead, like we're doing for ellipsoids. I'm inclined not to ticket an issue until this starts affecting users.

PR #203, opened 2021-01-22
Added Ellipsoid and Capsule shapes to TPE

ahcorde:
Added ellipsoid and capsule to TPE. This PR builds on top of this other PR https://github.com/ignitionrobotics/ign-physics/pull/202

Related with https://github.com/ignitionrobotics/ign-physics/issues/196 and https://github.com/ignitionrobotics/ign-physics/issues/195

It's a draft for now, I need to include some more tests.


ahcorde:
I opened this PR for review, I can add more tests if required

tpe/lib/src/Shape.hh
```diff
@@ -163,6 +212,38 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE CylinderShape : public Shape
...
+  /// \param[in] _other shape to copy from
+  public: Shape &operator=(const Shape &_other);
+
+  /// \brief Get ellipsoid radius
+  /// \return ellipsoid radius
+  public: math::Vector3d GetRadii() const;
+
+  /// \brief Set ellipsoid radius
+  /// \param[in] _radius ellipsoid radius
+  public: void SetRadii(math::Vector3d _radii);
```
chapulina, line 238:
```suggestion
  public: void SetRadii(const math::Vector3d &_radii);
```

tpe/plugin/worlds/shapes.world
```diff
@@ -96,5 +96,36 @@
         </visual>
       </link>
     </model>
+
+    <model name="ellipsoid">
```
chapulina, line 100: Should we add a capsule too?

chapulina (commented):
> It's a draft for now, I need to include some more tests.

I wasn't sure about the status of this PR, so I went ahead and gave it a quick review. It's looking good to me, I think it would be nice to add tests to `CollisionDetector_TEST`.

---

For future reference, I think it's also worth mentioning that TPE pretty much treats all shapes as boxes for the purpose of collision checking, because it only checks their AABBs. But Having these shapes supported means that worlds using complex shapes with other engines will work reasonably with TPE too.

chapulina:
The Homebrew and Ubuntu warnings are not introduced by this PR, but the Windows warnings are

ahcorde:
@chapulina fixed windows warnings

tpe/lib/src/CollisionDetector_TEST.cc
```diff
@@ -240,6 +297,39 @@ TEST(CollisionDetector, CheckCollisions)
   EXPECT_EQ(8u, contacts.size());
   contacts = cd.CheckCollisions(entities, true);
   EXPECT_EQ(1u, contacts.size());
+
+  entities[modelC->GetId()] = modelC;
+  // collision between model E and B but not model A, C, D
```
chapulina, line 302: Can you add a test case where `D`, the capsule, collides with something?

tpe/plugin/src/ShapeFeatures.cc
```diff
@@ -95,6 +95,83 @@ Identity ShapeFeatures::CastToCylinderShape(const Identity &_shapeID) const
...
+  // return invalid height if no collision found
+  return -1.0;
+}
+
+/////////////////////////////////////////////////
+Identity ShapeFeatures::AttachCapsuleShape(
+  const Identity &_linkID,
+  const std::string &_name,
+  const double _radius,
+  const double _height,
```
chapulina, line 154:
```suggestion
  const double _length,
```

chapulina (approved):
LGTM, I just have some minor final comments.

PR #204, opened 2021-01-22
[Citadel] Update tutorials

claireyywang:
- Use Doxygen code snippet and image 
- Add example code
- Update tutorial titles and content

Will forward port this to Dome and main if approved.


claireyywang:
@chapulina  I addressed your tutorial comments in this PR. Feel free to take a look if you have time .

tutorials/03_physics_plugins.md
```diff
@@ -1,8 +1,8 @@
-\page physicsplugin Understanding the Physics Plugin
+\page physicsplugins Understand Physics Plugin
```
chapulina, line 1: Note that changing the page name also changes the URL of the tutorial. Right now, the URL is https://ignitionrobotics.org/api/physics/3.1/physicsplugin.html. If we change them, anyone who had the old links on a website for example, will be left with broken links. So often it is better to stay with "bad" links that work for previous users, than to improve the URLs but break links.

These tutorials are relatively new though, so I think it's not too late to change them. I'll leave it up to you.

tutorials/02_installation.md
```diff
@@ -1,17 +1,17 @@
 \page installation Installation
 
 These instructions are for installing only Ignition Physics.
-If you're interested in using all the Ignition libraries, check out this [Ignition installation](https://ignitionrobotics.org/docs/latest/install).
+If you're interested in using all the Ignition libraries, check out this [Ignition installation](https://ignitionrobotics.org/docs/citadel/install).
 
-We recommend following the Binary Installation instructions to get up and running as quickly and painlessly as possible.
+We recommend following the [binary installation](#Binary-installation) instructions to get up and running as quickly and painlessly as possible.
```
chapulina, line 6: I think these links had been removed because now we have multiple sections for binary and source installation, according to the OS.

tutorials/02_installation.md
```diff
@@ -37,19 +37,16 @@ wget http://packages.osrfoundation.org/gazebo.key -O - | sudo apt-key add -
 On Ubuntu systems, `apt-get` can be used to install `ignition-plugin`:
 ```
 sudo apt-get update
-sudo apt-get install libignition-physics<#>-dev
+sudo apt-get install libignition-physics2-dev
 ```
 
-Be sure to replace `<#>` with a number value, such as `1` or `2`, depending on which version you need.
```
chapulina, line 43: I think it's ok to leave as it was, which is easier to maintain in the long run. I believe it was changed on purpose.

tutorials/04_physics_engines.md
```diff
@@ -0,0 +1,3 @@
+\page physicsengines "Use different physics engines"
+
+See [Physics engines](https://ignitionrobotics.org/api/gazebo/4.1/physics.html) in Ign-gazebo tutorial.
```
chapulina, line 3: This is targeting Citadel, so it should use `ign-gazebo3`

chapulina (commented):
Can we add a `CMakeLists` file to `examples/simple_tpe_plugin` so that it's possible to compile it?

claireyywang:
> Can we add a `CMakeLists` file to `examples/simple_tpe_plugin` so that it's possible to compile it?

I added an example `CMakeList` but don't think it's working correctly yet. I'm still not sure how exactly CMake links different libraries so it'd take some time to figure that out. I think it'd be preferable to merge the tutorial part of the this PR first, and take the CMake part out for later improvement. What do you think?

tutorials/09_use_custom_engine.md
```diff
@@ -0,0 +1,166 @@
...
+
+```
+ign-physics
+├── tpe
+│   ├── plugin                           Implementation of the plugin features interfacing the physics engines API
+│   │    ├── src
+│   │    |     ├── plugin.cc             Main file for the plugin declaration and plugin registering.
+│   │    |     ├── <FEATURES>.hh         The FeatureList header file.
+│   │    |     ├── <FEATURES>.cc         Implementation of the FeatureList using physics engine API
+│   │    |     ├── <FEATURES_TEST>.cc    Tests
```
chapulina, line 33:
```suggestion
│   │    |     └── <FEATURES_TEST>.cc    Tests
```

Nitpick :grin: The same on the next line.

tutorials/09_use_custom_engine.md
```diff
@@ -0,0 +1,166 @@
...
+└── CMakeList.txt                        CMake build script for Ignition Physics library.
+```
+
+We link the external physics engine library
+in `CMakeLists.txt` of the plugin, assuming the physics engine library is
+already installed. In our case, [TPE](https://github.com/ignitionrobotics/ign-physics/tree/main/tpe)
+is placed inside Ignition Physics and hence there is a `lib` folder under `tpe`.
+
+We declare and implement the \ref ignition::physics::FeatureList "FeatureList"
+interfacing with the physics engine API inside `plugin\src` folder
```
chapulina, line 45:
```suggestion
interfacing with the physics engine API inside `plugin/src` folder
```

chapulina (changes requested):
> I added an example CMakeList but don't think it's working correctly yet.

The plugin fails to compile for me with:

```
/home/chapulina/ws_citadel/src/ign-physics/examples/simple_tpe_plugin/plugin.cc:20:10: fatal error: Base.hh: No such file or directory
 #include "Base.hh"
          ^~~~~~~~~
```

Looking through the code, I imagine this is supposed to refer to [tpe/plugin/src/Base.hh](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics3/tpe/plugin/src/Base.hh), right? The reason that the example is not being able to find it, is that the header isn't being installed. I suspect this is on purpose though, because we don't need downstream users to be using `tpe-plugin`, only `tpe-lib`, right? I haven't looked closely enough to have a suggestion here, but I think that the example needs to be reworked to only use things from `tpe-lib`, not `tpe-plugin`.

> I think it'd be preferable to merge the tutorial part of the this PR first, and take the CMake part out for later improvement.

I think that the CMake failure exposed a potential problem in the tutorial. I think it's worth it getting it to work before we publish it, so we're sure we're giving users working instructions.

chapulina (approved):
I pushed some tweaks to:

* Make the example and test compile
* Maintain the current tutorial URLs, so we don't break existing links to the site
* Fix snippets

LGTM!

PR #205, opened 2021-01-22
Bump in edifice: ign-common4

chapulina:
Addresses https://github.com/ignition-tooling/release-tools/issues/361

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,7 +1,7 @@
 libbenchmark-dev
 libeigen3-dev
 libignition-cmake2-dev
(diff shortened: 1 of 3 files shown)


mjcarroll (approved):

scpeters:
the Ubuntu jenkins build is installing both common3 and common4 and failing (I believe since https://github.com/ignitionrobotics/ign-common/pull/170 is not yet merged)

perhaps it's due to [this line](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/dependencies_archive.sh#L596) in release-tools?

Issue #219, opened 2021-03-03
Expose API to check what links moved


chapulina:
A downstream application using Ignition Physics, such as Ignition Gazebo, likely wants to know at every iteration what links have moved after a step.

Currently, Gazebo [goes over all links](https://github.com/ignitionrobotics/ign-gazebo/blob/86e6ad93581362eba4e89a58dc1fef058806a081/src/systems/physics/Physics.cc#L1702) and gets their latest poses with [FrameDataRelativeToWorld](https://github.com/ignitionrobotics/ign-physics/blob/db42320c967607039cd597cc3513ded9db06d993/include/ignition/physics/FrameSemantics.hh#L108-L109), because it doesn't have  a way of knowing which ones changed.

## Desired behavior

We should offer a mechanism by which Gazebo could know what moved and wouldn't need to iterate over all links at every step.

## Implementation suggestion

The way Gazebo classic achieves this with the ODE engine is by registering a [callback that ODE calls for each link that has moved](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/physics/ode/ODELink.cc#L124). As a comparison, Gazebo classic doesn't have an equivalent mechanism for DART, and [iterates over all links instead](https://github.com/osrf/gazebo/blob/65753090cbf0d7b75143ac0e629cf18850dac985/gazebo/physics/dart/DARTPhysics.cc#L459-L473), which makes using DART slower than ODE.

We could take the same approach here. Ignition Physics could offer a Feature that allows registering a callback, and each physics engine may choose to provide that API. I don't know if DART offers such a function or if we'd need to add that functionality upstream. My understanding is that Bullet offers a callback API, but I did a quick search and couldn't find it, maybe @Lobotuerk knows what it is?

## Alternatives considered

Instead of a callback, maybe providing a function that returns only the links that moved could also do the trick, but it may perform worse depending on how that structure is populated and passed around.

adlarkin:
Since implementing callback functionality is more of an involved task and may not work for DART, I have proposed a temporary workaround in the meantime: https://github.com/ignitionrobotics/ign-gazebo/pull/669

adlarkin:
I believe that https://github.com/ignitionrobotics/ign-gazebo/pull/678 resolves this issue for DART and TPE. Can we close this, or does something also need to be done for Bullet?

chapulina:
I'm ok closing this, the main target of the feature was the default engine, which is currently DART.

PR #223, opened 2021-03-10
write modified link data to output in physics step

adlarkin:

# 🎉 New feature

Requires https://github.com/ignitionrobotics/ign-math/pull/196

Possible solution for https://github.com/ignitionrobotics/ign-physics/issues/219

## Summary

As mentioned in #219, downstream libraries that use `ign-physics` (like `ign-gazebo`) don't have a way of knowing which links moved in the most recent physics step. This PR takes advantage of a forward step's `output` by writing changed link poses to the output. The output can then be used by downstream libraries to only operate on modified links (take a look at https://github.com/ignitionrobotics/ign-gazebo/pull/678 for an example).

## Test it

Take a look at the tests added in this PR (files changed tab -> `tpe/plugin/src/SimulationFeatures_TEST.cc` or `dartsim/src/SimulationFeatures_TEST.cc`) to see how the modified pose data can be accessed from `ignition::physics::ForwardStep::Output`.

To get an understanding of how this change can be used/tested in downstream libraries, take a look at https://github.com/ignitionrobotics/ign-gazebo/pull/678.

## Checklist
- [x] Added tests
- [x] Added example and/or tutorial


adlarkin:
Note to self: tests for TPE may need to be updated since we are now writing data to output in each forward step (I will look into this)

dartsim/src/SimulationFeatures.cc
```diff
@@ -15,26 +15,33 @@
...
 void SimulationFeatures::WorldForwardStep(
     const Identity &_worldID,
-    ForwardStep::Output & /*_h*/,
+    ForwardStep::Output & _h,
     ForwardStep::State & /*_x*/,
     const ForwardStep::Input & _u)
 {
   IGN_PROFILE("SimulationFeatures::WorldForwardStep");
   auto *world = this->ReferenceInterface<DartWorld>(_worldID);
+  // TODO(adlarkin) sanity check to make sure world is not nullptr?
```
adlarkin, line 44: If `world` is `nullptr`, should we just `return` from the function early?

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +59,54 @@ void SimulationFeatures::WorldForwardStep(
...
+{
+  // TODO(adlarkin) implement this, if it's needed?
+}
+
+void SimulationFeatures::Write(WorldPoses &_poses) const
+{
+  // cache link poses from the previous iteration
+  // (using a static variable for now instead of a member variable
+  // since this method has to be const)
+  static std::unordered_map<std::size_t, math::Pose3d> prevLinkPoses;
```
adlarkin, line 77: As mentioned in the comment, I couldn't use a member variable since the `Write` override method is required to be `const` (so, since this method is `const`, I can't modify member variables inside of it). Any thoughts on how to get around this other than using a static variable?

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +59,54 @@ void SimulationFeatures::WorldForwardStep(
...
+    const auto info = link.second;
+
+    // make sure the link exists
+    if (info && info->link)
+    {
+      WorldPose wp;
+      wp.pose = ignition::math::eigen3::convert(
+          info->link->getWorldTransform());
+      wp.pose.Pos() = ignition::math::eigen3::convert(
+          info->link->getCOM());
```
adlarkin, line 94: This was inspired by the [DARTDoublePendulum test](https://github.com/ignitionrobotics/ign-physics/blob/c77ac49b01e7a57cd0a6f6ba76c46837c65210ab/test/plugins/DARTDoublePendulum.cc#L307-L309). Is there a reason why we need to set the pose's position to the link's center of mass, instead of just using the result from `getWorldTransform` in the previous line?

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +59,54 @@ void SimulationFeatures::WorldForwardStep(
...
-  // TODO(MXG): Fill in output and state
+  this->WriteRequiredData(_h);
+  this->Write(_h.Get<JointPositions>());
+  // TODO(MXG): Fill in state
+}
+
+void SimulationFeatures::Write(JointPositions &/*_positions*/) const
+{
+  // TODO(adlarkin) implement this, if it's needed?
+}
```
adlarkin, line 70: This was also inspired by the [DARTDoublePendulum test](https://github.com/ignitionrobotics/ign-physics/blob/c77ac49b01e7a57cd0a6f6ba76c46837c65210ab/test/plugins/DARTDoublePendulum.cc#L273). However, I don't think we need to do anything with joints here since we only need to know what links moved (https://github.com/ignitionrobotics/ign-physics/issues/219). If this is true, I'll remove this method.

tpe/plugin/src/SimulationFeatures.cc
```diff
@@ -58,6 +62,51 @@ void SimulationFeatures::WorldForwardStep(
...
   }
   world->Step();
+  this->WriteRequiredData(_h);
+  this->Write(_h.Get<JointPositions>());
+}
+
+void SimulationFeatures::Write(JointPositions &/*_positions*/) const
+{
+  // TODO(adlarkin) implement this? Is this needed?
+}
```
adlarkin, line 72: Similar to my comment about this method in the DART implementation, I don't believe that we need this method since we only need to know which links have moved (https://github.com/ignitionrobotics/ign-physics/issues/219). I can remove this method as well if this is true.

tpe/plugin/src/SimulationFeatures.cc
```diff
@@ -41,6 +43,8 @@ void SimulationFeatures::WorldForwardStep(
       << _worldID.id
       << "] not found."
       << std::endl;
+    // TODO(adlarkin) return here? de-referencing it below won't work
+    // if it == this->worlds.end()
```
adlarkin, line 47: I noticed that there's a check to see if the world wasn't found, but we don't seem to stop early if the world isn't found. Should I add a `return` statement here?

tpe/plugin/src/SimulationFeatures.cc
```diff
@@ -58,6 +62,51 @@ void SimulationFeatures::WorldForwardStep(
...
+{
+  // TODO(adlarkin) implement this? Is this needed?
+}
+
+void SimulationFeatures::Write(WorldPoses &_poses) const
+{
+  // cache link poses from the previous iteration
+  // (using a static variable for now instead of a member variable
+  // since this method has to be const)
+  static std::unordered_map<std::size_t, math::Pose3d> prevLinkPoses;
```
adlarkin, line 79: Same comment from the DART implementation applies here about using a static variable

adlarkin (commented):
@azeey the implementation is working, but I have a few questions. Would you mind taking a look at the code and addressing my comments below?

adlarkin:
This is now ready for review. I have updated this PR to target Edifice (`main`) based on the discussion that was had in https://github.com/ignitionrobotics/ign-gazebo/pull/678.

We'll need to make a minor release for `ign-math6` so that https://github.com/ignitionrobotics/ign-math/pull/196 can be used.

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +59,54 @@ void SimulationFeatures::WorldForwardStep(
...
+      // add the link to the output poses
+      auto iter = prevLinkPoses.find(id);
+      if ((iter == prevLinkPoses.end()) ||
+          !iter->second.Pos().Equal(wp.pose.Pos(), 1e-6) ||
+          !iter->second.Rot().Equal(wp.pose.Rot(), 1e-6))
+      {
+        _poses.entries.push_back(wp);
+      }
+
+      prevLinkPoses[id] = wp.pose;
```
iche033, line 107: do you need to keep this map up-to-date when links are removed?

dartsim/src/SimulationFeatures.cc
iche033: you could mark `prevLinkPoses` as `mutable` if you need to modify it in a const function

tpe/plugin/src/SimulationFeatures.cc
iche033: yeah I think we can return here otherwise it may just crash

dartsim/src/SimulationFeatures.cc
adlarkin: Thanks, I wasn't aware of the `mutable` keyword - I learned something new! This has been added in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/baeb7e6606c21ade93ced6af5d96c6139f41d878

tpe/plugin/src/SimulationFeatures.cc
adlarkin: Resolved in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/baeb7e6606c21ade93ced6af5d96c6139f41d878

dartsim/src/SimulationFeatures.cc
adlarkin: Added this check in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/2df9ea0c290f75ea72a9781ba003ca0b84d15ad1

tpe/plugin/src/SimulationFeatures.cc
adlarkin: Added in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/2df9ea0c290f75ea72a9781ba003ca0b84d15ad1

dartsim/src/SimulationFeatures.cc
adlarkin: Yeah, thanks for catching this. I believe I have handled this in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/84b9f4e6088ea23d383cc7178226fe459d53e7fc. Let me know what you think.

dartsim/src/SimulationFeatures.cc
azeey: Yeah, I would suggest we remove this for now instead of leaving it empty.

azeey (commented):
Looks pretty good! I left some minor comments.

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +64,57 @@ void SimulationFeatures::WorldForwardStep(
...
+      // add the link to the output poses
+      auto iter = this->prevLinkPoses.find(id);
+      if ((iter == this->prevLinkPoses.end()) ||
+          !iter->second.Pos().Equal(wp.pose.Pos(), 1e-6) ||
+          !iter->second.Rot().Equal(wp.pose.Rot(), 1e-6))
+      {
+        _poses.entries.push_back(wp);
+      }
+
+      newPoses[id] = wp.pose;
```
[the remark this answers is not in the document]
adlarkin: The reason why I left this outside of the `if` block is because of floating point precision. My thought was that I should save the most recent pose in `newPoses` no matter what each time, since there's a small tolerance (`1e-6`) in the pose equality check. Since `newPoses` is used as `prevLinkPoses` the next time around, this ensures that we are comparing the newest link pose to the actual previous pose, instead of to a pose that is _very close_ to the previous pose. Does that make sense? If you think that leaving this outside of the `if` block is okay, I can add a comment there that mentions what I just described to you. Otherwise, I can move this to be inside of the `if` block.

dartsim/src/SimulationFeatures.cc
```diff
@@ -15,26 +15,38 @@
...
 void SimulationFeatures::WorldForwardStep(
     const Identity &_worldID,
-    ForwardStep::Output & /*_h*/,
+    ForwardStep::Output & _h,
     ForwardStep::State & /*_x*/,
     const ForwardStep::Input & _u)
 {
   IGN_PROFILE("SimulationFeatures::WorldForwardStep");
   auto *world = this->ReferenceInterface<DartWorld>(_worldID);
+  if (!world)
```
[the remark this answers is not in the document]
adlarkin: I removed this check in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

dartsim/src/SimulationFeatures.cc
adlarkin: removed in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

tpe/plugin/src/SimulationFeatures.cc
adlarkin: removed in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +64,57 @@ void SimulationFeatures::WorldForwardStep(
...
+      }
+
+      newPoses[id] = wp.pose;
+    }
+  }
+
+  // Save the new poses so that they can be used to check for updates in the
+  // next iteration. Re-setting this->prevLinkPoses with the contents of
+  // newPoses ensures that we aren't caching data for links that were removed
+  this->prevLinkPoses.clear();
```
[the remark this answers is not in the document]
adlarkin: Nope, no need for this. Removed in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +64,57 @@ void SimulationFeatures::WorldForwardStep(
...
+
+void SimulationFeatures::Write(JointPositions &/*_positions*/) const
+{
+  // TODO(adlarkin) implement this, if it's needed?
+}
+
+void SimulationFeatures::Write(WorldPoses &_poses) const
+{
+  // remove link poses from the previous iteration
+  _poses.entries.clear();
```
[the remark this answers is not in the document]
adlarkin: reserve call added in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

dartsim/src/SimulationFeatures.cc
```diff
@@ -52,7 +64,57 @@ void SimulationFeatures::WorldForwardStep(
...
+
+  std::unordered_map<std::size_t, math::Pose3d> newPoses;
+
+  for (const auto &link : this->links.idToObject)
```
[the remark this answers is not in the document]
adlarkin: updated in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/ad12778a0af2f312222f108db4bd90ebf22502cb

dartsim/src/SimulationFeatures.cc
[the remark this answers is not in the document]
azeey: I would think you'd want to compare it to the last pose sent to the user though because you want to err on the side of sending too much to the user (eg. ign-gazebo). Consider, for example, if each successive time the pose changes by +1e-7, since you always compare against the previous actual pose, it will always look like there no new pose to send to the user. Over time, this could cause a significant amount of drift.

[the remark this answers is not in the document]
adlarkin: You're definitely correct. I believe that this should now be fixed in https://github.com/ignitionrobotics/ign-physics/pull/223/commits/494f6c567926e42805a94d29adb65e66cc0c8186

adlarkin:
All review feedback has been addressed, so I am opening this for another round of review.

azeey:

azeey (approved):

PR #220, opened 2021-03-04
[dartsim] Fix joint construction errors due to link name duplication or BodyNodes moving to other skeletons

azeey:
# 🦟 Bug fix

## Summary
This started out as a fix for link name collision error when two links in different models have the same name. https://github.com/ignitionrobotics/ign-physics/commit/b54cd35001de3d0f09c75db004ce63ec50aef849. However, I also found that when creating joints between links in different models, DART actually moves the BodyNode associated with the child link of the joint from its skeleton to the skeleton of the parent BodyNode. After the BodyNode has been moved, `SDFFeatures::FindBodyNode` was unable to find the child link because it looked for it in the original skeleton.

For example,
```xml
<?xml version="1.0" ?>
<sdf version="1.8">
  <world name="nested_model_world">
    <model name="parent_model">

      <model name="nested_model1">
        <link name="nested_link1"/>
      </model>
      <model name="nested_model2">
        <link name="nested_link2"/>
      </model>
      <model name="nested_model3">
        <link name="nested_link3"/>
      </model>

      <joint name="joint1" type="revolute">
        <parent>nested_model1::nested_link1</parent>
        <child>nested_model2::nested_link2</child>
      </joint>
      <joint name="joint2" type="revolute">
        <parent>nested_model2::nested_link2</parent>
        <child>nested_model3::nested_link3</child>
      </joint>

    </model>
  </world>
</sdf>
```
When `joint1` is created, the link `parent_model::nested_model2::nested_link2` is moved from it's original skeleton, `parent_model::nested_model2` to `parent_model::nested_model1`. When it's time to create `joint2`, it will look for it's parent link `nested_link2` in the skeleton `parent_model::nested_model2`, but it will not be found.

This PR adds a new map that tracks BodyNodes by their fully qualified name. I encountered the errors when testing with ign-gazebo, so I have added an alternative `LoadWorld` function in `SDFFeatures_TEST.cc` that loads an SDFormat world in a piecemeal manner like ign-gazebo.

PS: A simpler test would be to load the example world above and check if all the joints have been created. However, since the joints themselves move to a new skeleton, we need more changes in the code for `parentModel->GetJointCount()` to return the correct number. I'll create a separate issue for that.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
[...]

PR #227, opened 2021-03-17
[dartsim] Ensure Link and Model APIs continue to work after joint creation in DART

azeey:
# 🎉 New feature

## Summary
When a DART joint is created between two BodyNodes in DART, the child body node is moved from its original skeleton to the parent skeleton. This breaks APIs such as `Model::GetLink(std::string)`, `Model::GetLink(std::size_t)` and `Link::GetIndex()`. This PR fixes this by keeping track of links separately from DART. 

**Requires**
* #220 

## Test it
Run tests

## Checklist
- [x] Added tests


dartsim/src/EntityManagementFeatures.cc
```diff
@@ -272,22 +272,25 @@ std::size_t EntityManagementFeatures::GetLinkCount(
...
 {
-  DartBodyNode *const bn =
-      this->ReferenceInterface<ModelInfo>(_modelID)->model->getBodyNode(
-          _linkIndex);
+  auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
+
+  if (_linkIndex >= modelInfo->links.size())
+    return this->GenerateInvalidId();
+
+  auto linkInfo = modelInfo->links[_linkIndex];
```
adlarkin, line 287:
```suggestion
  const auto linkInfo = modelInfo->links[_linkIndex];
```
Based on the usage of `linkInfo` below, can this variable be marked `const`?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -304,25 +307,29 @@ Identity EntityManagementFeatures::GetLink(
...
-
-  // If the link doesn't exist in "links", it means the containing entity has
-  // been removed.
-  if (this->links.HasEntity(bn))
-  {
-    const std::size_t linkID = this->links.IdentityOf(bn);
-    return this->GenerateIdentity(linkID, this->links.at(linkID));
-  }
-  else
+  auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
```
adlarkin, line 310:
```suggestion
  const auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
```
Based on how I'm seeing this variable being used in the rest of this method, I believe this variable can be `const`.

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -304,25 +307,29 @@ Identity EntityManagementFeatures::GetLink(
...
+        const std::size_t linkID = this->links.IdentityOf(linkInfo->link);
+        return this->GenerateIdentity(linkID, this->links.at(linkID));
+      }
+      else
+      {
+        // TODO(addisu) It's not clear what to do when `GetLink` is called on a
+        // model that has been removed. Right now we are returning an invalid
+        // identity, but that could cause a segfault if the use doesn't check if
+        // returned value before using it.
+        return this->GenerateInvalidId();
```
adlarkin, line 328: Would it at least be worth removing the link from `modelInfo->links` if we find the link has been removed here?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -304,25 +307,29 @@ Identity EntityManagementFeatures::GetLink(
...
+      if (this->links.HasEntity(linkInfo->link))
+      {
+        const std::size_t linkID = this->links.IdentityOf(linkInfo->link);
+        return this->GenerateIdentity(linkID, this->links.at(linkID));
+      }
+      else
+      {
+        // TODO(addisu) It's not clear what to do when `GetLink` is called on a
+        // model that has been removed. Right now we are returning an invalid
+        // identity, but that could cause a segfault if the use doesn't check if
```
adlarkin, line 326:
```suggestion
        // identity, but that could cause a segfault if the user doesn't check the
```

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -393,22 +400,19 @@ const std::string &EntityManagementFeatures::GetLinkName(
...
 
   // If the model containing the link doesn't exist in "models", it means this
   // link belongs to a removed model.
-  if (this->models.HasEntity(model))
+  if (this->models.HasEntity(modelID))
   {
-    const std::size_t modelID = this->models.IdentityOf(model);
     return this->GenerateIdentity(modelID, this->models.at(modelID));
   }
   else
```
adlarkin, line 418: If the model has been removed, should we remove the model from `this->models` before returning `this->GenerateInvalidId()`?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -272,22 +272,25 @@ std::size_t EntityManagementFeatures::GetLinkCount(
...
   // If the link doesn't exist in "links", it means the containing entity has
   // been removed.
-  if (this->links.HasEntity(bn))
+  if (this->links.HasEntity(linkInfo->link))
   {
-    const std::size_t linkID = this->links.IdentityOf(bn);
+    const std::size_t linkID = this->links.IdentityOf(linkInfo->link);
     return this->GenerateIdentity(linkID, this->links.at(linkID));
   }
   else
```
adlarkin, line 296: If the link has been removed, should we remove it from `this->links` before returning `this->GenerateInvalidId()`?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -393,22 +400,19 @@ const std::string &EntityManagementFeatures::GetLinkName(
 std::size_t EntityManagementFeatures::GetLinkIndex(
```
adlarkin, line 400: I didn't see any tests written that use `EntityManagementFeatures::GetLinkIndex`, would it be worth adding some in?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -393,22 +400,19 @@ const std::string &EntityManagementFeatures::GetLinkName(
...
     const Identity &_linkID) const
 {
-  return this->ReferenceInterface<LinkInfo>(_linkID)
-      ->link->getIndexInSkeleton();
+  return this->links.idToIndexInContainer.at(_linkID);
 }
 
 /////////////////////////////////////////////////
 Identity EntityManagementFeatures::GetModelOfLink(
```
adlarkin, line 407: I didn't see any tests written that use `EntityManagementFeatures::GetModelOfLink`, would it be worth adding some in?

dartsim/src/SDFFeatures.cc
```diff
@@ -534,7 +534,7 @@ Identity SDFFeatures::ConstructSdfLink(
     const Identity &_modelID,
     const ::sdf::Link &_sdfLink)
 {
-  const auto &modelInfo = *this->ReferenceInterface<ModelInfo>(_modelID);
+  auto &modelInfo = *this->ReferenceInterface<ModelInfo>(_modelID);
```
adlarkin, line 537: Is there a reason why `modelInfo` is no longer `const`?

dartsim/src/SDFFeatures_TEST.cc
```diff
@@ -558,16 +558,37 @@ TEST_P(SDFFeatures_TEST, WorldWithNestedModel)
...
+
+  auto nestedModel2 = world->GetModel("parent_model::nested_model2");
+  ASSERT_NE(nullptr, nestedModel2);
+  EXPECT_EQ(1u, nestedModel2->GetLinkCount());
+  EXPECT_NE(nullptr, nestedModel2->GetLink("nested_link1"));
+
+  auto nestedModel3 = world->GetModel("parent_model::nested_model2");
+  ASSERT_NE(nullptr, nestedModel3);
+  EXPECT_EQ(1u, nestedModel3->GetLinkCount());
+  EXPECT_NE(nullptr, nestedModel3->GetLink("nested_link1"));
```
adlarkin, line 591: It looks like these tests are the same, can we remove one of them?

dartsim/src/SDFFeatures.cc
```diff
@@ -534,7 +534,7 @@ Identity SDFFeatures::ConstructSdfLink(
     const Identity &_modelID,
     const ::sdf::Link &_sdfLink)
 {
```
adlarkin, line 536: I don't see any tests written for `SDFFeatures::ConstructSdfLink`. Should we add one?

adlarkin (commented):
Looks pretty good! I left a few minor comments/questions.

dartsim/src/Base.hh
```diff
@@ -289,18 +292,30 @@ class Base : public Implements3d<FeatureList<Feature>>
...
 
     this->linksByName[_fullName] = _bn;
+    this->models.at(_modelID)->links.push_back(linkInfo);
+
+    // Even though DART keeps track of the index of this BodyNode in the
+    // skeleton, the BodyNode may be moved to another skeleton when a joint is
+    // constructed. Thus, we store the original index here.
+    this->links.idToIndexInContainer[id] = _bn->getIndexInSkeleton();
+    std::vector<std::size_t> &indexInContainerToID =
+        this->links.indexInContainerToID[_modelID];
```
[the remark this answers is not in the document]
azeey: `indexInContainerToID` is a map from the parent ID to a vector that contains child IDs. The index information is implicit in the location (index) of the child ID in the vector. So here, we first obtain a reference to the vector for the parent model using `_modelID` and then `push_back` the link's ID.

https://github.com/ignitionrobotics/ign-physics/blob/85156645f13c6af9b3214478ba035b028419005b/dartsim/src/Base.hh#L102-L114

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -272,22 +272,25 @@ std::size_t EntityManagementFeatures::GetLinkCount(
...
 /////////////////////////////////////////////////
 Identity EntityManagementFeatures::GetLink(
     const Identity &_modelID, const std::size_t _linkIndex) const
 {
-  DartBodyNode *const bn =
-      this->ReferenceInterface<ModelInfo>(_modelID)->model->getBodyNode(
-          _linkIndex);
+  auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
+
+  if (_linkIndex >= modelInfo->links.size())
```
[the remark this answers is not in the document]
azeey: Since `_linkIndex` is unsigned, it can't be less than 0.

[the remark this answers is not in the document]
adlarkin: I didn't notice that :man_facepalming: looks good to me then!

dartsim/src/Base_TEST.cc
```diff
@@ -68,7 +68,7 @@ TEST(BaseClass, RemoveModel)
     const std::string fullName = ::sdf::JoinName(
         world->getName(),
         ::sdf::JoinName(skel->getName(), pair.second->getName()));
-    base.AddLink(pair.second, fullName);
+    base.AddLink(pair.second, fullName, std::get<0>(res));
```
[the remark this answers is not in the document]
azeey: Added more checks in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2

dartsim/src/JointFeatures_TEST.cc
```diff
@@ -433,10 +433,8 @@ TEST_F(JointFeaturesFixture, LinkCountsInJointAttachDetach)
 
   // After attaching we expect each model to have 1 link, but the current
   // behavior is that there are 2 links in model1 and 0 in model2
```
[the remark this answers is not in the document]
azeey: Removed the second half of the comment in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2

dartsim/src/EntityManagementFeatures.cc
azeey: Yup, https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2

dartsim/src/EntityManagementFeatures.cc
azeey: While I agree that `modelInfo->links` and `Base::links` being out of sync causes an invalid state, I'd rather not modify `modelInfo->links` from a const function that's simply supposed to return a link identity. If this case actually occurs, I think we should fix the root cause, which is that the `RemoveModel` function is not working properly.

dartsim/src/EntityManagementFeatures.cc
azeey: Similar to my previous [comment](https://github.com/ignitionrobotics/ign-physics/pull/227#discussion_r598841185), I don't think we should have side effects in this function.

dartsim/src/EntityManagementFeatures.cc
azeey: Similar to my previous [comment](https://github.com/ignitionrobotics/ign-physics/pull/227#discussion_r598841185), I don't think we should have side effects in this function.

dartsim/src/EntityManagementFeatures.cc
azeey: Added some expectations in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2.

dartsim/src/SDFFeatures.cc
azeey: hmm, I think I was passing `modelInfo` to `AddLink` at one point. Reverted back to `const` in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2.

dartsim/src/SDFFeatures_TEST.cc
azeey: Good catch! It was supposed to be testing `nested_model3`. Fixed inhttps://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2

dartsim/src/SDFFeatures.cc
azeey: Unless there's a specific reason to do so, IMO, this is covered by the `ConstructWorld` tests in `SDFFeatures_TEST.cc`.

dartsim/src/EntityManagementFeatures.cc
azeey: You're right, I changed it to a const ref in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/16da5e14a35a9e3e437287a576e11f014425fe91

dartsim/src/EntityManagementFeatures.cc
azeey: https://github.com/ignitionrobotics/ign-physics/pull/227/commits/16da5e14a35a9e3e437287a576e11f014425fe91

dartsim/src/EntityManagementFeatures.cc
azeey: Added tests in https://github.com/ignitionrobotics/ign-physics/pull/227/commits/0bb7d9e9d787d815f7fd3db15aa2da9a815c40b2

adlarkin (approved):
Looks good to me! Just waiting on CI to come back as green, hopefully :crossed_fingers:

PR #228, opened 2021-03-17
[dartsim] Add empty nested model construction and nested model entity management

azeey:
# 🎉 New feature

## Summary
This adds two new features:
1.  `ConstructEmptyNestedModelFeature`, which allows users to create a nested model without having to use `ConstructSdfNestedModel`
2. `GetNestedModelFromModel`, which allow retrieving a nested model from it's parent `Model` entity.

**Requires:**
* ~#220~ 
* ~#227~

## Test it
Run tests

## Checklist
- [x] Added tests


include/ignition/physics/detail/GetEntities.hh
```diff
@@ -214,6 +214,58 @@ namespace ignition
...
+    GetNestedModelFromModel::Model<PolicyT, FeaturesT>::GetModelCount() const
+    {
+      return this->template Interface<GetNestedModelFromModel>()
+          ->GetNestedModelCount(this->identity);
+    }
+
+    /////////////////////////////////////////////////
+    template <typename PolicyT, typename FeaturesT>
+    auto GetNestedModelFromModel::Model<PolicyT, FeaturesT>::GetModel(
+        const std::size_t _index) -> ModelPtrType
```
adlarkin, line 229: I noticed that the declaration of this method doesn't have the `_index` parameter marked as `const` (I noticed the same trend when looking at other features):

https://github.com/ignitionrobotics/ign-physics/blob/ab959226520f908b4996aa52f749ba80d15099b8/include/ignition/physics/GetEntities.hh#L232-L237

Can you help me understand why these don't match, and how this still works/compiles?

dartsim/src/EntityManagementFeatures.hh
```diff
@@ -37,6 +37,7 @@ struct EntityManagementFeatureList : FeatureList<
   RemoveEntities,
   ConstructEmptyWorldFeature,
   ConstructEmptyModelFeature,
+  ConstructEmptyNestedModelFeature,
```
adlarkin, line 40: Do we also need to add the `GetNestedModelFromModel` feature? It looks like methods from both the `ConstructEmptyNestedModelFeature` and `GetNestedModelFromModel` features are being implemented below.

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -267,6 +267,64 @@ Identity EntityManagementFeatures::GetWorldOfModel(
...
+    const std::size_t nestedModelID = this->models.IdentityOf(nestedSkel);
+    return this->GenerateIdentity(nestedModelID,
+                                  this->models.at(nestedModelID));
+  }
+  else
+  {
+    return this->GenerateInvalidId();
+  }
+}
 /////////////////////////////////////////////////
```
adlarkin, line 328:
```suggestion
}

/////////////////////////////////////////////////
```

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -267,6 +267,64 @@ Identity EntityManagementFeatures::GetWorldOfModel(
...
+    const Identity &_modelID) const
+{
+  return this->ReferenceInterface<ModelInfo>(_modelID)->nestedModels.size();
+}
+
+/////////////////////////////////////////////////
+Identity EntityManagementFeatures::GetNestedModel(
+    const Identity &_modelID, const std::size_t _modelIndex) const
+{
+  auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
```
adlarkin, line 281: I believe that `modelInfo` can be marked `const` here.

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -267,6 +267,64 @@ Identity EntityManagementFeatures::GetWorldOfModel(
...
+  {
+    return this->GenerateInvalidId();
+  }
+}
+
+/////////////////////////////////////////////////
+Identity EntityManagementFeatures::GetNestedModel(
+    const Identity &_modelID, const std::string &_modelName) const
+{
+  auto modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
```
adlarkin, line 306: I believe that `modelInfo` can be marked `const` here.

dartsim/src/Base.hh
```diff
@@ -291,6 +293,15 @@ class Base : public Implements3d<FeatureList<Feature>>
     return std::forward_as_tuple(id, entry);
   }
 
+  public: inline std::tuple<std::size_t, ModelInfo &> AddNestedModel(
+              const ModelInfo &_info, const std::size_t _parentID,
+              const std::size_t _worldID)
```
adlarkin, line 298: If I recall correctly, according to the Ignition style guide, we don't mark method parameters with primitive data types as `const`. I'm not sure if `std::size_t` counts towards this criteria or not.

adlarkin, line 298: Should we add a test for the `AddNestedModel` method in `Base_TEST.cc`?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -636,7 +694,31 @@ Identity EntityManagementFeatures::ConstructEmptyModel(
...
+  dart::dynamics::SkeletonPtr model =
+      dart::dynamics::Skeleton::create(modelFullName);
+
+  dart::dynamics::SimpleFramePtr modelFrame =
+      dart::dynamics::SimpleFrame::createShared(
+        dart::dynamics::Frame::World(),
+        modelFullName + "_frame");
+
+  auto [modelID, modelInfo] =
+      this->AddModel({model, _name, modelFrame, ""}, worldID);  // NOLINT
```
adlarkin, line 721:
```suggestion
  auto [modelID, modelInfo] =
      this->AddNestedModel({model, _name, modelFrame, ""}, worldID);  // NOLINT
```
Did you mean to call `AddNestedModel` here, since this is a part of the `ConstructEmptyNestedModel` method?

dartsim/src/SDFFeatures.cc
```diff
@@ -420,9 +421,20 @@ Identity SDFFeatures::ConstructSdfModelImpl(
...
+    {
+      return this->AddNestedModel(  // NOLINT
+          {model, _sdfModel.Name(), modelFrame, _sdfModel.CanonicalLinkName()},
+          _parentID, worldID);
+    }
+    else
+    {
+      return this->AddModel(  // NOLINT
+          {model, _sdfModel.Name(), modelFrame, _sdfModel.CanonicalLinkName()},
+          _parentID);
```
adlarkin, line 435:
```suggestion
      return this->AddModel(  // NOLINT
          {model, _sdfModel.Name(), modelFrame, _sdfModel.CanonicalLinkName()},
          _parentID, worldID);
```
I believe the world ID parameter is missing here.

dartsim/src/EntityManagement_TEST.cc
```diff
@@ -62,6 +62,12 @@ TEST(EntityManagement_TEST, ConstructEmptyWorld)
   EXPECT_EQ(world, model->GetWorld());
   EXPECT_NE(model, world->ConstructEmptyModel("dummy"));
 
+  auto nestedModel = model->ConstructEmptyModel("empty nested model");
+  ASSERT_NE(nullptr, nestedModel);
+  EXPECT_EQ("empty nested model", nestedModel->GetName());
+  EXPECT_EQ(world, nestedModel->GetWorld());
+  EXPECT_NE(nestedModel, nestedModel->ConstructEmptyModel("dummy"));
```
adlarkin, line 69: It looks like we are testing `ConstructEmptyModel` here, but that was already done above. Did you mean to test `ConstructEmptyNestedModel` instead?

dartsim/src/EntityManagement_TEST.cc
```diff
@@ -62,6 +62,12 @@ TEST(EntityManagement_TEST, ConstructEmptyWorld)
...
+
```
adlarkin, line 70: Can we add tests for the other methods that were added to `EntityManagementFeatures` (`GetNestedModelCount` and `GetNestedModel` - both the index and name parameter versions)?

adlarkin (commented):
Looks pretty good! I have a few comments, and also asked a few questions to help me get a better grasp of `ign-physics` (I'm still trying to familiarize myself with the repository).

dartsim/src/Base.hh
mxgrey: When `const` is used for the parameter type of a function *definition* then it serves an important role in const-correctness, ensuring that the parameter's value is never modified anywhere inside the function definition.

The style guide probably says that the function *declaration* should not use `const` on primitive-type parameters, which is fine because whether or not you put `const` on pass-by-value parameter does not affect the function signature.

You can see more about this topic in [this discussion](https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters).

include/ignition/physics/GetEntities.hh
```diff
@@ -213,6 +213,56 @@ namespace ignition
...
+
+        /// \brief Get the number of Models inside this Model.
+        public: std::size_t GetModelCount() const;
+
+        /// \brief Get a Model that exists within this Model.
+        /// \param[in] _index
+        ///   Index of the model within this model.
+        /// \return A model reference. If _index is GetModelCount() or higher,
+        /// this will be a nullptr.
+        public: ModelPtrType GetModel(std::size_t _index);
```
mxgrey, line 237: I would suggest making this function more explicit by calling it `GetNestedModel`. Same for the overloads of `GetModel` below.

include/ignition/physics/GetEntities.hh
```diff
@@ -213,6 +213,56 @@ namespace ignition
       };
     };
 
+    /////////////////////////////////////////////////
+    /// \brief This feature retrieves the nested model pointer from the parent
+    /// model by specifying the model index and model index/name.
```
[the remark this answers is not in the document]
mxgrey: I think this is probably supposed to say

```
/// \brief This feature retrieves the nested model pointer from the parent
/// model by specifying the index or name of the nested model.
```

From the user's point of view, the parent model does not need to be specified, so I would recommend not mentioning that in the public API documentation, as it may be misleading.

dartsim/src/SDFFeatures.cc
mxgrey: It's not that `worldID` is missing, but rather `_parentID` should be replaced with `worldID`.

mxgrey: After looking at the code further, it does appear to be correct, although I would urge some caution because the type of entity that `_parentID` refers to is conditional. It may refer to a world or a model depending on whether the model that's being constructed is nested. This may be a strong recipe for some maintenance confusion and a future bug creeping in.

dartsim/src/Base.hh
adlarkin: Ah, yes, that's correct, I forgot about the nuances between definitions and declarations. This looks good to me then, so I'll mark this as resolved.

include/ignition/physics/ConstructEmpty.hh
```diff
@@ -73,6 +73,30 @@ class ConstructEmptyModelFeature : public virtual Feature
...
+class ConstructEmptyNestedModelFeature : public virtual Feature
+{
+  public: template <typename PolicyT, typename FeaturesT>
+  class Model : public virtual Feature::Model<PolicyT, FeaturesT>
+  {
+    public: using ModelPtrType = ModelPtr<PolicyT, FeaturesT>;
+
+    /// \brief Construct an empty model and attach a given name to it.
+    /// \return
+    ///   The ModelPtrType of the constructed model.
```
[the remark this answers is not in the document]
azeey: > It looks like all the other method documentation in this file has a new line between \return and the corresponding information, is there a reason for this formatting?

I don't know what the reason is, but I was being consistent with the style in the file. Doxygen works either way. I think it might be a preference thing.

> Also, I noticed documentation for other methods in this file omits param[in], but I'm not sure why. Is there any particular reason for leaving out method parameter documentation?

No, I think that's just an oversight. I'll add those in.

dartsim/src/EntityManagementFeatures.hh
azeey: I believe that's already included in the `GetEntities` `FeatureList`. https://github.com/azeey/ign-physics/blob/dartsim_nested_entities/include/ignition/physics/GetEntities.hh#L510

include/ignition/physics/detail/GetEntities.hh
azeey: This has been addressed by Grey in https://github.com/ignitionrobotics/ign-physics/pull/228#discussion_r599292319

dartsim/src/Base.hh
azeey: Thanks for the explanation @mxgrey!

dartsim/src/EntityManagementFeatures.hh
adlarkin: Thanks for pointing that out. Sorry for not seeing that the first time around!

include/ignition/physics/ConstructEmpty.hh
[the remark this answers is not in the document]
azeey: Added in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

include/ignition/physics/ConstructEmpty.hh
```diff
@@ -73,6 +73,30 @@ class ConstructEmptyModelFeature : public virtual Feature
   };
 };
 
+/////////////////////////////////////////////////
+/// \brief This feature constructs an empty nested model and return its pointer
```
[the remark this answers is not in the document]
azeey: https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

include/ignition/physics/GetEntities.hh
```diff
@@ -213,6 +213,56 @@ namespace ignition
...
+            const Identity &_modelID) const = 0;
+
+        public: virtual Identity GetNestedModel(
+            const Identity &_modelID, std::size_t _modelIndex) const = 0;
+
+        public: virtual Identity GetNestedModel(
+            const Identity &_modelID, const std::string &_modelName) const = 0;
+      };
+    };
     /////////////////////////////////////////////////
```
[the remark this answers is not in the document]
azeey: https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

include/ignition/physics/GetEntities.hh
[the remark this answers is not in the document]
azeey: Changed to say "...name or index of the nested model". https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

dartsim/src/EntityManagementFeatures.cc
azeey: https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

dartsim/src/EntityManagementFeatures.cc
azeey: Made it `const` in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82.

dartsim/src/EntityManagementFeatures.cc
azeey: Made it `const` in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/b836a0aa0fc7e9407e9f9d569491d3b9d32d9f82

dartsim/src/Base.hh
azeey: Added a test in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/a652ef2b94b139322aa4db718d4e6606099634e1. Let me know if that's sufficient.

dartsim/src/EntityManagementFeatures.cc
azeey: Good catch. I changed it to `AddNestedModel` in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/212c941319d4a7227ef38ae6bdbf196f2cec576b. Also in that commit, I added a test expectation that fails without the fix.

dartsim/src/EntityManagement_TEST.cc
azeey: Since `ConstructEmptyModel` was being called on a `Model` entity, it was actually calling `ConstructEmptyNestedModel` on the physics-plugin side. But to avoid confusion and per @mxgrey's [suggestion](https://github.com/ignitionrobotics/ign-physics/pull/228#discussion_r599303356), I have changed the user facing function to  `ConstructEmptyNestedModel` in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/d137bba029a5cea24d22ee5458ef32b740e32fcf

include/ignition/physics/GetEntities.hh
azeey: Done in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/d137bba029a5cea24d22ee5458ef32b740e32fcf. Also made the same change to `ConstructEmptyNestedModel`.

dartsim/src/EntityManagement_TEST.cc
azeey: Added more test expectations in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/212c941319d4a7227ef38ae6bdbf196f2cec576b.

dartsim/src/SDFFeatures.cc
azeey: The logic is `_parentID` and `worldID` are the same if the model is **not** nested. But I agree it is confusing. I changed it to `worldID` in https://github.com/ignitionrobotics/ign-physics/pull/228/commits/212c941319d4a7227ef38ae6bdbf196f2cec576b.

> This may be a strong recipe for some maintenance confusion and a future bug creeping in.

I agree. FWIW, this was done to avoid code duplication between `ConstructSdfModel` and `ConstructSdfNestedModel` where unlike `ConstructEmptyNestedModel`, `ConstructSdfNestedModel` is a feature that can construct a model and it's nested models from `sdf::Model`. i.e, `ConstructSdfNestedModel` is the same as `ConstructSdfModel` but can also handle nested models if they exist in the input `sdf::Model`.

azeey:
I believe I have addressed all the feedback. Please take a look.

dartsim/src/Base.hh
adlarkin: Test look good, thanks!

adlarkin (approved):

PR #229, opened 2021-03-17
[tpe] Add empty nested model construction and nested model entity management

azeey:
# 🎉 New feature

This is the same as #228, but for TPE.

## Summary
This adds two new features:
1.  `ConstructEmptyNestedModelFeature`, which allows users to create a nested model without having to use `ConstructSdfNestedModel`
2. `GetNestedModelFromModel`, which allow retrieving a nested model from it's parent `Model` entity.

**Requires:**
* ~#220~ 
* ~#227~ 
* ~#228~

## Test it
Run tests

## Checklist
- [x] Added tests


include/ignition/physics/ConstructEmpty.hh
```diff
@@ -73,6 +73,30 @@ class ConstructEmptyModelFeature : public virtual Feature
...
+{
+  public: template <typename PolicyT, typename FeaturesT>
+  class Model : public virtual Feature::Model<PolicyT, FeaturesT>
+  {
+    public: using ModelPtrType = ModelPtr<PolicyT, FeaturesT>;
+
+    /// \brief Construct an empty model and attach a given name to it.
+    /// \return
+    ///   The ModelPtrType of the constructed model.
+    public: ModelPtrType ConstructEmptyModel(const std::string &_name);
```
mxgrey, line 89: Similar to my comment [here](https://github.com/ignitionrobotics/ign-physics/pull/228/files#r599303356) I would recommend naming this function `ConstructEmptyNestedModel` to make the nested model's relationship to the parent model more explicit.

tpe/lib/src/Model.cc
```diff
@@ -75,6 +76,16 @@ Entity &Model::AddLink()
   return *it->second.get();
 }
 
+//////////////////////////////////////////////////
+std::size_t Model::GetLinkCount() const
+{
+  const auto &children = this->GetChildren();
+  return std::count_if(children.begin(), children.end(), [](auto _child)
+  {
+    return std::dynamic_pointer_cast<Link>(_child.second);
```
mxgrey, line 85: I'm somewhat alarmed that `GetLinkCount` would involve an O(N) iteration through an array and attempting a dynamic cast on each entry. I would strongly urge us to separate the `children` container into separate containers for links and nested models.

The way it is, I think users may get irritated that their very simple for-loop actually has O(N^2) complexity because checking the terminal condition has O(N) complexity on each iteration.

tpe/plugin/src/EntityManagementFeatures.cc
```diff
@@ -172,11 +172,54 @@ Identity EntityManagementFeatures::GetWorldOfModel(
...
+{
+  auto modelInfo = this->ReferenceInterface<WorldInfo>(_modelID);
+  if (modelInfo != nullptr)
+  {
+    tpelib::Entity &modelEnt = modelInfo->world->GetChildByName(_modelName);
+    for (const auto &[parentModelId, nestedModelInfo]: this->models)
+    {
+      if (nestedModelInfo != nullptr)
+      {
+        std::string name = nestedModelInfo->model->GetName();
```
mxgrey, line 207: Nitpick

```suggestion
        const std::string &name = nestedModelInfo->model->GetName();
```

mxgrey (changes requested):
The implementation of the feature for `dartsim` looks fine, but I would urge us to change the implementation for TPE. If we split the `children` container into separate `links` and `nestedModels` containers, then we can replace some `dynamic_cast`ing with `static_cast` and eliminate O(N) complexity counts for finding the element counts.

include/ignition/physics/ConstructEmpty.hh
azeey: Changed in https://github.com/ignitionrobotics/ign-physics/pull/229/commits/f0a849285efb0a501e89030cd318eb2ce7b7b175

tpe/lib/src/Model.cc
azeey: Since the `children` container is in the base class `tpelib::Entity`, I opted for adding extra bookkeeping in `tpelib::Model`. https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e17aed3bb5b6ae4e68658cb7ec39dbe4cdd72a16.

tpe/plugin/src/EntityManagementFeatures.cc
azeey: https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e17aed3bb5b6ae4e68658cb7ec39dbe4cdd72a16

tpe/lib/src/Model.cc
azeey: This adds a little more complexity when it comes to removing models. My approach in #230 is to override the functions `tpelib::Entity::RemoveChildById` and `tpelib::Entity::RemoveChildByName` in `tpelib::Model`. There is some extra work now when removing models, but it's a trad-off I'm willing to accept to keep `GetLinkCount` O(1).

tpe/plugin/src/EntityManagement_TEST.cc
```diff
@@ -79,6 +79,17 @@ TEST(EntityManagement_TEST, ConstructEmptyWorld)
   EXPECT_EQ(model, link->GetModel());
   EXPECT_EQ(link, model->GetLink(0));
   EXPECT_EQ(link, model->GetLink("empty link"));
+
+  EXPECT_EQ(0u, model->GetNestedModelCount());
+  auto nestedModel = model->ConstructEmptyNestedModel("empty nested model");
+  EXPECT_EQ(1u, model->GetNestedModelCount());
+  ASSERT_NE(nullptr, nestedModel);
+  EXPECT_EQ("empty nested model", nestedModel->GetName());
```
adlarkin, line 87:
```suggestion
  EXPECT_EQ("empty nested model", nestedModel->GetName());
  EXPECT_EQ(nestedModel, model->GetNestedModel("empty nested model"));
  EXPECT_EQ(nestedModel, model->GetNestedModel(0));
```
Can we add in some tests for the two new `GetNestedModel(` methods? It looks like you've already added tests for `GetNestedModelCount` and `ConstructEmptyNestedModel`, so I believe the two `GetNestedModel` methods are the only ones missing.

adlarkin (approved):
Looks great after addressing the comments from @mxgrey. I just have a few things to add before merging.

tpe/lib/src/Model.cc
```diff
@@ -16,6 +16,7 @@
 */
 
 #include <set>
+#include <algorithm>
```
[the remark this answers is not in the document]
azeey: We do not https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e5679df5d6698b29f31fc34d37428cbceb8cfc52

tpe/lib/src/Model.cc
```diff
@@ -34,6 +35,12 @@ class ignition::physics::tpelib::ModelPrivate
 
   /// \brief First inserted link id;
   public: std::size_t firstLinkId = kNullEntityId;
+
+  /// \brief Links in the model
+  public: std::vector<std::size_t> linkIds;
+
+  /// \brief Nested models links
```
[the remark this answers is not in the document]
azeey: https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e5679df5d6698b29f31fc34d37428cbceb8cfc52

tpe/plugin/src/EntityManagementFeatures.cc
```diff
@@ -172,11 +172,54 @@ Identity EntityManagementFeatures::GetWorldOfModel(
...
+    return this->GenerateIdentity(modelId, it->second);
+  }
+  return this->GenerateInvalidId();
+}
+
+/////////////////////////////////////////////////
+Identity EntityManagementFeatures::GetNestedModel(
+  const Identity &_modelID, const std::string &_modelName) const
+{
+  auto modelInfo = this->ReferenceInterface<WorldInfo>(_modelID);
```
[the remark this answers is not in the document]
azeey: https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e5679df5d6698b29f31fc34d37428cbceb8cfc52

tpe/plugin/src/EntityManagement_TEST.cc
azeey: This actually failed and I realized I had fixed it in the next PR (#230), so I brought in the relevant patch in https://github.com/ignitionrobotics/ign-physics/pull/229/commits/e5679df5d6698b29f31fc34d37428cbceb8cfc52. Do you mind taking one more look?

tpe/plugin/src/Base.hh
```diff
@@ -88,26 +88,38 @@ class Base : public Implements3d<FeatureList<Feature>>
...
-      }
-      else if (it->second == _containerId)
-      {
-        ++counter;
+        auto idMapIt = _idMap.find(it->first);
+        // only count if the entity is found in the idMap. This makes sure we're
+        // counting only the entities with the correct EntityType
+        if (idMapIt != _idMap.end())
+        {
+          if (counter == _index)
```
adlarkin, line 109: I'm having a hard time wrapping my mind around the purpose of the `counter` variable in this method, would you mind helping me understand why we need it?

adlarkin (approved):
Everything looks good to me after the patch that was brought in via https://github.com/ignitionrobotics/ign-physics/commit/e5679df5d6698b29f31fc34d37428cbceb8cfc52. The updated `indexInContainerToId` method definitely makes things easier in `EntityManagementFeatures`. I just left one question I had, as I am trying to still wrapping my mind around some of these physics concepts.

tpe/plugin/src/Base.hh
azeey: Sure. `childIdToParentId`, is a map from child Entity IDs to parent Entity IDs. These are IDs, not indices. But we are trying to look up an index in the container identified by `_containerId`. To do so, we have to go through all the key-value pairs in `childToParentId` and count the elements whose values equal `_containerId` and whose keys are found in `_idMap`. `_idMap` is a container of specific types of entities. We need it because `_containerId` doesn't differentiate between links and models. For example, if we want to find the second link in a model, `_idMap` would be `Base::links`, but if we wanted to find the second model in a world `_idMap` would be `Base::models`.

Counting the elements in `childIdToParentId` works because it's an ordered map and the entity IDs of all the entity types are created in a globally increasing sequence. So, when iterating through `childIdToParentId`, we get entities in the order of creation.

adlarkin: > Counting the elements in childIdToParentId works because it's an ordered map

Ah, I think this was the key piece I was missing - I glanced over the fact that we are using `std::map` and not something like `std::unordered_map`. Thanks for the explanation, that helps a lot!

PR #234, opened 2021-03-18
Rename CanonicalLink to RootLink

### LINKED ISSUE #233: Should rename the term "Canonical link" in FreeGroup
## Desired behavior
The term "canonical link" in SDFormat refers to a link to which the model frame is attached. That is, there's a coordinate frame called the model frame and it has a constant pose offset from the implicit frame defined by the canonical link. On the other hand, the term "canonical link" in the context of `FreeGroup` refers to a link in the model that has no constraints. For DART, this link would be the root of a kinematic tree and it's parent joint would be `FreeJoint`, which has 6DOF. Thus it is possible that the "canonical link" of a model per SDFormat be different from the "canonical link" of the `FreeGroup` of the model

For example:
```xml
<?xml version="1.0"?>
<sdf version="1.7">
  <model name="M1">
    <link name="L1"/>
    <link name="L2"/>
    <joint name="J1" type="fixed">
      <parent>L2</parent>
      <child>L1</child>
    </joint>
  </model>
</sdf>
```
The canonical link of `M1` per SDFormat is the link `L1`, however the canonical link of the FreeGroup of `M1` is `L2` because `L1` and `L2` are constrained by a joint and `L2` is the parent. 

This is also exacerbated by the fact that starting in [SDFormat 1.7](http://sdformat.org/tutorials?tut=pose_frame_semantics&cat=specification&#canonical-link), the canonical link can be specified by the user to be any link in the model.

Since the term is overloaded, and the Gazebo/SDFormat version of canonical link existed (maybe implicitly) before the `FreeGroup` counterpart, we should come up with a new name for `FreeGroup`'s canonical link.

## Implementation suggestion
Rename the FreeGroup "canonical link" to "unconstrained link"

azeey:
# 🦟 Bug fix

Fixes #233 

## Summary
See #233 for context.
This deprecates the end-user facing function `CanonicalLink` adds a new one called `UnconstrainedLink`. 

**Note**: While the changes in this PR won't break downstream users of ign-physics, it will break any physics plugins the implement the `FreeGroup` feature.

## Checklist
- [ ] Added tests


adlarkin (commented):
Changes look OK to me. I'll let @scpeters do one last final pass since I'm still pretty new to `ign-physics`.

Should we add a note to the migration guide about this change so that other users who have a plugin that implements the `FreeGroup` feature know what updates to make to reflect this change?

scpeters (commented):
looks good, just need a migration guide entry

mxgrey:
I might make a suggestion about the naming:

`UnconstrainedLink` sounds like it refers to a link that does not have any constraints attached to it. I.e. it has no parent **nor** child joints.

If `Canonical` is not considered a suitable name anymore, then instead of `UnconstrainedLink` I might recommend calling this the `RootLink`. If it's called `RootLink` then that explicitly conveys that it doesn't have parent joints but may have child joints.

azeey:
> I might make a suggestion about the naming:
> 
> `UnconstrainedLink` sounds like it refers to a link that does not have any constraints attached to it. I.e. it has no parent **nor** child joints.
> 
> If `Canonical` is not considered a suitable name anymore, then instead of `UnconstrainedLink` I might recommend calling this the `RootLink`. If it's called `RootLink` then that explicitly conveys that it doesn't have parent joints but may have child joints.

`RootLink` make more sense. Renamed in https://github.com/ignitionrobotics/ign-physics/pull/234/commits/14a6b219edd2420a537f8d6cbc9ddb033a92ad33

azeey:
Ubuntu CI failed due to infra issues.

scpeters (approved):

PR #238, opened 2021-03-23
Add "ChangedWorldPose" to ForwardStep::Output's expected data

adlarkin:

# 🎉 New feature

Related to #223 

Possible solution for #219

## Summary

#223 added functionality to DART and TPE to write changed link data from an iteration to `ForwardStep::Output` so that downstream libraries like `ign-gazebo` can use this data for more efficient post-physics processing/updates (https://github.com/ignitionrobotics/ign-gazebo/pull/678). In order for downstream libraries to make this functionality optional, the changed link data that is written to by DART and TPE must be of type `ExpectedData` instead of `RequiredData` (that way, downstream libraries can see if this data was actually written to or not). This PR makes this change.

## Test it

Using this PR with https://github.com/ignitionrobotics/ign-gazebo/pull/678, users can run the [`ign-gazebo` physics system integration test](https://github.com/ignitionrobotics/ign-gazebo/blob/040b4a16fe162c167852c519895fac01543c1876/test/integration/physics_system.cc). This will use the output link data that `ign-physics` writes to in order to update the simulation state in `ign-gazebo`. All checks in the integration test should pass.

Users can then check out DART and TPE before these engines were modified to write output data (commit https://github.com/ignitionrobotics/ign-physics/commit/109de85195755f855b32cae088849ce17245ff61):
```
cd ign-physics
git checkout 109de85 -- dartsim/ tpe/
```
Re-run the physics integration test. This will use data internal to `ign-gazebo` (a local link pose cache in combination with the EntityComponentManager) to update the simulation state. All checks in the integration test should pass.

_If you'd like to verify that the steps I described above use both data written to from `ign-physics` and data internal to `ign-gazebo`, you can add some print statements in each `if`/`else` block [here](https://github.com/ignitionrobotics/ign-gazebo/blob/040b4a16fe162c167852c519895fac01543c1876/src/systems/physics/Physics.cc#L1730-L1783) to see what is being called._

## Checklist
- [x] Added tests

### DIFF
--- a/dartsim/src/SimulationFeatures.hh
+++ b/dartsim/src/SimulationFeatures.hh
@@ -26,6 +26,7 @@
 #include <ignition/physics/CanWriteData.hh>
 #include <ignition/physics/ForwardStep.hh>
 #include <ignition/physics/GetContacts.hh>
+#include <ignition/physics/SpecifyData.hh>
 
 #include "Base.hh"
 
@@ -39,7 +40,8 @@ struct SimulationFeatureList : FeatureList<
 > { };
 
 class SimulationFeatures :
-    public CanWriteRequiredData<SimulationFeatures, ForwardStep::Output>,
+    public CanWriteExpectedData<SimulationFeatures,
+      ExpectData<ChangedWorldPoses>>,
     public virtual Base,
     public virtual Implements3d<SimulationFeatureList>
 {
@@ -49,7 +51,7 @@ class SimulationFeatures :
       ForwardStep::State &_x,
       const ForwardStep::Input &_u) override;
 
-  public: void Write(WorldPoses &_poses) const;
+  public: void Write(ChangedWorldPoses &_changedPoses) const;
 
   public: std::vector<ContactInternal> GetContactsFromLastStep(
       const Identity &_worldID) const override;
--- a/include/ignition/physics/ForwardStep.hh
+++ b/include/ignition/physics/ForwardStep.hh
@@ -49,6 +49,15 @@ namespace ignition
       std::string annotation;
     };
 
+    /// \brief ChangedWorldPoses has the same definition as WorldPoses.
+    /// This type provides a way to keep track of which poses have changed in a
+    /// simulation step.
+    struct ChangedWorldPoses
+    {
+      std::vector<WorldPose> entries;
+      std::string annotation;
+    };
+
     struct Point
     {
       ignition::math::Vector3d point;
@@ -150,7 +159,7 @@ namespace ignition
 
       public: using Output = SpecifyData<
           RequireData<WorldPoses>,
-          ExpectData<Contacts, JointPositions> >;
+          ExpectData<ChangedWorldPoses, Contacts, JointPositions> >;
 
       public: using State = CompositeData;
 
--- a/dartsim/src/SimulationFeatures.cc
+++ b/dartsim/src/SimulationFeatures.cc
@@ -59,15 +59,15 @@ void SimulationFeatures::WorldForwardStep(
 
   // TODO(MXG): Parse input
   world->step();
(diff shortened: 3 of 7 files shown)


dartsim/src/SimulationFeatures.hh
```diff
@@ -51,6 +51,12 @@ class SimulationFeatures :
 
   public: void Write(WorldPoses &_poses) const;
 
+  public: void Write(ChangedWorldPoses &_changedPoses) const;
+
+  public: void Write(Contacts &_contacts) const;
+
+  public: void Write(JointPositions &_jointPositions) const;
```
adlarkin, line 58: Since the `SimulationFeatures` class now inherits `CanWriteExpectedData` instead of `CanWriteRequiredData` (`ChangedWorldPoses` is of type expected data), `Write` methods need to be implemented for all of the expected (and required) data types. All of these methods have an empty implementation except for `Write(ChangedWorldPoses &_changedPoses)`, but I don't think there's a better way to solve this problem of empty methods since we have to adhere to the interface.

include/ignition/physics/ForwardStep.hh
```diff
@@ -49,6 +49,12 @@ namespace ignition
       std::string annotation;
     };
 
+    struct ChangedWorldPoses
+    {
+      std::vector<WorldPose> entries;
+      std::string annotation;
+    };
```
adlarkin, line 56: This is a copy of `WorldPoses`, which is defined above. I had to make a copy of the definition instead of doing something like `using ChangedWorldPoses = WorldPoses` in order to avoid naming/scope collisions. Is this okay, or is there something else I can do to avoid this duplication? I can also add some documentation that explains why `WorldPoses` and `ChangeWorldPoses` have the same definition.

adlarkin (commented):
@azeey, just a few questions for you about the implementation.

dartsim/src/SimulationFeatures.hh
```diff
@@ -39,7 +39,7 @@ struct SimulationFeatureList : FeatureList<
 > { };
 
 class SimulationFeatures :
-    public CanWriteRequiredData<SimulationFeatures, ForwardStep::Output>,
+    public CanWriteExpectedData<SimulationFeatures, ForwardStep::Output>,
```
azeey, line 42: This can be limited to just ` ExpectData<ChangedWorldPoses>` and then we don't have to create empty implementations.

```suggestion
    public CanWriteExpectedData<SimulationFeatures, ExpectData<ChangedWorldPoses>>,
```

include/ignition/physics/ForwardStep.hh
azeey: I think it's okay. The types do have to be different (not aliases of each other). Adding a comment would suffice.

dartsim/src/SimulationFeatures.hh
adlarkin: https://github.com/ignitionrobotics/ign-physics/pull/238/commits/1433ddbc73999fc149e25dc206448f4ee7b52ac2

include/ignition/physics/ForwardStep.hh
adlarkin: Comment has been added in https://github.com/ignitionrobotics/ign-physics/pull/238/commits/1433ddbc73999fc149e25dc206448f4ee7b52ac2

dartsim/src/SimulationFeatures.hh
adlarkin: Thanks for the tip, I used this in https://github.com/ignitionrobotics/ign-physics/pull/238/commits/1433ddbc73999fc149e25dc206448f4ee7b52ac2 and the empty implementations should no longer be needed.

azeey (approved):
LGTM!

PR #231, opened 2021-03-17
Support FreeGroup features for nested models 

azeey:
# 🎉 New feature

## Summary
This enables FreeGroups to work with nested models

**Requires:**
* ~#220~
* ~#227~
* ~#228~
* ~#229~
* ~#230~

## Test it
* Run tests
* Run the `nested_models.sdf` example world in ign-gazebo and use the transform tool to move the model.
  * `ign gazebo --physics-engine ignition-physics-tpe-plugin -v 4 nested_model.sdf`
  * `ign gazebo --physics-engine ignition-physics-dartsim-plugin -v 4 nested_model.sdf`
 
Before

https://user-images.githubusercontent.com/206116/111461661-7b5c1180-86eb-11eb-9177-761d9ce3fd36.mp4

After

https://user-images.githubusercontent.com/206116/111461667-7e570200-86eb-11eb-8a1f-0a4327140cb5.mp4

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #246, opened 2021-04-13
Bumps in fortress : ign-physics5

chapulina:
See https://github.com/ignition-tooling/release-tools/issues/428

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -7,4 +7,4 @@ libignition-math6-eigen3-dev
 libignition-plugin-dev
 libignition-utils1-cli-dev
 libignition-utils1-dev
-libsdformat11-dev
+libsdformat12-dev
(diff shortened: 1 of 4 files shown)


chapulina:
@osrf-jenkins run tests again, they should work now

adlarkin (commented):
Can this comment be updated?

https://github.com/ignitionrobotics/ign-physics/blob/bump_fortress_ign-physics5/sdf/include/ignition/physics/sdf/ConstructNestedModel.hh#L29-L31

chapulina:
> Can this comment be updated?

I don't think it should say `sdf12` because model composition got into `sdf11`. I think the most appropriate action would be to update it when nested model support is complete, see #231. CC @azeey

adlarkin:
> I don't think it should say `sdf12` because model composition got into `sdf11`. I think the most appropriate action would be to update it when nested model support is complete, see #231

Correct, what I was wondering is if the comment could be deleted altogether. However, since it appears support isn't fully completefd yet (#231), I think it's fine as is. I just wasn't aware of the current status of nested model support, so I was asking just to be sure :+1:

adlarkin (approved):

PR #250, opened 2021-04-26
make detail symbols hidden

adlarkin:

## Summary
After noticing that a `detail` class in `ign-gazebo` was not marked as `hidden` (https://github.com/ignitionrobotics/ign-gazebo/issues/759), I looked through the other ignition repositories and found that there are a few detail symbols in `ign-physics` that should be marked as `hidden` as well.

## Checklist
- [ ] Added tests
- [ ] `codecheck` passed (See
  [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See
  [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+)
to support the maintainers

### DIFF
--- a/include/ignition/physics/detail/CompositeData.hh
+++ b/include/ignition/physics/detail/CompositeData.hh
@@ -32,7 +32,7 @@ namespace ignition
     /// \brief Struct which contains information about a data type within the
     /// CompositeData.
     /// \private
-    struct IGNITION_PHYSICS_VISIBLE CompositeData::DataEntry
+    struct IGNITION_PHYSICS_HIDDEN CompositeData::DataEntry
     {
       /// \brief Default constructor
       public: DataEntry();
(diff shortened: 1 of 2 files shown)


scpeters:
> @j-rivero I'm already seeing some issues in CI: https://github.com/ignitionrobotics/ign-physics/pull/250/checks?check_run_id=244178102
> 
> I'm not sure what we need to do here, but @chapulina mentioned: "We should be able to hide all the details without thinking twice. If this causes some issue on tests or examples, this probably means that those classes shouldn't be in detail."

from the CI results, it looks like we can't hide detail headers

PR #261, opened 2021-06-08
Clean up functions that trigger GCC9 warnings

mjcarroll:
Removes unnecessary copy constructors and insures that all destructors are marked `virtual` where appropriate.

### DIFF
--- a/tpe/lib/src/AABBTree.cc
+++ b/tpe/lib/src/AABBTree.cc
@@ -53,9 +53,7 @@ AABBTree::AABBTree()
 }
 
 //////////////////////////////////////////////////
-AABBTree::~AABBTree()
-{
-}
+AABBTree::~AABBTree() = default;
 
 //////////////////////////////////////////////////
 void AABBTree::AddNode(std::size_t _id, const math::AxisAlignedBox &_aabb)
--- a/tpe/lib/src/Collision.hh
+++ b/tpe/lib/src/Collision.hh
@@ -47,7 +47,7 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Collision : public Entity
   public: Collision(const Collision &_other);
 
   /// \brief Destructor
-  public: ~Collision();
+  public: virtual ~Collision();
 
   /// \brief Assignment operator
   /// \param[in] _other collision
--- a/tpe/lib/src/CollisionDetector.cc
+++ b/tpe/lib/src/CollisionDetector.cc
@@ -59,9 +59,7 @@ CollisionDetector::CollisionDetector()
 }
 
 //////////////////////////////////////////////////
-CollisionDetector::~CollisionDetector()
-{
-}
+CollisionDetector::~CollisionDetector() = default;
 
 //////////////////////////////////////////////////
 std::vector<Contact> CollisionDetector::CheckCollisions(
--- a/tpe/lib/src/Entity.hh
+++ b/tpe/lib/src/Entity.hh
@@ -60,7 +60,7 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Entity
   protected: explicit Entity(std::size_t _id);
 
   /// \brief Destructor
-  public: ~Entity();
+  public: virtual ~Entity();
 
   /// \brief Assignment operator
   /// \param[in] _other Other entity to copy from
--- a/tpe/lib/src/Link.hh
+++ b/tpe/lib/src/Link.hh
@@ -39,7 +39,7 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Link : public Entity
   public: explicit Link(std::size_t _id);
 
   /// \brief Destructor
-  public: ~Link() = default;
+  public: virtual ~Link() = default;
 
   /// \brief Add a collision
   /// \return Newly created Collision
--- a/tpe/lib/src/Model.hh
+++ b/tpe/lib/src/Model.hh
@@ -42,7 +42,7 @@ class IGNITION_PHYSICS_TPELIB_VISIBLE Model : public Entity
   public: explicit Model(std::size_t _id);
 
   /// \brief Destructor
-  public: ~Model();
+  public: virtual ~Model();
 
   /// \brief Add a link
   /// \return Newly created Link
--- a/tpe/lib/src/Shape.cc
+++ b/tpe/lib/src/Shape.cc
@@ -57,18 +57,6 @@ BoxShape::BoxShape() : Shape()
 }
 
 //////////////////////////////////////////////////
-BoxShape::BoxShape(const BoxShape &_other)
-  : Shape()
-{
-  *this = _other;
-}
-
-//////////////////////////////////////////////////
-BoxShape::~BoxShape()
-{
-}
-
-//////////////////////////////////////////////////
 Shape &BoxShape::operator=(const Shape &_other)
 {
   auto other = static_cast<const BoxShape *>(&_other);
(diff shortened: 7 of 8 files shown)


chapulina:
The `LinkFeatures` test failure seems to be introduced by this PR?

mjcarroll:

scpeters:
I think the test failure will be resolved by #275

chapulina (approved):
No new warnings or failures are introduced by this, let's get it in.

scpeters:
> No new warnings or failures are introduced by this, let's get it in.

I disagree; the LinkFeatures test had a pre-existing problem but it was passing, while this PR turns it into a failure.

[Build Status](https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-main-bionic-amd64/31/) https://build.osrfoundation.org/view/ign-fortress/job/ignition_physics-ci-main-bionic-amd64/31/testReport/junit/(root)/UNIT_LinkFeatures_TEST/test_ran/

I think we should either:

* disable the LinkFeatures test on `main` until #275 is merged forward
* revert this until #275 is merged forward

chapulina:
> the LinkFeatures test had a pre-existing problem but it was passing, while this PR turns it into a failure.

Humm I'm still not sure this PR is the one introducing the failure. `LinkFeatures` also fails with a segfault in https://github.com/ignitionrobotics/ign-physics/pull/272, which didn't include this PR yet. 

I can't reproduce the failure locally, but I suspect it's not because of this PR. Something may have changed upstream (on SDF?) and affected `main`.

scpeters:
> > the LinkFeatures test had a pre-existing problem but it was passing, while this PR turns it into a failure.
> 
> Humm I'm still not sure this PR is the one introducing the failure. `LinkFeatures` also fails with a segfault in #272, which didn't include this PR yet.
> 
> I can't reproduce the failure locally, but I suspect it's not because of this PR. Something may have changed upstream (on SDF?) and affected `main`.

the most recent GitHub action  for this PR merged into `main` passes on focal but fails on bionic:

* https://github.com/ignitionrobotics/ign-physics/actions/runs/1017007343

scpeters:
> > > the LinkFeatures test had a pre-existing problem but it was passing, while this PR turns it into a failure.
> > 
> > 
> > Humm I'm still not sure this PR is the one introducing the failure. `LinkFeatures` also fails with a segfault in #272, which didn't include this PR yet.
> > I can't reproduce the failure locally, but I suspect it's not because of this PR. Something may have changed upstream (on SDF?) and affected `main`.
> 
> the most recent GitHub action for this PR merged into `main` passes on focal but fails on bionic:
> 
> * https://github.com/ignitionrobotics/ign-physics/actions/runs/1017007343

I'm also having trouble reproducing it locally

chapulina:
> passes on focal but fails on bionic:

I can't reproduce on Bionic locally though

azeey:
I was able to reproduce it locally with dependencies installed from nightlies.

azeey:
Here's my hypothesis. The test has been somewhat broken because `sdf::World::Load` is called instead of `sdf::Root::Load`. A consequence of that is the error message:
```
[Err] [SDFFeatures.cc:511] The link of the parent frame [base_link] of joint [upper_joint] in model [double_pendulum_with_base] could not be resolved. The joint will not be constructed
[Err] [SDFFeatures.cc:517] Error Code 9: [:L530708400]: Msg: Frame has invalid pointer to FrameAttachedToGraph.
```

that indicates that the parent link of the joint could not be resolved, which is because a frame graph is not available when `sdf::World::Load` is called. But when printing the error message, it segfaults because of https://github.com/ignitionrobotics/sdformat/issues/590. The issue is fixed by https://github.com/ignitionrobotics/sdformat/pull/610, but that hasn't made it to `main` yet.

scpeters:
> Here's my hypothesis. The test has been somewhat broken because `sdf::World::Load` is called instead of `sdf::Root::Load`. A consequence of that is the error message:
> 
> ```
> [Err] [SDFFeatures.cc:511] The link of the parent frame [base_link] of joint [upper_joint] in model [double_pendulum_with_base] could not be resolved. The joint will not be constructed
> [Err] [SDFFeatures.cc:517] Error Code 9: [:L530708400]: Msg: Frame has invalid pointer to FrameAttachedToGraph.
> ```
> 
> that indicates that the parent link of the joint could not be resolved, which is because a frame graph is not available when `sdf::World::Load` is called. But when printing the error message, it segfaults because of [ignitionrobotics/sdformat#590](https://github.com/ignitionrobotics/sdformat/issues/590). The issue is fixed by [ignitionrobotics/sdformat#610](https://github.com/ignitionrobotics/sdformat/pull/610), but that hasn't made it to `main` yet.

https://github.com/ignitionrobotics/sdformat/pull/610 was merged forward to `main` in https://github.com/ignitionrobotics/sdformat/pull/613 on June 30th

azeey:
> [ignitionrobotics/sdformat#610](https://github.com/ignitionrobotics/sdformat/pull/610) was merged forward to `main` in [ignitionrobotics/sdformat#613](https://github.com/ignitionrobotics/sdformat/pull/613) on June 30th

Ah, you're right. So I dug a little deeper and I can now reproduce it with sdformat build from source. The trick is to build it with g++ 7.5 as it's done by the nightly debbuilder. My new hypothesis is that `sdf::operator<<(std::ostream, const std::Error&)` is inlined in ign-physics code because it's defined in `sdf/Error.hh`. And since ign-physics is compiled with g++ 8, there might be some weird ABI issues between symbols accessed by the inlined code and the ones available in `libsdformat.so`.

I see two options:

1. Move `sdf::operator<<(std::ostream, const std::Error&)` to `Error.cc` so it won't get inlined. 
2. Use g++8 in the sdformat nightly debbuilder job.

I have tested locally that both would work.

scpeters:
> > [ignitionrobotics/sdformat#610](https://github.com/ignitionrobotics/sdformat/pull/610) was merged forward to `main` in [ignitionrobotics/sdformat#613](https://github.com/ignitionrobotics/sdformat/pull/613) on June 30th
> 
> Ah, you're right. So I dug a little deeper and I can now reproduce it with sdformat build from source. The trick is to build it with g++ 7.5 as it's done by the nightly debbuilder. My new hypothesis is that `sdf::operator<<(std::ostream, const std::Error&)` is inlined in ign-physics code because it's defined in `sdf/Error.hh`. And since ign-physics is compiled with g++ 8, there might be some weird ABI issues between symbols accessed by the inlined code and the ones available in `libsdformat.so`.
> 
> I see two options:
> 
> 1. Move `sdf::operator<<(std::ostream, const std::Error&)` to `Error.cc` so it won't get inlined.
> 2. Use g++8 in the sdformat nightly debbuilder job.
> 
> I have tested locally that both would work.

I prefer moving it to `Error.cc`

scpeters:
> > > [ignitionrobotics/sdformat#610](https://github.com/ignitionrobotics/sdformat/pull/610) was merged forward to `main` in [ignitionrobotics/sdformat#613](https://github.com/ignitionrobotics/sdformat/pull/613) on June 30th
> > 
> > 
> > Ah, you're right. So I dug a little deeper and I can now reproduce it with sdformat build from source. The trick is to build it with g++ 7.5 as it's done by the nightly debbuilder. My new hypothesis is that `sdf::operator<<(std::ostream, const std::Error&)` is inlined in ign-physics code because it's defined in `sdf/Error.hh`. And since ign-physics is compiled with g++ 8, there might be some weird ABI issues between symbols accessed by the inlined code and the ones available in `libsdformat.so`.
> > I see two options:
> > 
> > 1. Move `sdf::operator<<(std::ostream, const std::Error&)` to `Error.cc` so it won't get inlined.
> > 2. Use g++8 in the sdformat nightly debbuilder job.
> > 
> > I have tested locally that both would work.
> 
> I prefer moving it to `Error.cc`

trying it out in https://github.com/ignitionrobotics/sdformat/pull/623

PR #272, opened 2021-06-24
4 ➡️ 5

chapulina:
# ➡️ Forward port

Port `ign-physics4` to `main`

Branch comparison: https://github.com/ignitionrobotics/ign-physics/compare/main...ign-physics4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)


mjcarroll:
The segfault in the LinkFeaturesTest is the same as I was getting here: https://github.com/ignitionrobotics/ign-physics/pull/261

I noticed that it seems to impact using sdformat from debs, but not from source on my system.

scpeters:
> The segfault in the LinkFeaturesTest is the same as I was getting here: #261
> 
> I noticed that it seems to impact using sdformat from debs, but not from source on my system.

I've looked into this test failure, and I believe there's a bug in the test itself. I'm adding a new API to libsdformat (see https://github.com/osrf/sdformat/pull/601 and https://github.com/osrf/sdformat/pull/602) to help detect when graphs are not properly loaded for a world, which should help with confirming the source of this bug

scpeters (approved):

PR #275, opened 2021-07-09
Add Gravity Feature, fix LinkFeatures_TEST

scpeters:
# 🎉 New feature

This adds a feature for getting and setting the gravity vector in a `World`. This was motivated by a failing test in #261.

## Summary

The [LinkFeatures_TEST](https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics4_4.1.0/dartsim/src/LinkFeatures_TEST.cc#L84-L91) currently loads an `sdf::World` object then copies it in order to modify the `<gravity/>` SDFormat element, with a comment noting "// TODO Add a world property feature to set gravity instead of this hack". In `ignition-physics3` and earlier, this was a useful hack, but it doesn't work in `ignition-physics4` due to an internal change in `libsdformat11`. I've illustrated the problem by adding some failing test expectations in e270f65 that call `sdf::World::ValidateGraphs`. Loading the `sdf::World` from an `sdf::ElementPtr` does not work anymore; you have to load an `sdf::Root` in order for the `FrameAttachedToGraph` and `PoseRelativeToGraph` to be properly loaded. I believe this is also related to the test failures observed in #261.

To fix the issue, I've added `World::GetGravity` and `World::SetGravity` APIs in a new `Gravity` Feature in `World.hh` (
94048ea). The APIs use `FrameSemantics.hh` so you can specify the `FrameID` for the coordinates in which the gravity vector is specified. I then updated the `LoadWorld` function in `LinkFeatures_TEST` in ec232ac so it can load the world properly. I removed the `LoadWorld` function from `ShapeFeatures_TEST` in 573f001 since it has the same flaw but is unused.

## Test it

Compile the code as of e270f65 and observe the failure of `UNIT_LinkFeatures_TEST`, and then observe that the test passes with the tip of the branch.

## Checklist
- [X] Added tests
- [X] Updated documentation (as needed)

### DIFF
--- a/dartsim/src/LinkFeatures_TEST.cc
+++ b/dartsim/src/LinkFeatures_TEST.cc
@@ -79,13 +79,15 @@ TestWorldPtr LoadWorld(
 {
   sdf::Root root;
   const sdf::Errors errors = root.Load(_sdfFile);
-  EXPECT_TRUE(errors.empty());
+  EXPECT_TRUE(errors.empty()) << errors;
   const sdf::World *sdfWorld = root.WorldByIndex(0);
   // Make a copy of the world so we can set the gravity property
   // TODO(addisu) Add a world property feature to set gravity instead of this
   // hack
   sdf::World worldCopy;
   worldCopy.Load(sdfWorld->Element());
+  auto graphErrors = worldCopy.ValidateGraphs();
+  EXPECT_EQ(0u, graphErrors.size()) << graphErrors;
 
   worldCopy.SetGravity(math::eigen3::convert(_gravity));
   return _engine->ConstructWorld(worldCopy);
--- a/dartsim/src/LinkFeatures_TEST.cc
+++ b/dartsim/src/LinkFeatures_TEST.cc
@@ -74,26 +74,6 @@ class LinkFeaturesFixture : public ::testing::Test
   protected: TestEnginePtr engine;
 };
 
-TestWorldPtr LoadWorld(
-    const TestEnginePtr &_engine,
-    const std::string &_sdfFile,
-    const Eigen::Vector3d &_gravity = Eigen::Vector3d{0, 0, -9.8})
-{
-  sdf::Root root;
-  const sdf::Errors errors = root.Load(_sdfFile);
-  EXPECT_TRUE(errors.empty()) << errors;
-  const sdf::World *sdfWorld = root.WorldByIndex(0);
-  EXPECT_NE(nullptr, sdfWorld);
-
-  auto graphErrors = sdfWorld->ValidateGraphs();
-  EXPECT_EQ(0u, graphErrors.size()) << graphErrors;
-
-  TestWorldPtr world = _engine->ConstructWorld(*sdfWorld);
-  EXPECT_NE(nullptr, world);
-  world->SetGravity(_gravity);
-  return world;
-}
-
 // A predicate-formatter for asserting that two vectors are approximately equal.
 class AssertVectorApprox
 {
@@ -116,6 +96,31 @@ class AssertVectorApprox
 
   private: double tol;
 };
+
+TestWorldPtr LoadWorld(
+    const TestEnginePtr &_engine,
+    const std::string &_sdfFile,
+    const Eigen::Vector3d &_gravity = Eigen::Vector3d{0, 0, -9.8})
+{
+  sdf::Root root;
+  const sdf::Errors errors = root.Load(_sdfFile);
+  EXPECT_TRUE(errors.empty()) << errors;
+  const sdf::World *sdfWorld = root.WorldByIndex(0);
+  EXPECT_NE(nullptr, sdfWorld);
+
+  auto graphErrors = sdfWorld->ValidateGraphs();
+  EXPECT_EQ(0u, graphErrors.size()) << graphErrors;
+
+  TestWorldPtr world = _engine->ConstructWorld(*sdfWorld);
+  EXPECT_NE(nullptr, world);
+  world->SetGravity(_gravity);
+
+  AssertVectorApprox vectorPredicate(1e-10);
+  EXPECT_PRED_FORMAT2(vectorPredicate, _gravity,
+                      world->GetGravity());
+
+  return world;
+}
 
 // Test setting force and torque.
 TEST_F(LinkFeaturesFixture, LinkForceTorque)
--- a/dartsim/src/LinkFeatures_TEST.cc
+++ b/dartsim/src/LinkFeatures_TEST.cc
@@ -125,7 +125,7 @@ TEST_F(LinkFeaturesFixture, LinkForceTorque)
   {
     AssertVectorApprox vectorPredicate(1e-10);
     EXPECT_PRED_FORMAT2(vectorPredicate, Eigen::Vector3d::Zero(),
-                        world->GetGravity(physics::FrameID::World()));
+                        world->GetGravity());
   }
 
   // Add a sphere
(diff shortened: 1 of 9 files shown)


dartsim/src/WorldFeatures.cc
```diff
@@ -80,6 +80,22 @@ const std::string &WorldFeatures::GetWorldCollisionDetector(const Identity &_id)
...
+/////////////////////////////////////////////////
+void WorldFeatures::SetWorldGravity(
+    const Identity &_id, const LinearVectorType &_gravity)
+{
+  auto world = this->ReferenceInterface<dart::simulation::World>(_id);
+  world->setGravity(_gravity);
+}
+
+/////////////////////////////////////////////////
+WorldFeatures::LinearVectorType WorldFeatures::GetWorldGravity(
```
chapulina, line 92: Can we return `const &`?

chapulina (commented):
How about adding unit tests to `WorldFeatures_TEST`?

dartsim/src/WorldFeatures.cc
scpeters: I think we should be careful with returning `const` references because I don't think it works well for values computed in variables on the stack. It works great for accessors to member variables, and indeed dart's `getGravity` method in [dart/simulation/World.hpp](https://github.com/dartsim/dart/blob/v6.10.1/dart/simulation/World.hpp#L124) returns a `const` reference, but changing this API to return a `const` reference would potentially constrain other implementations of this API

I could be wrong in my understanding of c++ here and am open to feedback, but that's my current reasoning for why not to return a `const` reference

chapulina: Ok, that makes sense. I see we have a few const ref returns in `ign-physics` already, but they're mainly returning strings, which I think are safer to assume are coming straight from member variables.

include/ignition/physics/World.hh
```diff
@@ -61,6 +62,81 @@ namespace ignition
...
+      public: template <typename PolicyT>
+      class Implementation : public virtual Feature::Implementation<PolicyT>
+      {
+        public: using LinearVectorType =
+            typename FromPolicy<PolicyT>::template Use<LinearVector>;
+
+        /// \brief Implementation API for setting the gravity vector, which is
+        /// expressed in the World frame..
+        /// \param[in] _id Identity of the world.
+        /// \param[in] _gravity Name of gravity.
```
azeey, line 127: nit: "Name"->"Value"?

azeey (approved):
LGTM! Thanks for adding this feature.

include/ignition/physics/World.hh
```diff
@@ -61,6 +62,81 @@ namespace ignition
...
+        /// to express _gravity.
+        public: void SetGravity(
+            const LinearVectorType &_gravity,
+            const FrameID &_forceInCoordinatesOf = FrameID::World());
+
+        /// \brief Get the World gravity vector as a Relative Gravity
+        /// (a quantity that contains information about the coordinates
+        /// in which it is expressed).
+        /// \return Relative Gravity vector.
+        public: RelativeGravityType GetGravity() const;
```
scpeters, line 105: it now occurs to me that the two `GetGravity` functions will be confusing for compilers since they can both be called with no arguments and have different return types. I'm inclined to remove this one that returns `RelativeGravityType`

azeey: It's weird that the compiler doesn't catch this. I'd also vote for removing the function that returns `RelativeGravityType`.

dartsim/src/LinkFeatures_TEST.cc
```diff
@@ -79,16 +81,17 @@ TestWorldPtr LoadWorld(
...
-  worldCopy.SetGravity(math::eigen3::convert(_gravity));
-  return _engine->ConstructWorld(worldCopy);
+  EXPECT_NE(nullptr, sdfWorld);
+
+  auto graphErrors = sdfWorld->ValidateGraphs();
+  EXPECT_EQ(0u, graphErrors.size()) << graphErrors;
+
+  TestWorldPtr world = _engine->ConstructWorld(*sdfWorld);
+  EXPECT_NE(nullptr, world);
+  world->SetGravity(_gravity);
```
azeey, line 93: Can you add an expectation on `GetGravity` just to exercise the API?

include/ignition/physics/World.hh
scpeters: 5324195

scpeters:
> How about adding unit tests to `WorldFeatures_TEST`?

done in 315424c

include/ignition/physics/World.hh
scpeters: removed in 7fa6ab2

dartsim/src/LinkFeatures_TEST.cc
scpeters: I had intended to do that, but then I realized that I would need to move the `AssertVectorApprox` class definition up above this function in order to use it in the expectation. So I stalled out on doing it. I'll do it now though

scpeters: done in https://github.com/ignitionrobotics/ign-physics/pull/275/commits/ef1012d9170f962756d8dfa7668d6a43a099eea9

dartsim/src/WorldFeatures_TEST.cc
```diff
@@ -98,6 +123,76 @@ TEST_F(WorldFeaturesFixture, CollisionDetector)
...
+                        initialLinkPosition,
+                        pos);
+  }
+
+  auto linkFrameID = link->GetFrameID();
+
+  // Get default gravity in link frame, which is pitched by pi/4
+  EXPECT_PRED_FORMAT2(vectorPredicate10,
+                      Eigen::Vector3d(6.92964645563, 0, -6.92964645563),
+                      world->GetGravity(linkFrameID));
```
scpeters, line 157: somehow this call to `GetGravity` is causing an assertion in KinematicsFeatures.cc

https://github.com/ignitionrobotics/ign-physics/blob/e21c8c9f94d364a10f0671ec1eb347f2c32c3348/dartsim/src/KinematicsFeatures.cc#L33-L40

related to call to `detail::Resolve` in `detail/World.hh`:

https://github.com/ignitionrobotics/ign-physics/blob/ef1012d9170f962756d8dfa7668d6a43a099eea9/include/ignition/physics/detail/World.hh#L82-L100

I think we should relax the assertion, unless I'm going something wrong here

scpeters: I've disabled the assertion in 37d09ae

azeey: I'm not sure if it hurts to ask the world frame what it's frame data is. Could be a problem for some physics engines? Looking at the code https://github.com/ignitionrobotics/ign-physics/blob/37d09ae717292623e6c1345efb26f66e458b0275/include/ignition/physics/detail/FrameSemantics.hh#L58-L62, maybe we should set `currentCoordinates = RotationType::Identity();`  if the `_relativeTo` is the world frame, as is done a couple of lines below.

scpeters: ooh, that's smart. I could leave the assert unchanged that way

scpeters: I've reverted the removal of the assert and followed your suggestion in 12b4cd442362efe76912558471d1f0d7e03d338e

azeey (approved):

PR #289, opened 2021-08-30
Fix TPE Link velocity not being updated and Model velocity not having any effect.

luca-della-vedova:
# 🦟 Bug fix

## Summary

This PR fixes two separate issues regarding TPE:

The first one was likely introduced in the performance optimization PRs (such as #223). Velocity control of models using TPE now doesn't work anymore. To try it out compare the results of:
`ign gazebo /home/luca/ws_edifice/src/ign-gazebo/examples/worlds/velocity_control.sdf --physics-engine ignition-physics-tpe-plugin`
and
`ign gazebo /home/luca/ws_edifice/src/ign-gazebo/examples/worlds/velocity_control.sdf `
The green vehicle does not move under TPE.
My suspicion is that it is due to the fact that instead of sending all the links we now send only updated links and since the velocity is updated at the model level the links themselves are not updated.
The fix I came up with involves iterating over the models in the update step (skipping static models) and, when a model had its pose updated, add all its links to the output of `Step` so that Gazebo will correctly update them. This fix makes the example work again, I'm not sure about the performance overhead but hopefully because of the static check it shouldn't be too large for worlds with a lot of static objects.

The second issue has probably always been there. As part of [ign-gazebo #966](https://github.com/ignitionrobotics/ign-gazebo/issues/966), I have been trying to implement velocity control using `LinearVelocity` and `AngularVelocity` components, however it seems velocity was always read as 0 in TPE.
I noticed that in [KinematicsFeature](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics4/tpe/plugin/src/KinematicsFeatures.cc#L48-L49) velocity is updated for models. However, in the physics update step we only send link information, not model information.
As a consequence I decided to also calculate and send Link velocity information in the global frame, this fixes the ign-gazebo issue (that can be tested with the `luca/linVelCmdTest` in `ign-gazebo`), as far as TPE is concerned.

I only managed to add a test for the second issue, a simple velocity loopback that fails before the PR and succeeds afterwards, I think the test for the last issue is a bit more complex since it involves the data being passed to `ign-gazebo`.

## Checklist
- [x] Added tests

### DIFF
--- a/tpe/plugin/src/SimulationFeatures.cc
+++ b/tpe/plugin/src/SimulationFeatures.cc
@@ -72,6 +72,8 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
   _changedPoses.entries.reserve(this->links.size());
 
   std::unordered_map<std::size_t, math::Pose3d> newPoses;
+  // Store the updated links to avoid duplicated entries in _changedPoses
+  std::unordered_set<std::size_t> updatedLinkIds;
 
   for (const auto &[id, info] : this->links)
   {
@@ -79,11 +81,11 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
     if (info)
     {
       const auto nextPose = info->link->GetPose();
-      auto iter = this->prevLinkPoses.find(id);
+      auto iter = this->prevEntityPoses.find(id);
 
       // If the link's pose is new or has changed, save this new pose and
       // add it to the output poses. Otherwise, keep the existing link pose
-      if ((iter == this->prevLinkPoses.end()) ||
+      if ((iter == this->prevEntityPoses.end()) ||
           !iter->second.Pos().Equal(nextPose.Pos(), 1e-6) ||
           !iter->second.Rot().Equal(nextPose.Rot(), 1e-6))
       {
@@ -91,6 +93,7 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
         wp.pose = nextPose;
         wp.body = id;
         _changedPoses.entries.push_back(wp);
+        updatedLinkIds.insert(id);
         newPoses[id] = nextPose;
       }
       else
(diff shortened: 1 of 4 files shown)


tpe/plugin/src/SimulationFeatures.cc
```diff
@@ -98,6 +98,40 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
...
+  for (const auto &[id, info] : this->models)
+  {
+    // make sure the model exists
+    if (info)
+    {
+      if (info->model->GetStatic())
+        continue;
+      const auto nextPose = info->model->GetPose();
+      // Note, now prevLinkPoses also containes prevModelPoses
+      // Data structure has been kept to avoid breaking ABI
```
iche033, line 112: `containes` -> `contains`

About ABI - I think we can change the data structure without breaking ABI since `SimulationFeatures.h` is in the `tpe/plugin/src` directory and it's not installed.

So maybe also rename `prevLinkPoses` to `prevEntityPoses`?

tpe/plugin/src/SimulationFeatures.cc
```diff
@@ -98,6 +98,40 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
...
+      if ((iter == this->prevLinkPoses.end()) ||
+          !iter->second.Pos().Equal(nextPose.Pos(), 1e-6) ||
+          !iter->second.Rot().Equal(nextPose.Rot(), 1e-6))
+      {
+        for (const auto &linkEnt : info->model->GetChildren())
+        {
+          WorldPose wp;
+          wp.pose = linkEnt.second->GetPose();
+          wp.body = linkEnt.second->GetId();
+          _changedPoses.entries.push_back(wp);
```
iche033, line 126: do you know if we'll end up pushing duplicate link entries into `_changedPoses`? I see that `_changedPoses` has a fixed size set [here](https://github.com/ignitionrobotics/ign-physics/blob/542f56d4d4e34daa3c6bbd2f6da44be762034cef/tpe/plugin/src/SimulationFeatures.cc#L72):

```
_changedPoses.entries.reserve(this->links.size());
```

So could this push more entries into the vector that the reserved size?

iche033 (commented):
nice, thanks for catching this. Confirms that the vehicles now move with vel cmds. Got a comment about potential duplicate link entries and vector size.

luca-della-vedova:
Fixed both the comments [here](https://github.com/ignitionrobotics/ign-physics/pull/289/commits/1f949e4e633305bb491433baa496f2fca3b04ecd). I changed the variable name to `prevEntityPoses` and added a `unordered_set` containing the `id` of links that have already been pushed to the vector to avoid adding duplicated entries.

iche033 (approved):
looks good to me.

adlarkin:
Thanks for the fixes, @luca-della-vedova! From what I recall with my recent testing on DART, these issues did not exist on DART - I believe that they only existed with TPE. Did you by chance test these things with DART to make sure that these issues do not exist with DART? If there are no issues with DART, then I believe we can close https://github.com/ignitionrobotics/ign-gazebo/issues/966 once we get some feedback on https://github.com/ignitionrobotics/ign-gazebo/issues/966#issuecomment-906606556.

PR #298, opened 2021-10-12
Refactor Bullet plugin implementation

joxoby:
The previous implementation was using instances if `btRigidBody` to represent links and instances derived from `btConstraint` to model the joints.

This implementation uses instances of `btMultiBody` multi-link models and `btRigidBody` for single-link ones.

`btMultiBody` is the Bullet implementation of Featherstone's algorithm, which uses a reduced coordinates representation.

Related ign-gazebo PR: https://github.com/ignitionrobotics/ign-gazebo/pull/1223


joxoby:
I've updated the base branch to `ign-physics5`, so we should update the label to `fortress`.

azeey:
Thanks for the contribution @joxoby. I see that this uses Bullet 3.17, but that version of Bullet is not available on Ubuntu Bionic or Focal. Is having 3.17 a strict requirement for the PR?

joxoby:
> Thanks for the contribution @joxoby. I see that this uses Bullet 3.17, but that version of Bullet is not available on Ubuntu Bionic or Focal. Is having 3.17 a strict requirement for the PR?

No, it's not. I reverted the change.

dgoel:
Just checking if there's any plan to review this PR?

azeey:
I haven't had time to review this, but I think it might have been removed from "In Progress" due to the migration to the new project board under the gazebosim org.

btw, @joxoby I see CI is failing. Anyway you can fix the failing tests?

PR #325, opened 2022-01-12
Fix Apache license version

jspricke:


ahcorde (changes requested):
Changes LGTM but this should target `ign-physics3`

jspricke:
> this should target `ign-physics3`

Why? I found them while doing a code review of `ign-physics5` for Debian. Also 5 is your default branch.

jspricke:
>Please refer to our [contribution guidelines](https://ignitionrobotics.org/docs/all/contributing#process):
>
>"_Pull requests should target the lowest possible supported version where the changes can be added in a backwards-compatible way (no API / ABI / behavior break in released branches)._"

That sounds uncommon, you may want to revisit how other projects doing 
it, i.e. backporting the changes.

>If that's too much of a hassle for you, we can backport the PR after it's merged though.

Please do, for example `dartsim/src/WorldFeatures_TEST.cc` is not in 
ign-physics3, so I would need to fill more then one PR?

>You'll still need to sign your commit for the DCO checker though, thanks!

Done.

scpeters (approved):

ahcorde (approved):

PR #328, opened 2022-03-08
Remove explicitly-defined copy constructor/operator

mjcarroll:
Cleans up warnings eg:

```
/home/mjcarroll/workspaces/ign_garden/src/ign-physics/tpe/lib/src/Shape.cc: In copy constructor ‘ignition::physics::tpelib::CapsuleShape::CapsuleShape(const ignition::physics::tpelib::CapsuleShape&)’:
/home/mjcarroll/workspaces/ign_garden/src/ign-physics/tpe/lib/src/Shape.cc:98:11: warning: implicitly-declared ‘ignition::physics::tpelib::CapsuleShape& ignition::physics::tpelib::CapsuleShape::operator=(const ignition::physics::tpelib::CapsuleShape&)’ is deprecated [-Wdeprecated-copy]
   98 |   *this = _other;
      |           ^~~~~~
```

Signed-off-by: Michael Carroll <michael@openrobotics.org>.

### DIFF
--- a/tpe/lib/src/Shape.cc
+++ b/tpe/lib/src/Shape.cc
@@ -11,7 +11,6 @@
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
- *
  * limitations under the License.
  *
 */
--- a/tpe/lib/src/Shape.cc
+++ b/tpe/lib/src/Shape.cc
@@ -11,6 +11,7 @@
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
+ *
  * limitations under the License.
  *
 */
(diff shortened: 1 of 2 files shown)


tpe/lib/src/Shape.cc
```diff
@@ -11,6 +11,7 @@
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
+ *
```
ahcorde, line 14: ?

ahcorde (changes requested):
do we need to tick-tock this methods?

tpe/lib/src/Shape.cc
mjcarroll: Whoops.

mjcarroll:
> do we need to tick-tock this methods?

The corresponding methods should be created by the compiler in these cases.

Furthermore, I'm not entirely convinced that some of these were doing the right thing.  The assignment operator is assuming that the RHS was of the same type, even when that wouldn't have been the same.

mjcarroll:
Not a great sign from codecov, either :sweat_smile: 

```
+ Misses       1398     1370      -28     
```

azeey (approved):

scpeters (approved):

Issue #332, opened 2022-03-14
Warning: implicitly-declared copy constructors are deprecated [-Wdeprecated-copy] 


chapulina:
## Environment
* OS Version:  Ubuntu Focal
* Source or binary build? Source
  
  

## Description
* Expected behavior:  No warnings
* Actual behavior:  Compiler warnings

## Steps to reproduce

Compile on Focal, or check Jenkins CI:

https://build.osrfoundation.org/job/ignition_physics-ci-pr_any-ubuntu_auto-amd64/1535/consoleFull

## Output

```
                 from /home/jenkins/workspace/ignition_physics-ci-pr_any-ubuntu_auto-amd64/ign-physics/src/SpecifyData_TEST.cc:22:
/home/jenkins/workspace/ignition_physics-ci-pr_any-ubuntu_auto-amd64/ign-physics/include/ignition/physics/SpecifyData.hh: In member function ‘ignition::physics::RequireData<StringData>& ignition::physics::RequireData<StringData>::operator=(const ignition::physics::RequireData<StringData>&)’:
/home/jenkins/workspace/ignition_physics-ci-pr_any-ubuntu_auto-amd64/ign-physics/include/ignition/physics/SpecifyData.hh:173:11: warning: implicitly-declared ‘ignition::physics::ExpectData<StringData>& ignition::physics::ExpectData<StringData>::operator=(const ignition::physics::ExpectData<StringData>&)’ is deprecated [-Wdeprecated-copy]
  173 |     class RequireData<Required> : public virtual ExpectData<Required>
      |           ^~~~~~~~~~~~~~~~~~~~~
```

chapulina:
I removed the requirement for `ignition_physics-ci-pr_any-ubuntu_auto-amd64` to pass before merging PRs into `main`, and updated 

* https://github.com/ignition-tooling/release-tools/issues/398

PR #335, opened 2022-03-15
Fix ExpectData compiler warnings

mjcarroll:
Addresses #332

### DIFF
--- a/include/ignition/physics/SpecifyData.hh
+++ b/include/ignition/physics/SpecifyData.hh
@@ -57,6 +57,9 @@ namespace ignition
       /// the newly copied MapData.
       public: ExpectData(const ExpectData &_other);
 
+      /// \brief Assignment operator.
+      public: ExpectData& operator=(const ExpectData &_other);
+
       /// TODO(anyone) Implement move constructor and assignment operator. Due
       /// to the multiple inheritence used to implement SpcifyData, care must be
       /// taken when implementing move constructs to avoid calling a move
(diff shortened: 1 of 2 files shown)


include/ignition/physics/SpecifyData.hh
```diff
@@ -48,15 +48,6 @@ namespace ignition
...
       public: ExpectData();
 
-      /// \brief Copy constructor.
-      /// The copy constructor of the base class, CompositeData, is called
-      /// before this copy constructor and it does the actual copying of the
-      /// data contained in the MapData object of `_other`. Thus, this copy
-      /// constructor simply calls the default constructor to initialize the
-      /// MapData iterator of this object to point the appropriate iterator in
-      /// the newly copied MapData.
-      public: ExpectData(const ExpectData &_other);
```
azeey, line 58: This was added in https://github.com/ignitionrobotics/ign-physics/pull/77 to fix a crash. Does it not occur anymore when removing it?

mjcarroll: It does, I meant to open this as a draft so that I didn't lose my train of thought.

I believe implementing the assignment operator is the correct solution to suppressing the warning in this case.

mjcarroll:

azeey (approved):
As I was reviewing this PR, I realized that the assumption that the base class' (`CompositeData`) copy ctor will be called only holds in certain scenarios. Here's a code sample that demonstrates when `CompositeData` copy ctor gets called: https://godbolt.org/z/6o6jWGvj8.

From that I see that having a default copy assignment operator would work as expected, but the way we have implemented `ExpectData`'s copy constructor doesn't work properly if we tried to use it directly to copy `ExpectData` objects. It works fine when we use the derived class `SpecifyData`. I'll open a separate issue about this.

scpeters (approved):

PR #344, opened 2022-04-14
Bumps in garden : ign-physics6

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/685


chapulina:
We don't need to block on Windows here because of 

* https://github.com/ignitionrobotics/ign-physics/issues/334

Merging!

Issue #25, opened 2019-10-18
Support of closed kinematic chains


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/issues/25)) by Diego Ferigo (Bitbucket: [dgferigo](https://bitbucket.org/%7B994b27a5-7d6d-42b2-8aa7-2ea82fbe7bf6%7D/)).**

----------------------------------------
Doing some experiments with the SDF of Minitaur quadruped \(taken from [KodlabPenn/kodlab\_gazebo](https://github.com/KodlabPenn/kodlab_gazebo)\), I noticed that the current state of the dart physics plugin does not yet support closed kinematic chains \(parallel robot\).

The four legs on Minitaur are composed of a [five-bar linkage](https://repository.upenn.edu/cgi/viewcontent.cgi?article=1766&context=ese_papers). The SDFs hosted in the GitHub repo work well in Gazebo 10 Classic. However, the simulation of the model in Ignition Gazebo does not apply the constraint to keep the linkage closed, and the links forming the legs fall apart as soon as they touch the ground.

I found [this TODO](https://github.com/ignitionrobotics/ign-physics/blob/39bc98ce499bdd400e1e23a66c29bbd16adda352/dartsim/src/FreeGroupFeatures.cc#L50:52) in the code, therefore I open this issue to track the status of the missing support.

peci1:
I vouch for resolving this issue. It is in SDF spec that the models can contain closed kinematic chains.

Until this is resolved, I'd expect a note on https://ignitionrobotics.org/docs/citadel/comparison#physics or https://ignitionrobotics.org/docs/all/release-features.

peci1:
According to https://github.com/dartsim/dart/issues/465 and https://dartsim.github.io/#dynamics it should be possible with DART, maybe with some workarounds.

Other possibly useful resources: https://dartsim.github.io/tutorials_collisions.html.

peci1:
We have a possible use for closed kinematic chains in the SubT virtual challenge (with the Absolem robot).

chapulina:
> I'd expect a note on

@peci1 , we'd appreciate a PR against the docs adding more rows to the comparison table with high-level features that you identify as missing. Thanks!

peci1:
> > I'd expect a note on
> 
[...]

PR #345, opened 2022-04-28
Add `WeldJointConstraint` for fixed joints if the parent is not a FreeJoint

arjo129:

# 🎉 New feature

Allows for Fixed Joints to be created when the parent is not a FreeJoint. This is required when we try to simulate multiple suction grippers gripping onto the same object.

:warning: :warning: :warning: :warning: 
image

## Summary

Currently, if we try to attach two models to the same item using fixed joints, we are unable to do so. This is because Dart's WeldJoint requires that the rigid body skeletons to form a tree. We can solve this by using WeldJointConstraints instead and that is what I'm trying to use.

## Test it
I've introduced a unit test [here](https://github.com/ignitionrobotics/ign-physics/blob/54c29f1d0d871cd4ddd04f144344475a9314a227/dartsim/src/JointFeatures_TEST.cc#L896).
2022-04-28T16:51:49 107398391
Essentially what happens is there are two floating boxes and a box in the middle that's resting. We start the system out by creating the two fixed joints between the box resting on the big box and the floating box. The middle box will now have two parents. However there should be no movement as the middle box will be holding the other two boxes that are floating in mid air. We run this for 100 steps to make sure that there is no movement. This is because the middle box is holding on to the two side boxes. Then we release the joints the two boxes should fall away.

## TODO
Since this affects the physics in a lot of places it will be important to do the following before merging
- [x] Check performance - Seems to be able to do realtime speeds in the example shown below
- [x] Figure out how to cast joints to weld contraints - Skip casting if is a constraint
- [x] Do an upstream integration test.
- [x] Get a :heavy_check_mark: in the CI. - Good ol' :window: not passing
- [ ] Check other affected functions.
- [ ] Find out how to deal with `ObjectToId` as weld constraints are not added to this list. - Pending reviewer feedback.

### Real world gripping demo
Manual integration test for this behaviour (skip to 1min mark):
https://user-images.githubusercontent.com/542272/167532762-380d3fa8-8f51-4b51-91ce-4a5fdc439e55.mp4

## Checklist
- [x] Added tests

### DIFF
--- a/dartsim/src/Base.hh
+++ b/dartsim/src/Base.hh
@@ -231,7 +231,7 @@ class Base : public Implements3d<FeatureList<Feature>>
   public: using DartBodyNodePtr = dart::dynamics::BodyNodePtr;
   public: using DartJoint = dart::dynamics::Joint;
   public: using DartJointPtr = dart::dynamics::JointPtr;
-  public: using DartWeldJointConsPtr =dart::constraint::WeldJointConstraintPtr;
+  public: using DartWeldJointConsPtr = dart::constraint::WeldJointConstraintPtr;
   public: using DartShapeNode = dart::dynamics::ShapeNode;
   public: using DartShapeNodePtr = std::shared_ptr<DartShapeNode>;
   public: using ModelInfoPtr = std::shared_ptr<ModelInfo>;
@@ -381,14 +381,15 @@ class Base : public Implements3d<FeatureList<Feature>>
     this->joints.idToObject[id] = std::make_shared<JointInfo>();
     this->joints.idToObject[id]->constraint = _joint;
     this->joints.idToObject[id]->type = JointInfo::JointType::CONSTRAINT;
-    //this->joints.objectToID[_joint] = id;
-    //dart::dynamics::SimpleFramePtr jointFrame =
-    //    dart::dynamics::SimpleFrame::createShared(
-    //        _joint->getChildBodyNode(), _joint->getName() + "_frame",
-    //        _joint->getTransformFromChildBodyNode());
+    // TODO(arjo): Refactor the joints.objectToId to support constraints.
+    // this->joints.objectToID[_joint] = id;
+    // dart::dynamics::SimpleFramePtr jointFrame =
+    //     dart::dynamics::SimpleFrame::createShared(
+    //         _joint->getChildBodyNode(), _joint->getName() + "_frame",
+    //         _joint->getTransformFromChildBodyNode());
 
-    //this->joints.idToObject[id]->frame = jointFrame;
-    //this->frames[id] = this->joints.idToObject[id]->frame.get();
+    // this->joints.idToObject[id]->frame = jointFrame;
+    // this->frames[id] = this->joints.idToObject[id]->frame.get();
 
     return id;
   }
--- a/dartsim/src/Base.hh
+++ b/dartsim/src/Base.hh
@@ -21,6 +21,7 @@
 #include <dart/dynamics/BodyNode.hpp>
 #include <dart/dynamics/SimpleFrame.hpp>
 #include <dart/dynamics/Skeleton.hpp>
+#include <dart/constraint/WeldJointConstraint.hpp>
 #include <dart/simulation/World.hpp>
 
 #include <memory>
@@ -73,6 +74,15 @@ struct JointInfo
 {
   dart::dynamics::JointPtr joint;
   dart::dynamics::SimpleFramePtr frame;
+
+  enum JointType
+  {
+    JOINT,
+    CONSTRAINT
+  };
+
+  JointType type{JOINT};
+  dart::constraint::WeldJointConstraintPtr constraint;
 };
 
 struct ShapeInfo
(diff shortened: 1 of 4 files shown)


dartsim/src/Base.hh
```diff
@@ -364,6 +376,25 @@ class Base : public Implements3d<FeatureList<Feature>>
...
+    this->joints.idToObject[id]->type = JointInfo::JointType::CONSTRAINT;
+    // TODO(arjo): Refactor the joints.objectToId to support constraints.
+    // this->joints.objectToID[_joint] = id;
+    // dart::dynamics::SimpleFramePtr jointFrame =
+    //     dart::dynamics::SimpleFrame::createShared(
+    //         _joint->getChildBodyNode(), _joint->getName() + "_frame",
+    //         _joint->getTransformFromChildBodyNode());
+
+    // this->joints.idToObject[id]->frame = jointFrame;
+    // this->frames[id] = this->joints.idToObject[id]->frame.get();
```
iche033, line 393: remove this code or is this relevant to the todo comment?

dartsim/src/JointFeatures.cc
```diff
@@ -466,9 +500,16 @@ Identity JointFeatures::AttachFixedJoint(
   {
     // child already has a parent joint
     // TODO(scpeters): use a WeldJointConstraint between the two bodies
```
iche033, line 502: remove this todo comment now that the functionality is implemented?

dartsim/src/Base.hh
```diff
@@ -353,6 +364,7 @@ class Base : public Implements3d<FeatureList<Feature>>
     this->joints.idToObject[id] = std::make_shared<JointInfo>();
     this->joints.idToObject[id]->joint = _joint;
     this->joints.objectToID[_joint] = id;
+     this->joints.idToObject[id]->type = JointInfo::JointType::JOINT;
```
iche033, line 367: indentation: extra space here

iche033 (commented):
the changes look good to me and it's now possible to create more than one detachable joints with the same link in our application as demonstrated in https://github.com/osrf/mbzirc/pull/108.

I think it would be good to have one of the physics experts take a look at these changes as well

azeey:
I think this can be a way to solve https://github.com/gazebosim/gz-physics/issues/25. In Gazebo classic, I believe kinematic loops were supported by breaking one of the links in the loop in half and attaching a constraint, which is similar to what's being done here except that the addition of the `WeldConstraint` only happens here when handling fixed joints. @scpeters, do you think we can make this the recommended way of dealing with kinematic loops where the user creates the fixed joint that creates the loop which triggers ign-physics to create a WeldConstraint? It's more of a manual approach compared to what's done in Gazebo classic.

dartsim/src/JointFeatures.cc
```diff
@@ -466,9 +500,16 @@ Identity JointFeatures::AttachFixedJoint(
...
-    return this->GenerateInvalidId();
+    auto worldId = this->GetWorldOfModelImpl(
```
arjo129, line 503: For one of the Physics folks: Should we snap the model to the joint position as well in this case? AddJoint automatically snaps two items together. Adding a constraint however doe not do that.

dartsim/src/Base.hh
arjo129: Yeah honestly I'm not sure where `this->frames`  and `this->jointsidToObject` is used. Would be great if someone who knew  the physics subsystem well could advice on whether we need it.

arjo129:
> I think this can be a way to solve #25. In Gazebo classic, I believe kinematic loops were supported by breaking one of the links in the loop in half and attaching a constraint, which is similar to what's being done here except that the addition of the `WeldConstraint` only happens here when handling fixed joints. @scpeters, do you think we can make this the recommended way of dealing with kinematic loops where the user creates the fixed joint that creates the loop which triggers ign-physics to create a WeldConstraint? It's more of a manual approach compared to what's done in Gazebo classic.

I'm open to reworking this to follow an approach closer to that of gazebo classic. To keep things automated we could potentially add a joint, a bodyNode and a weld constraint instead of just the weld constraint.

scpeters:
see https://github.com/gazebosim/gz-physics/pull/352

azeey (changes requested):
I think there are some thorny problems with this approach. The main issue being the fact that we don't have a `dart::dynamics::Joint` in the `JointInfo` object we associate with the Joint entity. There are many features (mostly in `JointFeatures.cc`) that depend on that joint pointer being valid. Using these features would cause a segfault in the current state of things. For example, `EntityManagementFeatures::GetJointName` [calls `dart::dynamics::Joint::getName`](https://github.com/gazebosim/gz-physics/blob/e91c58a49522cfdc5793fe8b980566eb26531ba4/dartsim/src/EntityManagementFeatures.cc#L554), but that wouldn't work if `this->ReferenceInterface<JointInfo>(_jointID)->joint` is `nullptr`. I don't think `gz-sim` calls `GetName` on a joint entity, so that might be why it's working fine right now. However, the same applies to `JointFeatures::GetJointPosition` `JointFeatures::GetJointVelocity`, etc. and if we added the `JointStatePublisher` system to the model that contains a kinematic loop, I'm not sure if it would work. Since we know it's a fixed joint, we can add conditionals in those functions that return 0 if the joint is a `WeldJointConstraint`, but we'd need to something similar in all the places `this->ReferenceInterface<JointInfo>(_jointID)->joint` is used and we would need some extra bookkeeping.

We also used fixed joints to model force/torque sensors. `JointFeatures::GetJointTransmittedWrenchInJointFrame` would have to be reworked to support `WeldJointConstraint`s.

I haven't reviewed @scpeters' alternative PR, but that may not suffer from the same issues as this since the `WeldJointConstraint` would be an internal object that is not associated with a Joint entity.

scpeters:
alternative fix merged in #352

PR #351, opened 2022-06-05
dartsim: fix handling inertia matrix pose rotation

scpeters:
# 🦟 Bug fix

Fixes a bug in moment of inertia calculation

## Summary

When loading a model from SDF in the dartsim plugin, the moment of inertia matrix is currently applying any rotations in the `//inertial/pose` two times, since the rotations are applied explicitly in SDFFeatures.cc, but they are already applied in [math::Inertial::Moi](https://github.com/gazebosim/gz-math/blob/ign-math6/include/ignition/math/Inertial.hh#L131-L132).

I confirmed the bug with a slightly modified version of the [inertia_rotations](https://github.com/osrf/gazebo/blob/gazebo11/test/worlds/inertia_rotations.world) test world from gazebo classic. That world has multiple 1x4x9 boxes with identical inertia values that are expressed with different inertial pose rotations. The boxes are initially standing up on a ground plane. Setting gravity to `[2, 0, -9.8]` causes the boxes to tip over, and they should move at the same rate and hit the ground at the same time. This works properly in gazebo-classic with DART but not fortress or citadel.

I'll upload a copy of the example world for use with fortress as well as screen captures of my testing and then mark this ready for review.

* https://github.com/gazebosim/gz-physics/blob/fd4fead203ef5da7743a48e501e575a0c688cbaf/dartsim/worlds/inertia_rotations_ign.world
* inertia_pose_rotations_ign_fortress
* inertia_pose_rotations_ign_gazebo11

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -591,11 +591,6 @@ Identity SDFFeatures::ConstructSdfLink(
   const ignition::math::Inertiald &sdfInertia = _sdfLink.Inertial();
   bodyProperties.mInertia.setMass(sdfInertia.MassMatrix().Mass());
 
-  // TODO(addisu) Resolve the pose of inertials when frame information is
-  // availabile for ignition::math::Inertial
-  const Eigen::Matrix3d R_inertial{
-        math::eigen3::convert(sdfInertia.Pose().Rot())};
-
   const Eigen::Matrix3d I_link = math::eigen3::convert(sdfInertia.Moi());
 
   bodyProperties.mInertia.setMoment(I_link);
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -596,10 +596,7 @@ Identity SDFFeatures::ConstructSdfLink(
   const Eigen::Matrix3d R_inertial{
         math::eigen3::convert(sdfInertia.Pose().Rot())};
 
-  const Eigen::Matrix3d I_link =
-      R_inertial
-      * math::eigen3::convert(sdfInertia.Moi())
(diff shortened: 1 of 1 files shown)


dartsim/src/SDFFeatures.cc
```diff
@@ -596,10 +596,7 @@ Identity SDFFeatures::ConstructSdfLink(
   const Eigen::Matrix3d R_inertial{
```
chapulina, line 596: This variable isn't used anymore. And I wonder if the TODO above still applies?

chapulina (commented):
Thanks for the fix, I was wondering about this. Is there a way we can write a test here to catch that?

Also, would it be possible to target this at Citadel?

JoanAguilar (approved):
Looks good to me. I second the idea of having an automated test to catch future regressions.

scpeters:
> Also, would it be possible to target this at Citadel?

yes, I'll target this at Citadel

scpeters:
> Thanks for the fix, I was wondering about this. Is there a way we can write a test here to catch that?

> I second the idea of having an automated test to catch future regressions.

it will be easiest to write the test using the common test framework, which is now added in Garden, so I'll make a separate PR for garden once this is merged forward

dartsim/src/SDFFeatures.cc
chapulina: Done in d56f652e1b9d51646403229ed45e7d2496d8fc99

chapulina:
> yes, I'll target this at Citadel

Let's merge this into Fortress so it catches the next bus to Garden. We should backport it to Citadel soon.

chapulina (approved):

PR #347, opened 2022-05-17
Try creating a loop joint

arjo129:
🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

# 🎉 New feature

Closes #<NUMBER>

## Summary
This PR supercedes #345 with a more general solution that would also help address #25 and solve a lot of issues related to casting joints. Open to feedback.

# TODO
- [ ] clean up all the excess joints and links  created during the process of loop joint creation.
- [ ] fix SetJointTransformChild to use only the joint.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>
## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #352, opened 2022-06-05
dartsim: support non-tree kinematics in AttachFixedJoint

scpeters:
# 🦟 Bug fix

Alternative to #345 and #347 

## Summary

With inspiration from #345 and #347, this implements a variation on the approach from gazebo-classic for handling non-tree kinematics (including kinematic loops) with the reduced coordinate Featherstone formulation by splitting links and welding them back together with a constraint. This adds data to the `LinkInfo` struct and a `SplitAndWeldLink` helper function to `Base.hh`. The test made by @arjo129 is included as well.

This only implements support for kinematic loops in the `AttachFixedJoint` method, but it should be applicable in other `Attach*Joint` methods and the `ConstructSdfJoint` method to support non-tree kinematics in more cases.

This implementation differs slightly from the approach in gazebo-classic. In gazebo11, all joints that are created dynamically using [gazebo::physics::Model::CreateJoint](http://osrf-distributions.s3.amazonaws.com/gazebo/api/11.0.0/classgazebo_1_1physics_1_1Model.html#aaf0402c1e9df1bc45ccecc135bf576af) use the approach of splitting and welding a link, even if the joint could simply be added as part of the existing Skeleton (see the call to `DARTModelPrivate::CreateLoopJointAndNodePair` in `DARTModel::CreateJointHelper` in [DARTModel.cc](https://github.com/osrf/gazebo/blob/gazebo11_11.11.0/gazebo/physics/dart/DARTModel.cc#L372-L373)). Within [DARTModlPrivate::CreateLoopJointAndNodePair](https://github.com/osrf/gazebo/blob/gazebo11_11.11.0/gazebo/physics/dart/DARTModelPrivate.hh#L147-L162), a new joint and node pair are created for the desired joint type by calling `DARTModelPrivate::CreateJointAndNodePair` and then welded to the desired child link with blended inertia by calling `DARTLink::AddSlaveBodyNode`.

In this implementation, the split-and-weld approach is only taken if the target child link already has a parent joint that is not FreeJoint. To minimize changes to the existing code, when an existing parent joint is detected, the split-and-weld is done first, by creating a new FreeJoint and node pair and passing it to the `SplitAndWeldLink` helper function, which blends the inertia between the nodes and creates a WeldJointConstraint. The newly welded body node is then moved to a WeldJoint using the `moveTo` API. This move requires the transform for the WeldJointConstraint to be updated to ensure consistency.

UPDATE: with thanks to @azeey for the changes in #367, the welded links are now properly removed when joints are detached.

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/JointFeatures_TEST.cc
+++ b/dartsim/src/JointFeatures_TEST.cc
@@ -883,6 +883,142 @@ TEST_F(JointFeaturesFixture, JointAttachDetach)
   EXPECT_NEAR(0.0, dartBody2->getLinearVelocity().z(), 1e-3);
 }
 
+////////////////////////////////////////////////
+// Essentially what happens is there are two floating boxes and a box in the
+// middle that's resting. We start the system out by creating the two
+// fixed joints between the boxes resting on the big box. The middle box will
+// now have two parents. However there should be no movement as the middle box
+// will be holding the other two boxes that are floating in mid air. We run
+// this for 100 steps to make sure that there is no movement. This is because
+// the middle box is holding on to the two side boxes. Then we release the
+// joints the two boxes should fall away.
+TEST_F(JointFeaturesFixture, JointAttachMultiple)
+{
+  sdf::Root root;
+  const sdf::Errors errors =
+      root.Load(TEST_WORLD_DIR "joint_constraint.sdf");
+  ASSERT_TRUE(errors.empty()) << errors.front();
+
+  auto world = this->engine->ConstructWorld(*root.WorldByIndex(0));
+  dart::simulation::WorldPtr dartWorld = world->GetDartsimWorld();
+  ASSERT_NE(nullptr, dartWorld);
+
+  // M1 and M3 are floating boxes
+  const std::string modelName1{"M1"};
+  const std::string modelName2{"M2"};
+  const std::string modelName3{"M3"};
+  const std::string bodyName{"link"};
+
+  auto model1 = world->GetModel(modelName1);
+  auto model2 = world->GetModel(modelName2);
+  auto model3 = world->GetModel(modelName3);
+
+  auto model1Body = model1->GetLink(bodyName);
+  auto model2Body = model2->GetLink(bodyName);
+  auto model3Body = model3->GetLink(bodyName);
+
+  const dart::dynamics::SkeletonPtr skeleton1 =
+      dartWorld->getSkeleton(modelName1);
+  const dart::dynamics::SkeletonPtr skeleton2 =
+      dartWorld->getSkeleton(modelName2);
+  const dart::dynamics::SkeletonPtr skeleton3 =
+      dartWorld->getSkeleton(modelName3);
+  ASSERT_NE(nullptr, skeleton1);
+  ASSERT_NE(nullptr, skeleton2);
+  ASSERT_NE(nullptr, skeleton3);
+
+  auto *dartBody1 = skeleton1->getBodyNode(bodyName);
+  auto *dartBody2 = skeleton2->getBodyNode(bodyName);
+  auto *dartBody3 = skeleton3->getBodyNode(bodyName);
+
+  ASSERT_NE(nullptr, dartBody1);
+  ASSERT_NE(nullptr, dartBody2);
+  ASSERT_NE(nullptr, dartBody3);
+
+  const math::Pose3d initialModel1Pose(0, -0.2, 0.45, 0, 0, 0);
+  const math::Pose3d initialModel2Pose(0, 0.2, 0.45, 0, 0, 0);
+  const math::Pose3d initialModel3Pose(0, 0.6, 0.45, 0, 0, 0);
+
+  EXPECT_EQ(initialModel1Pose,
+            math::eigen3::convert(dartBody1->getWorldTransform()));
+  EXPECT_EQ(initialModel2Pose,
+            math::eigen3::convert(dartBody2->getWorldTransform()));
+  EXPECT_EQ(initialModel3Pose,
+            math::eigen3::convert(dartBody3->getWorldTransform()));
+
+  physics::ForwardStep::Output output;
+  physics::ForwardStep::State state;
+  physics::ForwardStep::Input input;
+
+  // Create the first joint. This should be a normal fixed joint.
+  const auto poseParent1 = dartBody1->getTransform();
+  const auto poseChild1 = dartBody2->getTransform();
+  auto poseParentChild1 = poseParent1.inverse() * poseChild1;
+  auto fixedJoint1 = model2Body->AttachFixedJoint(model1Body);
+  fixedJoint1->SetTransformFromParent(poseParentChild1);
+
+  EXPECT_EQ(initialModel1Pose,
+            math::eigen3::convert(dartBody1->getWorldTransform()));
+  EXPECT_EQ(initialModel2Pose,
+            math::eigen3::convert(dartBody2->getWorldTransform()));
+  EXPECT_EQ(initialModel3Pose,
+            math::eigen3::convert(dartBody3->getWorldTransform()));
+
+  // Create the second joint. This should be a WeldJoint constraint
+  const auto poseParent2 = dartBody3->getTransform();
+  const auto poseChild2 = dartBody2->getTransform();
+  auto poseParentChild2 = poseParent2.inverse() * poseChild2;
+  auto fixedJoint2 = model2Body->AttachFixedJoint(model3Body);
+  fixedJoint2->SetTransformFromParent(poseParentChild2);
+
+  EXPECT_EQ(initialModel1Pose,
+            math::eigen3::convert(dartBody1->getWorldTransform()));
+  EXPECT_EQ(initialModel2Pose,
+            math::eigen3::convert(dartBody2->getWorldTransform()));
+  EXPECT_EQ(initialModel3Pose,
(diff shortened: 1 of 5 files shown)


arjo129 (dismissed):
I've added the unit tests in but they do not pass. Neither does the down stream pr in osrf/mbzirc#108 work. I'm not too sure whats wrong with it. Either ways I'm not on mbzirc right now. I believe @aaronchongth is taking over this aspect of it.

aaronchongth:
I might be wrong, but could the unit test that was ported over not be related to kinetic loops since both the parents of the added fixed joints are free floating and not attached to each other?

scpeters:
> I might be wrong, but could the unit test that was ported over not be related to kinetic loops since both the parents of the added fixed joints are free floating and not attached to each other?

I think the test is ok. It may not precisely be a "loop", but it doesn't follow the tree structure because the middle box will have two parents after attaching the fixed joints.

I think I have a fix in https://github.com/gazebosim/gz-physics/pull/352/commits/21a608d08188eaef02b4b9433c9c088d57be5f84

scpeters:
I've updated the description and marked this ready for review

aaronchongth:
Thanks for explaining @scpeters !

Attaching works like a charm! I am having trouble detaching the models though, it looks like one of the models will refuse to detach.

https://user-images.githubusercontent.com/5383623/172987447-0165fcd3-7131-44a4-997c-f5e9b70cd4b9.mp4

It looks like the first parent that attaches, is unable to detach afterwards, but the second parent is free to attach and detach. After detaching, both the still-attached parent and the model will be stuck in space somehow, until the other previously detached parent bumps into it.

I tried dropping shapes onto the stuck platform, but it doesn't budge.

dartsim/src/JointFeatures_TEST.cc
```diff
@@ -883,6 +883,142 @@ TEST_F(JointFeaturesFixture, JointAttachDetach)
...
+  fixedJoint2->SetTransformFromParent(poseParentChild2);
+
+  EXPECT_EQ(initialModel1Pose,
+            math::eigen3::convert(dartBody1->getWorldTransform()));
+  EXPECT_EQ(initialModel2Pose,
+            math::eigen3::convert(dartBody2->getWorldTransform()));
+  EXPECT_EQ(initialModel3Pose,
+            math::eigen3::convert(dartBody3->getWorldTransform()));
+
+  const std::size_t numSteps = 100;
```
aaronchongth, line 981: Regarding the issue I mentioned, it can be recreated in this test if `numSteps` is bumped up to `100000`

scpeters: I've updated the test in https://github.com/gazebosim/gz-physics/pull/352/commits/fb11680f89dad22d10405d2db88908072e6907c5 to expect the falling velocity to be linearly increasing with time, which exposes the failure of on of the boxes to fall after detaching.

scpeters: the `DetachJoint` function is exiting [here](https://github.com/gazebosim/gz-physics/blob/fb11680f89dad22d10405d2db88908072e6907c5/dartsim/src/JointFeatures.cc#L380-L383) since the welded node is not in the `links` data structure

scpeters: potential fix in https://github.com/gazebosim/gz-physics/pull/352/commits/fe92bd5f1ac0b9895be2acd5adea0e3d6703cc9e

aaronchongth:
After testing it more downstream, I faced some interesting behaviors when detaching the models one after another. Consistently, detaching one of the joints does not seem to do anything, while detaching the other one causes both to be detached.

https://user-images.githubusercontent.com/5383623/173373967-02499037-0aa4-45cf-9f22-9bd6adb00e54.mp4

https://user-images.githubusercontent.com/5383623/173377309-a246b866-a641-4952-aec7-67e9a6dcbde5.mp4

 I [duplicated the test](https://github.com/gazebosim/gz-physics/blob/aaron/detach_joints/dartsim/src/JointFeatures_TEST.cc#L1028-L1319) in this PR to detach one at a time, but could not seem to replicate this behvaior. I was wondering if you knew what might be the cause?

iche033:
> After testing it more downstream, I faced some interesting behaviors when detaching the models one after another. Consistently, detaching one of the joints does not seem to do anything, while detaching the other one causes both to be detached.

I looked into this a bit and turns out that it's a downstream issue. We are generating the same detach topic name for both UAVs (one overriding another). So something to be fixed on our end.

scpeters:
> > After testing it more downstream, I faced some interesting behaviors when detaching the models one after another. Consistently, detaching one of the joints does not seem to do anything, while detaching the other one causes both to be detached.
> 
> I looked into this a bit and turns out that it's a downstream issue. We are generating the same detach topic name for both UAVs (one overriding another). So something to be fixed on our end.

ok, I've also done some testing and if I only detach one joint, I would expect both boxes to ball, but only one is doing so. There may still be issues in the Detaching functionality here

iche033:
> ok, I've also done some testing and if I only detach one joint, I would expect both boxes to fall, but only one is doing so. There may still be issues in the Detaching functionality here

@aaronchongth was also able to reproduce the issue. Once the joint between a drone and an object is detached, we noticed that the drone is able to continue flying but the object remains static. Let us know if there's anything we can help here.

scpeters:
> > ok, I've also done some testing and if I only detach one joint, I would expect both boxes to fall, but only one is doing so. There may still be issues in the Detaching functionality here
> 
> @aaronchongth was also able to reproduce the issue. Once the joint between a drone and an object is detached, we noticed that the drone is able to continue flying but the object remains static. Let us know if there's anything we can help here.

I've tried debugging it, but I may need some advice from @mxgrey and @azeey about how to figure out what is going on

arjo129:
Just a random thought: I don't see us removing the WeldConstraint in the detach function. Does this have something to do with it.

mxgrey:
I haven't looked at the code yet, but based on the conversation I might suspect that if the `WeldConstraint` is being kept alive inside the `World` where one `BodyNode` was removed from the constraint while the other was kept inside the constraint, then DART will believe that the `BodyNode` that remains inside the constraint is meant to be rigidly attached to "the world" because a whenever a `BodyNode*` has a `nullptr` value it's treated as "the world".

I'll take a look to check my hypothesis later today.

mxgrey:
Like @arjo129 suggested, I don't see [`ConstraintSolver::removeConstraint`](https://github.com/dartsim/dart/blob/e85afb9effb144d77a08bcb9f4502e71319c08af/dart/constraint/ConstraintSolver.hpp#L104) ever get called to remove the weld constraint [that gets added here](https://github.com/gazebosim/gz-physics/pull/352/files#diff-fe114bf42e6ddbb1a9104656d43082539e6e979d376c3e34001b7152e8856d54R394). I believe this is almost certainly the cause of the problem.

What I would suggest is to maintain a hash map of `BodyNode* -> ConstraintBasePtr` where each `BodyNode*` in the map is a shard that was created for the sake of this WeldConstraint technique. When `DetachJoint` is called, we should check that hash map to see if the child `BodyNode` of the joint is one of the keys. If it is, then we should delete the shard, redistribute the inertia across all the remaining shards for that link, and then finally remove the `ConstraintBasePtr` weld constraint from the world.

All that being said, it's worth noting that this link splitting technique is not actually needed for this PR. The ordinary way of using the weld constraint would completely suffice for `AttachFixedJoint`. The motive for splitting the link back in Gazebo classic was to allow other joint types (e.g. Revolute, Prismatic), to form non-tree structures. If a link ever gets assigned more than one parent where one of the parents is a non-fixed joint, then we split the link into BodyNode shards, glue the shards together with weld constraints, and give each shard a different parent joint. Doing that allows each parent joint to be _any_ joint type. But if we are only implementing this for fixed joint then there's no benefit to the splitting.

I might suggest reworking this PR so that the link splitting logic gets applied to all the joint types instead of just the fixed joint. Otherwise it would be much cleaner to just remove the link splitting entirely.

scpeters:
I've merged #367 into this branch, so I think this is ready for re-review

iche033:
tested and confirmed that this PR works for us. I'll leave it to @azeey to give the final blessing.

dartsim/src/JointFeatures_TEST.cc
```diff
@@ -883,6 +884,147 @@ TEST_F(JointFeaturesFixture, JointAttachDetach)
...
+  EXPECT_EQ(initialModel2Pose,
+            math::eigen3::convert(dartBody2->getWorldTransform()));
+  EXPECT_EQ(initialModel3Pose,
+            math::eigen3::convert(dartBody3->getWorldTransform()));
+
+  // Create the second joint. This should be a WeldJoint constraint
+  const auto poseParent2 = dartBody3->getTransform();
+  const auto poseChild2 = dartBody2->getTransform();
+  auto poseParentChild2 = poseParent2.inverse() * poseChild2;
+  auto fixedJoint2 = model2Body->AttachFixedJoint(model3Body);
```
azeey, line 977: It would be great to add test expectations on the resulting masses after a WeldJointConstraint is created and destroyed, but feel free to punt for now.

azeey (approved):

dartsim/src/JointFeatures_TEST.cc
scpeters: I'll do this in a follow-up

PR #354, opened 2022-06-06
Added Common test

ahcorde:

# 🎉 New feature

## Summary

This draft PR try to set up a common arquitecture to test all physics engine under the same tests avoiding duplication and improving coverage.

The general idea is similar to `gz-rendering` with `GetParam()`, we will get the physics engine that we are testing and then we need to define the set of features associated to a specific physics engine (at compile time). Right now, Bullet supports a small set of features and TPE doesn't support joints. For this reason I used `auto` in the function signature of `commonTest` and `specificTest` to allow the compiler to compile test with less features that others.

I don't know if you have a better sugestion or improvements @mjcarroll @scpeters 

## Checklist
- [ ] Added tests

### DIFF
--- a/bullet/src/EntityManagementFeatures.cc
+++ b/bullet/src/EntityManagementFeatures.cc
@@ -168,6 +168,104 @@ bool EntityManagementFeatures::RemoveModelByName(
   return false;
 }
 
+const std::string &EntityManagementFeatures::GetEngineName(const Identity &) const
+{
+  static const std::string engineName = "bullet";
+  return engineName;
+}
+
+std::size_t EntityManagementFeatures::GetEngineIndex(const Identity &) const
+{
+  return 0;
+}
+
+std::size_t EntityManagementFeatures::GetWorldCount(const Identity &) const
+{
+  return worlds.size();
+}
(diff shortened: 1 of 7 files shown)


mjcarroll:
So based on some of my conversation with @scpeters last week, we should diverge from how ignition rendering is set up a little bit.  Since this is partially intended to provide a test suite for new library implementors, we should avoid hard-coding our current implementations via gtest parameters.

As an alternative, I think we should pass in a path to a library (either via environment variable or command line flags) and then evaluate which set of features that the library has via a list of `HasFeature` checks.  We can then choose to dispatch to tests based on which features that the library implements.  This will be a little bit more dynamic and allow features to be added without updating the test suite.

scpeters:
> As an alternative, I think we should pass in a path to a library (either via environment variable or command line flags) and then evaluate which set of features that the library has via a list of `HasFeature` checks. We can then choose to dispatch to tests based on which features that the library implements. This will be a little bit more dynamic and allow features to be added without updating the test suite.

I believe we discussed running each test with every physics plugin and using a gtest SKIP command if the physics features needed by that test aren't available in the plugin.

PR #365, opened 2022-06-23
Improved bullet EntityManagementFeatures GetEntities

ahcorde:
# 🎉 New feature

## Summary

Improved bullet EntityManagementFeatures GetEntities

## Checklist
- [ ] Added tests

### DIFF
--- a/bullet/src/Base.hh
+++ b/bullet/src/Base.hh
@@ -248,9 +248,9 @@ class Base : public Implements3d<FeatureList<Feature>>
   public: using CollisionInfoPtr = std::shared_ptr<CollisionInfo>;
   public: using JointInfoPtr  = std::shared_ptr<JointInfo>;
 
-  public: std::unordered_map<std::size_t, WorldInfoPtr> worlds;
-  public: std::unordered_map<std::size_t, ModelInfoPtr> models;
-  public: std::unordered_map<std::size_t, LinkInfoPtr> links;
+  public: std::map<std::size_t, WorldInfoPtr> worlds;
+  public: std::map<std::size_t, ModelInfoPtr> models;
+  public: std::map<std::size_t, LinkInfoPtr> links;
   public: std::unordered_map<std::size_t, CollisionInfoPtr> collisions;
   public: std::unordered_map<std::size_t, JointInfoPtr> joints;
 
--- a/bullet/src/EntityManagementFeatures.cc
+++ b/bullet/src/EntityManagementFeatures.cc
@@ -53,6 +53,40 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
     {_name, collisionConfiguration, dispatcher, broadphase, solver, world});
 }
 
+Identity EntityManagementFeatures::ConstructEmptyModel(
+  const Identity &_worldID, const std::string &_name)
+{
+  return this->AddModel(_worldID, {_name, _worldID, false, math::Pose3d(), {}});
+}
+
+Identity EntityManagementFeatures::ConstructEmptyLink(
+  const Identity &_modelID, const std::string &_name)
+{
+  double mass = 1;
+  btVector3 linkInertiaDiag(1, 1, 1);
+  // Fixed links have 0 mass and inertia
+  if (this->models.at(_modelID)->fixed)
+  {
+    mass = 0;
+    linkInertiaDiag = btVector3(0, 0, 0);
+  }
+
+  auto myMotionState = std::make_shared<btDefaultMotionState>(btTransform());
+  auto collisionShape = std::make_shared<btCompoundShape>();
+  btRigidBody::btRigidBodyConstructionInfo
+    rbInfo(mass, myMotionState.get(), collisionShape.get(), linkInertiaDiag);
+
+  auto body = std::make_shared<btRigidBody>(rbInfo);
+  body.get()->setActivationState(DISABLE_DEACTIVATION);
+
+  return this->AddLink(
+    _modelID,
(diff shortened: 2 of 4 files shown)


bullet/src/EntityManagementFeatures.cc
```diff
@@ -268,19 +396,55 @@ Identity EntityManagementFeatures::GetNestedModel(
...
+  return this->links.size();
 }
 
 Identity EntityManagementFeatures::GetLink(
-    const Identity &, std::size_t) const
+    const Identity &_modelID, std::size_t _linkIndex) const
 {
+  if (this->models.find(_modelID.id) == this->models.end())
+  {
+    return GenerateIdentity(0);
```
mxgrey, line 407: This should be returning `this->GenerateInvalidId()`, not an identity of 0. The sentinel value for an invalid entity is [max](https://github.com/gazebosim/gz-physics/blob/e91c58a49522cfdc5793fe8b980566eb26531ba4/include/ignition/physics/Entity.hh#L38-L39), not zero. All the other cases of `this->GenerateIdentity(0)` should also be fixed, except for returning the "engine" entity which is like a pseudo-entity.

mxgrey (changes requested):

bullet/src/EntityManagementFeatures.cc
```diff
@@ -216,36 +281,99 @@ Identity EntityManagementFeatures::GetEngineOfWorld(const Identity &) const
...
 {
-  static const std::string modelName = "bulletModel";
+  // Check if the model exists
+  if (this->models.find(_modelID.id) == this->models.end())
+  {
+    static const std::string modelName = "";
+    return modelName;
+  }
+
+  static const std::string modelName = this->models.at(_modelID)->name;
```
mjcarroll, line 339: I know it was this way before, but why the use of `static` here?

bullet/src/EntityManagementFeatures.cc
```diff
@@ -216,36 +281,99 @@ Identity EntityManagementFeatures::GetEngineOfWorld(const Identity &) const
...
+
   return 0;
 }
 
-Identity EntityManagementFeatures::GetWorldOfModel(const Identity &) const
+Identity EntityManagementFeatures::GetWorldOfModel(
+  const Identity &_modelId) const
 {
-  return this->GenerateIdentity(0);
+  if (this->models.find(_modelId.id) == this->models.end())
```
mjcarroll, line 369: Could also use a bit shorter, since you aren't doing anything with the result:

```suggestion
  if (!this->models.count(_modelId.id))
```

bullet/src/Base.hh
```diff
@@ -248,9 +248,9 @@ class Base : public Implements3d<FeatureList<Feature>>
   public: using CollisionInfoPtr = std::shared_ptr<CollisionInfo>;
   public: using JointInfoPtr  = std::shared_ptr<JointInfo>;
 
-  public: std::unordered_map<std::size_t, WorldInfoPtr> worlds;
-  public: std::unordered_map<std::size_t, ModelInfoPtr> models;
-  public: std::unordered_map<std::size_t, LinkInfoPtr> links;
+  public: std::map<std::size_t, WorldInfoPtr> worlds;
+  public: std::map<std::size_t, ModelInfoPtr> models;
+  public: std::map<std::size_t, LinkInfoPtr> links;
```
mjcarroll, line 253: The reason for this changed didn't jump out at me.  Is there a reason to use `map` vs `unordered_map`?

ahcorde: We use the index of the map, I believe `std::unordered_map` is not keeping the order of the data structrue

bullet/src/EntityManagementFeatures.cc
mxgrey: I think this must be a copy/paste error from line 335. It kind of makes sense on line 335 but seems definitely wrong to use here since it means every single valid model will always report having the same name, and that name will be whichever was the first one to be asked for it.

bullet/src/Base.hh
ahcorde: we should return the index here:

https://github.com/gazebosim/gz-physics/blob/f15cb9c51750d68a029b80f1ef7169ae23c7812f/bullet/src/EntityManagementFeatures.cc#L343-L364

we use the index here:

https://github.com/gazebosim/gz-physics/blob/f15cb9c51750d68a029b80f1ef7169ae23c7812f/bullet/src/EntityManagementFeatures.cc#L287-L307

bullet/src/EntityManagementFeatures.cc
```diff
@@ -53,6 +53,40 @@ Identity EntityManagementFeatures::ConstructEmptyWorld(
...
+  body.get()->setActivationState(DISABLE_DEACTIVATION);
+
+  return this->AddLink(
+    _modelID,
+    {_name, _modelID, math::Pose3d(), math::Pose3d(),
+     mass, linkInertiaDiag,
+     myMotionState, collisionShape, body});
+}
+
+
```
mjcarroll, line 89: nit: extra whitespace

bullet/src/EntityManagementFeatures.cc
mjcarroll: I think this should work: https://github.com/gazebosim/gz-physics/pull/385

Basically, if it's in the map, return the reference, otherwise return a static `std::string` that will have a stable reference.  We should be able to follow this pattern in the other places where we do this.

Returning a string reference is not a great API for this reason, but I suppose it's a bit late for that.

ahcorde:
this PR is targeting `bullet` not `bullet-featherstone`, I will close it for now

PR #366, opened 2022-06-23
Add WeldJointConstraint for fixed joints if the parent is not a FreeJoint

iche033:
# 🎉 New feature

## Summary

Revives the PR from @arjo129: https://github.com/gazebosim/gz-physics/pull/345. 

I'm pasting the description from the original PR below:

## Summary

Currently, if we try to attach two models to the same item using fixed joints, we are unable to do so. This is because Dart's WeldJoint requires that the rigid body skeletons to form a tree. We can solve this by using WeldJointConstraints instead and that is what I'm trying to use.

## Test it
I've introduced a unit test [here](https://github.com/ignitionrobotics/ign-physics/blob/54c29f1d0d871cd4ddd04f144344475a9314a227/dartsim/src/JointFeatures_TEST.cc#L896).
2022-04-28T16:51:49 107398391
Essentially what happens is there are two floating boxes and a box in the middle that's resting. We start the system out by creating the two fixed joints between the box resting on the big box and the floating box. The middle box will now have two parents. However there should be no movement as the middle box will be holding the other two boxes that are floating in mid air. We run this for 100 steps to make sure that there is no movement. This is because the middle box is holding on to the two side boxes. Then we release the joints the two boxes should fall away.

## TODO
Since this affects the physics in a lot of places it will be important to do the following before merging
- [x] Check performance - Seems to be able to do realtime speeds in the example shown below
- [x] Figure out how to cast joints to weld contraints - Skip casting if is a constraint
- [x] Do an upstream integration test.
- [x] Get a :heavy_check_mark: in the CI. - Good ol' :window: not passing
- [ ] Check other affected functions.
- [ ] Find out how to deal with `ObjectToId` as weld constraints are not added to this list. - Pending reviewer feedback.

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/Base.hh
+++ b/dartsim/src/Base.hh
@@ -231,7 +231,7 @@ class Base : public Implements3d<FeatureList<Feature>>
   public: using DartBodyNodePtr = dart::dynamics::BodyNodePtr;
   public: using DartJoint = dart::dynamics::Joint;
   public: using DartJointPtr = dart::dynamics::JointPtr;
-  public: using DartWeldJointConsPtr =dart::constraint::WeldJointConstraintPtr;
+  public: using DartWeldJointConsPtr = dart::constraint::WeldJointConstraintPtr;
   public: using DartShapeNode = dart::dynamics::ShapeNode;
   public: using DartShapeNodePtr = std::shared_ptr<DartShapeNode>;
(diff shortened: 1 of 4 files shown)


iche033:
Closing this as I can just reopen the original PR

PR #367, opened 2022-06-24
Suggestions to #352

azeey:
# 🦟 Bug fix

## Summary
* Remove welded dummy bodies and the associated `WeldJointConstraint` when detaching a joint.
* Fixes integer divisions.

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/Base.hh
+++ b/dartsim/src/Base.hh
@@ -370,11 +370,15 @@ class Base : public Implements3d<FeatureList<Feature>>
   private: static math::Inertiald DivideInertial(
                const math::Inertiald &_wholeInertial, std::size_t _count)
   {
+    if (_count == 1)
+    {
+      return _wholeInertial;
+    }
     math::Inertiald dividedInertial;
     math::MassMatrix3d dividedMassMatrix;
     dividedMassMatrix.SetMass(_wholeInertial.MassMatrix().Mass() /
                               static_cast<double>(_count));
-    dividedMassMatrix.SetMoi(_wholeInertial.Moi() *
+    dividedMassMatrix.SetMoi(_wholeInertial.MassMatrix().Moi() *
                              (1. / static_cast<double>(_count)));
     dividedInertial.SetMassMatrix(dividedMassMatrix);
     dividedInertial.SetPose(_wholeInertial.Pose());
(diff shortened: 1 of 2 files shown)


iche033 (approved):
I tested repeatedly attaching and detaching the joints and this works well for us. I printed out the inertial values in `AssignInertialToBody` and they look consistent. I also went through our tutorials for grasping objects of different size and mass and the behavior looks good.

dartsim/src/Base.hh
```diff
@@ -367,6 +367,31 @@ class Base : public Implements3d<FeatureList<Feature>>
...
   }
 
+  private: static math::Inertiald DivideInertial(
+               const math::Inertiald &_wholeInertial, std::size_t _count)
+  {
+    math::Inertiald dividedInertial;
+    math::MassMatrix3d dividedMassMatrix;
+    dividedMassMatrix.SetMass(_wholeInertial.MassMatrix().Mass() /
+                              static_cast<double>(_count));
+    dividedMassMatrix.SetMoi(_wholeInertial.Moi() *
```
scpeters, line 377: you have to be careful here with the inertial rotations. Since you are calling `Inertial::SetPose` below, you should not set and get the `Moi` as a `Matrix3`, since that has the inertial pose rotations embedded within. I would recommend using `MassMatrix3` APIs like `SetDiagonalMoments` and `SetOffDiagonalMoments` or `SetInertiaMatrix`

dartsim/src/Base.hh
```diff
@@ -367,6 +367,31 @@ class Base : public Implements3d<FeatureList<Feature>>
     return id;
   }
 
+  private: static math::Inertiald DivideInertial(
+               const math::Inertiald &_wholeInertial, std::size_t _count)
+  {
+    math::Inertiald dividedInertial;
```
scpeters, line 373: optional:

if `_count == 1`, then just `return _wholeInertial;`

dartsim/src/Base.hh
```diff
@@ -398,29 +423,68 @@ class Base : public Implements3d<FeatureList<Feature>>
...
+      // happen.
+      ignerr << "Could not find welded body node for link " << _link->name
+             << ". Merging of link and welded body failed.";
+      return;
+    }
+
+    if (_link->inertial)
+    {
+      std::size_t nodeCount = _link->weldedNodes.size();
+      if (nodeCount > 0)
```
scpeters, line 471: `nodeCount` is `1 + _link->weldedNodes.size()` in `SplitAndWeldLink` and I think it should be the same here

then this if statement wouldn't be necessary

dartsim/src/Base.hh
azeey: I've replaced it with `MassMatrix3d::Moi` (not Inertiald::Moi) which does not take poses into account in https://github.com/gazebosim/gz-physics/pull/367/commits/e772f869fa11ee06456dc7e5093b10e1827c2f7e. I could go with the APIs you recommended if this is still confusing/error prone. It's just less lines of code this way.

dartsim/src/Base.hh
azeey: Thanks for catching that. https://github.com/gazebosim/gz-physics/pull/367/commits/e772f869fa11ee06456dc7e5093b10e1827c2f7e

aaronchongth (approved):
Just tested through dragging and lifting examples with repeated attach and detach, all is working well and the behavior looks great!

scpeters (approved):

PR #373, opened 2022-06-30
Add bullet-featherstone plugin

mxgrey:
# 🎉 New feature

## Summary

This PR adds a `gz-physics-bullet-featherstone-plugin` which wraps the Featherstone API of the bullet physics engine. This is created as a separate plugin from `gz-physics-bullet-plugin` because the API of Bullet's Featherstone dynamics is different enough that there isn't really feature parity or easy interchangeability between them. Users will need to choose whether they are willing to settle for fewer features in exchange for the better performance of Bullet's Featherstone implementation.

This PR depends on https://github.com/gazebosim/gz-sim/pull/1560

## Test it

I have been testing the plugin using the [NAO demo world](https://app.gazebosim.org/OpenRobotics/fuel/worlds/NAO%20joint%20control). Just download, unzip, and run

```
gz sim -v 4 --physics-engine gz-physics-bullet-featherstone-plugin nao_world.sdf
```

However, I have been noticing an inexplicable bug... The `bullet-featherstone` plugin works perfectly fine when `gz-physics` is compiled in `Debug` mode, but fails to respond to changing joint force commands when `gz-physics` is compiled in `Release` or `RelWithDebInfo` mode. I don't have an explanation for this at the moment. I've tried compiling bullet from source in both single and double precision mode, but the effect is the same either way.

The strangest part is that the initial joint force commands seem to be respected because the arms manage to remain pointing straight outward. Also if you adjust the position slider _before_ starting the simulation, then the joint will manage to reach that position, but it stops responding to any changes after the simulation begins.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #374, opened 2022-07-07
Added suggestions to #373

ahcorde:

# 🎉 New feature

## Summary

Added more features to bullet-featherstone

## Checklist
- [x] Added tests

### DIFF
--- a/bullet-featherstone/src/Base.hh
+++ b/bullet-featherstone/src/Base.hh
@@ -266,7 +266,8 @@ class Base : public Implements3d<FeatureList<Feature>>
     {
       // We are adding the root link. This means the model should not already
       // have a root link
-      assert(model->linkEntityIds.empty());
+      // This check makes `ConstructEmptyLink` to fail
+      // assert(model->linkEntityIds.empty());
     }
     model->linkEntityIds.push_back(id);
 
--- a/bullet-featherstone/src/EntityManagementFeatures.cc
+++ b/bullet-featherstone/src/EntityManagementFeatures.cc
@@ -265,6 +265,162 @@ bool EntityManagementFeatures::RemoveModelByName(
     this->GenerateIdentity(it->second, this->models.at(it->second)));
 }
 
+
+/////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


mxgrey:
I would discourage adding many of these features to the bullet-featherstone plugin. Things like constructing empty models and empty links is not helpful or meaningful for the Featherstone API. In the Featherstone a model *needs* to have a base link, and a link *needs* to belong to a model.

I assume that these features were probably added so that the bullet-featherstone plugin can be tested against more of the "common" physics tests, but I think the common testing framework should just have each test skip over any plugins which don't provide the features that the test requires.

ahcorde:
> I would discourage adding many of these features to the bullet-featherstone plugin. Things like constructing empty models and empty links is not helpful or meaningful for the Featherstone API. In the Featherstone a model _needs_ to have a base link, and a link _needs_ to belong to a model.
> 
> I assume that these features were probably added so that the bullet-featherstone plugin can be tested against more of the "common" physics tests, but I think the common testing framework should just have each test skip over any plugins which don't provide the features that the test requires.

Do you know which features make sense to include to the `bullet-featherstone` plugin ?

mxgrey:
> Do you know which features make sense to include to the bullet-featherstone plugin ?

We could consider adding frame semantics for shapes and joints if that's getting used anywhere in `gz-sim`.

We might also be able to support adding and removing collision shapes, but it's not totally clear to me if that will work nicely with Bullet's Featherstone API. It's also probably not a very useful feature when we don't support attaching/detaching links.

We should be able to add a feature for creating fixed joints between the links of models. That might also allow bullet-featherstone to support closed kinematic chains, similar to what's been done recently with the dartsim plugin.

mxgrey (approved):
I've gone through and removed a couple features that I don't think are appropriate for `bullet-featherstone` and also fixed some implementation issues that I noticed. I hope you don't mind that I committed directly to the branch for expediency.

I think it's fine to merge this into the upstream `add-bullet-featherstone` branch if no one else has any concerns about it.

PR #382, opened 2022-07-12
Added Shapes Features to bullet-featherstone

ahcorde:
# 🎉 New feature

## Summary

Added Shapes Features to bullet-featherstone.

TODOs
 - Added test
 - fix boundingbox

## Checklist
- [ ] Added tests


bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -61,14 +61,30 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
     // needs to change.
-    model = this->FrameInterface<ModelInfo>(_id);
+    // model = this->FrameInterface<ModelInfo>(_id);
+    // I commented this line because it's returning a bad object.
+    // I'm not able to access getBaseWorldTransform
+    for (auto const [key, value] : this->links)
+    {
+      for (unsigned int i = 0; i < value->collisionEntityIds.size(); ++i)
+      {
+        if (value->collisionEntityIds[i] == _id.ID())
```
mxgrey, line 71: Are you seeing a case where the incoming `_id` is for a collision entity? The [kinematics feature list](https://github.com/gazebosim/gz-physics/blob/166ff209ff4e74c2ab8e5daed1214b2e4aa6e786/bullet-featherstone/src/KinematicsFeatures.hh#L31-L33) doesn't include [`ShapeFrameSemantics`](https://github.com/gazebosim/gz-physics/blob/166ff209ff4e74c2ab8e5daed1214b2e4aa6e786/include/gz/physics/FrameSemantics.hh#L189) so it's not supposed to be possible for a user to ask for the frame data of a collision object.

If you're seeing that happen then it probably indicates an error in the template system. If you can point me at a way to recreate it, that would be very helpful.

bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -232,7 +232,7 @@ std::optional<Structure> ValidateModel(const ::sdf::Model &_sdfModel)
       for (std::size_t i = 0; i < model.LinkCount(); ++i)
       {
         const auto *link = model.LinkByIndex(i);
-        if (parentOf.count(link) == 0)
+        if (parentOf.size() != 0 && parentOf.count(link) == 0)
```
mxgrey, line 235: I'm not sure why we would check the size here. `parentOf.size() == 0` implies that there is only one link and that link has no parent. In that case `parentOf.count(link) == 0` will be true. Inside of this branch we then double-check that if a `rootLink` was selected while iterating over the model then that root link is the same as the one we're currently looking at. I think it's reasonable to double-check that because if that's not true then there's an error in the logic of the parser.

Was there a specific situation that motivated these additional conditions? If so it would be helpful if I can reproduce it to figure out what might be wrong with the parser's logic.

PR #383, opened 2022-07-14
ign -> gz Provisional Finale: Source Migration : gz-physics

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/767

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,7 +9,7 @@
 
 ### Gazebo Physics 5.1.0 (2021-11-12)
 
-1. Remove unused ign_auto_headers.hh.in
+1. Remove unused gz_auto_headers.hh.in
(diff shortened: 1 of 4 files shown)


ahcorde (approved):

Changelog.md
```diff
@@ -69,7 +69,7 @@
 
 ### Gazebo Physics 4.3.0 (2021-11-11)
 
-1. Remove unused ign_auto_headers.hh.in
+1. Remove unused gz_auto_headers.hh.in
```
ahcorde, line 72: no really sure if you should modify the Changelog

methylDragon: This came up before, I think we were ok with modifying it to whittle down the number of cases, so I'll just accept this one too :o

PR #384, opened 2022-07-14
Support fluid added mass

chapulina:
# 🎉 New feature

* Part of https://github.com/gazebosim/gz-sim/issues/1462
* Requires 
    * https://github.com/gazebosim/gz-math/pull/459
    * https://github.com/gazebosim/sdformat/pull/1077

## Summary

Uses the APIs added as part of the [Fluid Added Mass Proposal](http://sdformat.org/tutorials?tut=added_mass_proposal&branch=chapulina/added_mass).

Ideally the added mass functionality would be part of an optional feature. I have to look a bit deeper to figure out how to take that out of the link construction.

## Test it

I'm working on tests and a `gz-sim` PR

## Checklist
- [ ] Added tests
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers


dartsim/src/SDFFeatures.cc
```diff
@@ -608,6 +608,31 @@ Identity SDFFeatures::ConstructSdfLink(
 
   bodyProperties.mInertia.setLocalCOM(localCom);
 
+  // If added mass is provided, add that to DART's computed spatial tensor
+  // TODO(chapulina) Put in another feature
+  if (sdfInertia.FluidAddedMass().has_value())
+  {
+    // TODO check off-diagonal blocks
```
scpeters, line 615: I would comment here about how the order of variables differs between SDFormat / Gazebo and Dartsim / Featherstone, which is why the submatrices need to be shuffled.

scpeters (commented):
> Ideally the added mass functionality would be part of an optional feature. I have to look a bit deeper to figure out how to take that out of the link construction.

this is beyond the scope of this PR; we don't currently have granularity in our SDF construction features. I'll see if we have an issue about this yet

chapulina:
We're past feature freeze ❄️ . The plan is to not merge this PR into `gz-physics6` until after the stable release (i.e. October).

srmainwaring:
@scpeters I've been experimenting with this and think there may be a problem with the way the fluid added mass influences  the gravity force in dart and this affects hydrostatic calculations. 

If a body is placed at its equilibrium position (say a sphere with mass = 1/2 volume x water density with centre of buoyancy = centre of mass = centre of geometry placed with it's origin at the free surface), then the net force (hydrostatic + gravity) should be zero and the body should not accelerate.

What happens however is that the hydrostatic force is calculated in a plugin with the body mass, whereas the gravity force appears to be calculated in dart using the spatial tensor (which has been modified with the fluid added mass, so the total adjusted mass is > body mass). The forces are no longer balanced and the object accelerates downwards. 

I'm still going through the details to understand how dart implements this, but the problem seems to be that gravity is added globally for the world in the physics engine, whereas the remainder of the forces are added externally via plugins.

One solution may be to have an option to switch gravity off in the core physics engine and add it back in via a system plugin that adds the gravity force for each link using the mass matrix mass?

dartsim/src/SDFFeatures.cc
```diff
@@ -600,6 +600,31 @@ Identity SDFFeatures::ConstructSdfLink(
...
+    math::Matrix6d featherstoneMatrix;
+    featherstoneMatrix.SetSubmatrix(math::Matrix6d::TOP_LEFT,
+        sdfInertia.FluidAddedMass().value().Submatrix(
+        math::Matrix6d::BOTTOM_RIGHT));
```
srmainwaring, line 611: This appears to be missing the term `mMass*C*C.transpose()` that appears in the Dart spatial tensor:

https://github.com/dartsim/dart/blob/26267f384cd0e5f4c00a6b074005902313df855b/dart/dynamics/Inertia.cpp#L393-L398

```c++
void Inertia::computeSpatialTensor()
{
  Eigen::Matrix3d C = math::makeSkewSymmetric(mCenterOfMass);

// Top left
mSpatialTensor.block<3,3>(0,0) = getMoment() + mMass*C*C.transpose();
```

JoanAguilar:
Hi @srmainwaring, thanks for your comment. We suspected that the current added mass implementation might have an effect on how DART computes the gravity force, your tests seem to confirm that. We concur that disabling gravity in DART and applying it via a plugin might be a good solution to the problem.

srmainwaring:
Here's a suggestion for calculating the missing term in the spatial tensor when the mass block no longer comprises a 3x3 diagonal matrix with identical elements:

```python
import numpy as np
import scipy as sp

fAddedMass = np.array([
52732.7109003133, -0.0461102712478722, 5.88390664387022, 0.296417351644741, 246367.266847758, -1.69953653780694,
-0.0678279438415122, 52944.8240664938, -1.47610970294622, -246974.480351485, -0.376605346811632, 34.7833617031108,
5.56653991122738, 0.0361713254775427, 197297.142256690, 0.802697521049529, -87.4935912381782, -0.0504125898702509,
0.410840691063419, -247341.339091709, 5.02714890072835, 1214533.79620113, 2.83888902153701, -166.779120688123,
246734.857503928, -0.346433003966073, -87.1795788996776, 2.60143644204397, 1215195.98272167, -2.36242713115735,
-1.30857894347732, 34.6478012553392, 0.0173325214982212, -165.977540011457, -1.09781377470450, 108.258384753611
]).reshape(6, 6)

# decomposition of the fluid added mass matrix
M    = fAddedMass[0:3, 0:3]          # mass block
I    = fAddedMass[3:6, 3:6]          # moi block
CM   = fAddedMass[3:6, 0:3]          # cross-terms
A    = sp.linalg.sqrtm(M)            # srqt of mass block
invA = np.linalg.inv(A)              # inverse sqrt of mass block
CA   = np.matmul(CM, invA)           # srqt of mMass*C*C.transpose() term
mCCT = np.matmul(CA, CA.transpose()) # mMass*C*C.transpose()

# contributions to the spatial tensor
TL   = I + mCCT   
BL   = CM
TR   = CM.transpose()  
BR   = M

# assemble spatial tensor
S    = np.zeros([6, 6])
S[0:3, 0:3] = TL
S[3:6, 0:3] = BL
S[0:3, 3:6] = TR
S[3:6, 3:6] = BR

with np.printoptions(precision=0, suppress=True):
    print(f"fAddedMass:\n{fAddedMass}")
    print(f"I:\n{I}")
    print(f"mCCT:\n{mCCT}")
    print(f"S:\n{S}")

fAddedMass:
[[  52733.      -0.       6.       0.  246367.      -2.]
 [     -0.   52945.      -1. -246974.      -0.      35.]
 [      6.       0.  197297.       1.     -87.      -0.]
 [      0. -247341.       5. 1214534.       3.    -167.]
 [ 246735.      -0.     -87.       3. 1215196.      -2.]
 [     -1.      35.       0.    -166.      -1.     108.]]
I:
[[1214534.       3.    -167.]
 [      3. 1215196.      -2.]
 [   -166.      -1.     108.]]
mCCT:
[[1155500.       2.    -162.]
 [      2. 1154465.      -6.]
 [   -162.      -6.       0.]]
S:
[[2370034.       5.    -329.       0.  246735.      -1.]
 [      5. 2369661.      -8. -247341.      -0.      35.]
 [   -328.      -7.     108.       5.     -87.       0.]
 [      0. -247341.       5.   52733.      -0.       6.]
 [ 246735.      -0.     -87.      -0.   52945.      -1.]
 [     -1.      35.       0.       6.       0.  197297.]]
```
I've used the add mass tensor from the example above - if the calculation is correct then the additional term `mMass*C*C.transpose()` is the same order as `I`.

dartsim/src/SDFFeatures.cc
srmainwaring: I've created a patch in the commits below - the first includes debug statements which are cleaned up in the second.

https://github.com/srmainwaring/gz-physics/commit/9dba6cd30f14a017a13e01b6088c92071b5ea2ad
https://github.com/srmainwaring/gz-physics/commit/b60be398a990684da092bf7e3f4eec1f296058ce

```diff
+
+    // Calculate CoM term.
+    //
+    // Roy Featherstone, Rigid Body Dynamics Algorithms, Springer
+    // The spatial inertia matrix is defined in §2.13, p33, eq(2.63).
+    auto M = sdfInertia.FluidAddedMass().value().Submatrix(
+        math::Matrix6d::TOP_LEFT);
+
+    auto CM = sdfInertia.FluidAddedMass().value().Submatrix(
+        math::Matrix6d::BOTTOM_LEFT);
+
+    auto invM = M.Inverse();
+    auto C = CM * invM;
+    auto CMCT = CM * C.Transposed();
+
+    // I + C * M * C.transpose
     featherstoneMatrix.SetSubmatrix(math::Matrix6d::TOP_LEFT,
         sdfInertia.FluidAddedMass().value().Submatrix(
-        math::Matrix6d::BOTTOM_RIGHT));
+        math::Matrix6d::BOTTOM_RIGHT) + CMCT);
```

JoanAguilar:
Revisiting this PR, the major thing that seems to be missing for this to merge is unit tests.

Beyond that, there are a couple of side discussions we probably want to keep track of:

1. I think we cleared out with @srmainwaring (in a separate discussion) that the implementation in this PR is correct, and that the specification of the added mass matrix in an SDF file is always done in the link frame; it is probably for our own benefit to document this somewhere.

2. Per @srmainwaring [comment](https://github.com/gazebosim/gz-physics/pull/384#issuecomment-1258240733), the addition of added mass results in incorrect gravitational effects. @srmainwaring already proposed a solution in the previously mentioned comment:

> One solution may be to have an option to switch gravity off in the core physics engine and add it back in via a system plugin that adds the gravity force for each link using the mass matrix mass?

dartsim/src/SDFFeatures.cc
mjcarroll: Right now, we are only allowing added mass to be specified in the link frame.  These computations would be required for translating to the center of mass.  For now, this is unnecessary.

mjcarroll (approved):

PR #387, opened 2022-07-19
Fix bullet-featherstone collisions

mxgrey:
This PR fixes collisions for bullet-featherstone, and also updates the CMakeLists.txt to be compatible with the latest main branches of gz-cmake.

I've tested these changes with a modified `shapes.sdf`. The first modification is to raise up the shapes by 1.0m to make the collision behavior more clear. Doing that results in [most of the shapes just dropping through the plane](https://user-images.githubusercontent.com/1307217/179689598-bea56b84-aea6-4247-93a8-ef44ff8f6884.mp4).

It looks to me like Bullet's `btStaticPlaneShape` is only able to detect collisions with boxes. I think this would qualify as an unexpected behavior that could negatively impact users, so we may want to consider replacing instances of `PlaneShape` with a big box instead.

When I further modify `shapes.sdf` to use a big box for the ground plane, I get these two results:
1. [Everything works as expected](https://user-images.githubusercontent.com/1307217/179690085-ebdef889-768a-4959-97a1-2c00bce772a3.mp4)
2. [The ellipsoid drops through the ground](https://user-images.githubusercontent.com/1307217/179690150-caa3a5f6-4934-4586-a14a-e98a0027fa93.mp4)

I haven't found any pattern for when/why the ellipsoid drops. I run the simulation with the exact same SDF repeatedly and I'd estimate the ellipsoid drops through the ground ~80% of the time. That's an awfully large false negative, so I don't think we should just ignore it. We may want to declare that ellipsoid shapes aren't supported for bullet-featherstone if we can't solve the failure to detect the collision.


bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -639,9 +639,14 @@ bool SDFFeatures::AddSdfCollision(
     return false;
   }
 
-  double mu = 1.0;
-  double mu2 = 1.0;
-  double restitution = 0.0;
+//  double mu = 1.0;
```
ahcorde, line 642: remove?

ahcorde (changes requested):
just a minor comment, if we merge this PR we should open an issue with the ellipsoid issue.

bullet-featherstone/src/SDFFeatures.cc
mxgrey: Right, I was experimenting with the default parameters to see how they impact the collision behavior. I've returned them back to their default values here [862411b](https://github.com/gazebosim/gz-physics/pull/387/commits/862411b02426503ef149ecdd3ff168fe56360435)

mxgrey:
Thanks for the review, @ahcorde . I've addressed your inline comment and opened an issue for this: https://github.com/gazebosim/gz-physics/issues/389

ahcorde (approved):

PR #395, opened 2022-07-28
ABI tweaks to help bullet-featherstone

mxgrey:
These small tweaks to the ABI are needed for #373, and if we can merge these before code freeze then we can introduce `bullet-featherstone` post-release without and harmful effects.

### DIFF
--- a/include/gz/physics/FrameSemantics.hh
+++ b/include/gz/physics/FrameSemantics.hh
@@ -162,7 +162,7 @@ namespace gz
         /// \tparam T The stored pointer is cast to this type.
         /// \param[in] _frameID the frame identifier to cast an interface from
         /// \return A raw pointer from the stored shared_ptr but cast to the
-        ///         /// provided type T
+        ///         provided type T
         protected: template<typename T>
         T *FrameInterface(const FrameID &_frameID) const;
       };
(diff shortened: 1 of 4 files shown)


include/gz/physics/FrameSemantics.hh
```diff
@@ -156,6 +156,15 @@ namespace gz
         /// \return A FrameID object that corresponds to _identity.
         protected: virtual FrameID GenerateFrameID(
             const Identity &_identity) const;
+
+        /// \brief An implementation class can use this function to get the
+        /// interface that _frameID holds onto in its reference-counted pointer.
+        /// \tparam T The stored pointer is cast to this type.
+        /// \param[in] _frameID the frame identifier to cast an interface from
+        /// \return A raw pointer from the stored shared_ptr but cast to the
+        ///         /// provided type T
```
scpeters, line 165: is the extra `///` intentional?

scpeters (approved):

mjcarroll (approved):

include/gz/physics/FrameSemantics.hh
mxgrey: Nope, that was just my IDE trying to be helpful. Fixed by [0550d56](https://github.com/gazebosim/gz-physics/pull/395/commits/0550d566e4be8b598a0d91330c106f523cbaaa19)

PR #441, opened 2022-10-17
🎈 6.1.0

mjcarroll:
# 🎈 Release

Preparation for 6.1.0 release.

Comparison to 6.0.0: https://github.com/gazebosim/gz-physics/compare/gz-physics6_6.0.0...gz-physics6

Needed by https://github.com/gazebosim/gz-sim/pull/1560

## Checklist
- [ ] Asked team if this is a good time for a release
- [ ] There are no changes to be ported from the previous major version
- [ ] No PRs targeted at this major version are close to getting in
- [ ] Bumped minor for new features, patch for bug fixes
- [ ] Updated changelog
- [ ] Link to PR updating dependency versions in appropriate repository in [gazebo-release](https://github.com/gazebo-release) (as needed): <LINK>


scpeters:
I've reviewing why GetCollision was added to sdf/ConstructCollision.hh

scpeters:
> I've reviewing why GetCollision was added to sdf/ConstructCollision.hh

https://github.com/gazebosim/gz-physics/pull/373/files#r997339782

> why was this API added instead of using GetShapeFromLink::GetShape?

mjcarroll:
> why was this API added instead of using GetShapeFromLink::GetShape?

Since we were using the DART implementation as a reference, and DART doesn't have that implemented, it likely was an oversight.  I believe that there were some additional physics API template gymnastics that weren't possible to resolve, but @ahcorde has a better memory for that.

ahcorde:
> > why was this API added instead of using GetShapeFromLink::GetShape?
> 
> Since we were using the DART implementation as a reference, and DART doesn't have that implemented, it likely was an oversight. I believe that there were some additional physics API template gymnastics that weren't possible to resolve, but @ahcorde has a better memory for that.

Yes, I was not able to recast the pointer in `gz-sim`. I'm open to suggestions

mjcarroll:
Pending #444

scpeters (approved):

scpeters:
we're going to need to update the debian metadata; I would recommend making a few test debbuilds before doing the actual release

scpeters:
> we're going to need to update the debian metadata; I would recommend making a few test debbuilds before doing the actual release

actually, it's already been built by the nightly jobs without any complaints:

* https://build.osrfoundation.org/view/ign-garden/job/gz-physics6-debbuilder/663/

I believe it's installing both the `bullet` and `bullet-featherstone` plugins together in the `-bullet*` debs, which I think is fine

* https://github.com/gazebo-release/gz-physics6-release/blob/main/ubuntu/debian/libgz-physics-bullet-dev.install
* https://github.com/gazebo-release/gz-physics6-release/blob/main/ubuntu/debian/libgz-physics-bullet.install

so I think the release can proceed

PR #446, opened 2022-11-01
Suppress MSVC warning that is not applicable

mjcarroll:
This triggers a warning in downstream libraries such as:

https://build.osrfoundation.org/job/ign_gazebo-pr-win/5195/msbuild/

```
'gz::physics::detail::ComposePlugin::type': base-class 'gz::physics::detail::ComposePlugin::n,void>' is already a base-class of 'gz::physics::detail::ComposePlugin::type'
```
 
After talking with @mxgrey, while this is typically a relevant warning by MSVC, it is not applicable so we can suppress it.

### DIFF
--- a/include/gz/physics/FeatureList.hh
+++ b/include/gz/physics/FeatureList.hh
@@ -22,6 +22,14 @@
 
 #include <gz/physics/Feature.hh>
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+/// Suppress warnings about "base-class is already a base-class"
+/// Typically, this would indicate a diamond pattern in inheritance,
+/// but this is a false positive for the plugin mechanism.
+#pragma warning(disable: 4584)
+#endif  // defined(_MSC_VER)
+
 namespace gz
 {
   namespace physics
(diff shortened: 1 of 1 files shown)


include/gz/physics/FeatureList.hh
```diff
@@ -22,6 +22,14 @@
 
 #include <gz/physics/Feature.hh>
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+/// Suppress warnings about "base-class is already a base-class"
+/// Typically, this would indicate a diamond pattern in inheritance,
+/// but this is a false positive for the plugin mechanism.
```
mxgrey, line 29: If I were to be pedantic, the warning is correct that a base class is being repeated so it's not technically a false positive.

However, the repeated base classes in this case have no fields in them, so there's no risk of ambiguity and therefore won't cause any problems. The only reason the base classes get repeated is to coerce some recursive template magic into compiling; they're not doing anything beyond that.

mxgrey (approved):

include/gz/physics/FeatureList.hh
mjcarroll: Updated the comment to reflect this.

scpeters (approved):

nkoenig (approved):

azeey:
This fix can also be applied to ign-physics5 to fix warnings in ign-gazebo6 to partially resolve https://github.com/gazebosim/gz-sim/issues/2384

@Mergifyio backport ign-physics5

azeey:
@Mergifyio backport ign-physics5

PR #459, opened 2022-12-03
Report skipped test as skipped instead of passing

azeey:
# 🦟 Bug fix

## Summary
Currently, running `make test` reports "Passed" even though there might be some tests skipped. This fixes that by checking the output of gtest for the string `[  SKIPPED  ]`. The use of the `SKIP_REGULAR_EXPRESSION` property on tests is only available from cmake 3.16, so I've updated the minimum required version. This version is available in all platforms we support for gz-physics6.

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.16.0 FATAL_ERROR)
 
 #============================================================================
 # Initialize the project
--- a/test/common_test/CMakeLists.txt
+++ b/test/common_test/CMakeLists.txt
@@ -20,6 +20,9 @@ function(configure_common_test PHYSICS_ENGINE_NAME test_name)
       ${test_name}
       $<TARGET_FILE:${PROJECT_LIBRARY_TARGET_NAME}-${PHYSICS_ENGINE_NAME}-plugin>
   )
+  set_property(TEST ${test_name}_${PHYSICS_ENGINE_NAME} PROPERTY
+    SKIP_REGULAR_EXPRESSION "\[[ \t]*SKIPPED[ \t]*]"
+  )
 endfunction()
 
 set(GZ_PHYSICS_RESOURCE_DIR "${CMAKE_SOURCE_DIR}/resources")


azeey:
Now `ctest` correctly reports the skipped tests, but jenkins doesn't seem to notice 😞

azeey:
> Now `ctest` correctly reports the skipped tests, but jenkins doesn't seem to notice disappointed

That's because the common tests were not creating junit files. This is fixed now.

ahcorde (approved):

azeey:
I'd like to get @j-rivero's  ✔️  before merging.

scpeters:
> Now `ctest` correctly reports the skipped tests, but jenkins doesn't seem to notice 😞

none of the common tests are generating junit files. I noticed this the other day and am planning to address it.

azeey:
I reverted the cmake upgrade because the core issue was that the tests were not generating junit files. Once that's fixed, skipped tests are reported properly.

j-rivero:
> I'd like to get @j-rivero's heavy_check_mark before merging.

I was wondering why are not using `gz_build_tests` here like in the other directories?

azeey:
These (common) tests are different because they are binaries that take the physics engine plugin as a command line argument. We are generating a test (calling `add_test`) for each test and plugin combination.

j-rivero:
> These (common) tests are different because they are binaries that take the physics engine plugin as a command line argument. We are generating a test (calling `add_test`) for each test and plugin combination.

Oook. I think that the code is mainly a subset of https://github.com/gazebosim/gz-cmake/blob/gz-cmake3/cmake/GzUtils.cmake#L1907-L1936 so it should be fine. We might have problems if the gz-cmake code changes at some point. As a small suggestion maybe adding a timeout to avoid possible locks is a good idea. Something like `set_tests_properties(${target_name} PROPERTIES TIMEOUT 240)`.

j-rivero (approved):

scpeters:
> As a small suggestion maybe adding a timeout to avoid possible locks is a good idea. Something like `set_tests_properties(${target_name} PROPERTIES TIMEOUT 240)`.

adding `set_tests_properties(${target_name} PROPERTIES TIMEOUT 240)` sounds good to me

PR #464, opened 2022-12-29
Resolve joints in nested models

srmainwaring:
# 🦟 Bug fix

Fixes [gz-sim: #1844](https://github.com/gazebosim/gz-sim/issues/1844)

## Summary

The details of the bug and how to reproduce it are documented in [gz-sim: #1844](https://github.com/gazebosim/gz-sim/issues/1844).

In brief: joints in nested models were not being found correctly by the function `EntityManagementFeatures::GetJoint` which resulted in the `Physics` system plugin in `gz-sim` not reporting the joint position and not begin able to actuate the joint.

The change handles the three cases:

1. A joint between two links in a nested model.
2. A joint between a link in a top level model and link in a nested model.
3. A joint between two links in separate nested models.

There is also an issue where DART renames joints in nested models that have duplicate unscoped names. DART expects all names for a given skeleton (= gazebo model) to be unique. It is not practical to expect all Gazebo models to use different names, so we elected to register the joints for nested models using their top-level model scoped names. It may be worth doing this for links and collisions to avoid warnings such as:

```bash
Msg [NameManager::issueNewName] (Skeleton::BodyNode | model2) The name [base_link] is a duplicate, so it has been renamed to [base_link(1)]
Msg [NameManager::issueNewName] (Skeleton::N4dart8dynamics9ShapeNodeE | model2) The name [base_link:collision] is a duplicate, so it has been renamed to [base_link:collision(1)]
Msg [NameManager::issueNewName] (Skeleton::BodyNode | model2) The name [rotor_link] is a duplicate, so it has been renamed to [rotor_link(1)]
Msg [NameManager::issueNewName] (Skeleton::N4dart8dynamics9ShapeNodeE | model2) The name [rotor_link:collision] is a duplicate, so it has been renamed to [rotor_link:collision(1)]
```

but this would be for a separate PR.

An updated version of the test world is attached: [test_nested_model.sdf.zip](https://github.com/gazebosim/gz-physics/files/10321934/test_nested_model.sdf.zip). Note: the tests were carried out using a modified version of the JointPositionController that handles nested models by searching for joints with scoped names. This modification will be submitted in `gz-sim` as a separate PR. 

The joints are moved from the command line using:

```bash
# model1 works without this fix
#-----------------------------

# move the rotor in model1
gz topic -t "/model1/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# model2 illustrates case 1
#-----------------------------

# move the rotor in model2
gz topic -t "/model2/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# move the rotor in nested model21
gz topic -t "/model21/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# model3 illustrates case 2
#-----------------------------

# move the rotor in model3
gz topic -t "/model3/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# move the rotor in nested model31
gz topic -t "/model31/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# move the rotor in nested model32
gz topic -t "/model32/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# model4 illustrates case 3
#-----------------------------

# move the rotor in nested model41
gz topic -t "/model41/cmd_rotor" -m gz.msgs.Double -p "data: 1"

# move the rotor in nested model42
gz topic -t "/model42/cmd_rotor" -m gz.msgs.Double -p "data: 1"
```

Figure: `test_nested_model.sdf` showing all joints repositioned using the joint position controller plugin.

nested-joint-pos-ctrl_2

## Testing (Updated)

There was one test failure in `SDFFeatures_TEST` (L601-L607). The test expected querying a parent model for a joint contained in a nested model to return a non-null pointer. However querying a parent for a grandchild joint should return null in the Gazebo interface even if DART moves the grand child joint to the parent.

```c++
  /// \todo(srmainwaring) this is failing. DART associates the nested joint
  /// with the skeleton of the top level model when the nested model is
  /// joined to the parent model, but Gazebo should not find grandchild
  /// joints when querying a parent model.
  auto nestedJoint = parentModel->GetJoint("nested_joint");
  // EXPECT_NE(nullptr, nestedJoint);
  EXPECT_EQ(nullptr, nestedJoint);
```

## Checklist
- [x] Added tests


azeey (commented):
Thanks for the PR @srmainwaring. It looks like we were handling links in nested models properly for `GetLink`, but not joints in nested models for `GetJoint`. What do you think about storing a list of joints inside `ModelInfo` the same way we store `links` and `nestedModels`?

srmainwaring:
> What do you think about storing a list of joints inside `ModelInfo` the same way we store `links` and `nestedModels`?

Sure - that'd work and be consistent with the link handling. However with the links we still get a DART warning about renaming links when the nested model links are joined to the container model skeleton if there are any duplicated links (which is likely as most models will have a `base_link`). It's not clear when you see this message whether it is something to worry about, and whether it affects the way you should refer to links / joints in plugins.

IIRC the link code uses a world scoped full name for the links. If the same approach is used for nested joints, what is the convention for referring to these joints from plugins? It might be useful to have a document describing the various use cases for referring to joint and link names in plugins for flat and nested models and the Gazebo teams recommendation of best practice. I expect that needs to be something that works for other physics engines as well and is not just driven by DART's particular treatment.

azeey:
> It's not clear when you see this message whether it is something to worry about, and whether it affects the way you should refer to links / joints in plugins.

As you can see in `LinkInfo`, we store the original name of the link so that name changes by DART do not affect the way `GetLink` works.
https://github.com/gazebosim/gz-physics/blob/2a93515f7995a8a060d62acd93c88294e4cae65f/dartsim/src/Base.hh#L62-L64

We could do the same for joints. Also, I think you're right in that we should use fully qualified names when adding entities to DART to avoid this name collision. However, we don't want this to leak into the user API of gz-physics. Ideally, we'd store the original local (unscoped) names of links and joints and allow `GetLink` and `GetJoint` to work with those names.

> IIRC the link code uses a world scoped full name for the links. 

It only uses the full names for other book keeping tasks. For `GetLink`, the user would still use the local/unscoped name
https://github.com/gazebosim/gz-physics/blob/2a93515f7995a8a060d62acd93c88294e4cae65f/dartsim/src/EntityManagementFeatures.cc#L381

We currently don't support accessing links of a nested model from the grand parent model. The issue addressed by keeping links in `ModelInfo` is the failure to find the body in the original skeleton because it has been moved due to a joint. I thought that's what we were trying to resolve for joints in this PR. I'd rather punt on accessing nested elements from grand parent models using scoped names. I think it's better to use `entitiesFromScopedName` in gz-sim for now.

srmainwaring:
@azeey all makes sense. The PR for the `JointPositionController` https://github.com/gazebosim/gz-sim/pull/1851 uses `entitiesFromScopedName` to resolve scoped names. There might need to be a little tweak given the changes suggested here, but I think for the most part it will work.

I'll start drafting your suggested changes in this PR following the approach for links.

srmainwaring:
> What do you think about storing a list of joints inside `ModelInfo` the same way we store `links` and `nestedModels`?

@azeey I've reworked the PR to use a similar approach to that used for links. The example provided above works.

I've promoted a method GetModelOfLinkImpl to dartsim::Base as it's required in JointFeatures which does not subclass from EntityManagementFeatures. There is some further consolidation that can be done in JointFeatures to reduce duplicated code, but would appreciate your feedback that this is on the right track before further refinements.

dartsim/src/Base.hh
```diff
@@ -488,17 +493,37 @@ class Base : public Implements3d<FeatureList<Feature>>
...
     const std::size_t id = this->GetNextEntity();
-    this->joints.idToObject[id] = std::make_shared<JointInfo>();
+    auto jointInfo = std::make_shared<JointInfo>();
+    this->joints.idToObject[id] = jointInfo;
+    jointInfo->joint = _joint;
+    // The name of the dart Joint during creation is assumed to be the
+    // Gazebo-specified name.
+    jointInfo->name = _joint->getName();
+
     this->joints.idToObject[id]->joint = _joint;
```
azeey, line 507: nit: I think this can be removed now since `jointInfo->joint` is set above.

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -436,25 +442,37 @@ Identity EntityManagementFeatures::GetJoint(
...
-  // If the joint doesn't exist in "joints", it means the containing entity has
-  // been removed.
-  if (this->joints.HasEntity(joint))
-  {
-    const std::size_t jointID = this->joints.IdentityOf(joint);
-    return this->GenerateIdentity(jointID, this->joints.at(jointID));
-  }
-  else
+  const auto &modelInfo = this->ReferenceInterface<ModelInfo>(_modelID);
+  for (const auto &jointInfo : modelInfo->joints)
```
azeey, line 451: Now that we have `jointsByName`,  can we just lookup the joint's name in there instead of having a for loop here?

dartsim/src/SDFFeatures_TEST.cc
```diff
@@ -537,6 +537,45 @@ TEST_P(SDFFeatures_TEST, WorldIsParentOrChild)
 }
 
 /////////////////////////////////////////////////
+/// \todo(srmainwaring) remove when behaviour resolved.
```
azeey, line 540: This would actually be useful to keep.

dartsim/src/SDFFeatures_TEST.cc
```diff
@@ -559,8 +598,13 @@ TEST_P(SDFFeatures_TEST, WorldWithNestedModel)
...
   ASSERT_NE(nullptr, nestedModel);
 
+  /// \todo(srmainwaring) this is failing. DART associates the nested joint
+  /// with the skeleton of the top level model when the nested model is
+  /// joined to the parent model, but Gazebo should not find grandchild
+  /// joints when querying a parent model.
   auto nestedJoint = parentModel->GetJoint("nested_joint");
-  EXPECT_NE(nullptr, nestedJoint);
+  // EXPECT_NE(nullptr, nestedJoint);
+  EXPECT_EQ(nullptr, nestedJoint);
```
azeey, line 607: I agree with your analysis and the change you made here. We should also add

```
EXPECT_NE(nullptr, nestedModel->GetJoint("nested_joint"));
```

azeey (commented):
I've made some small comments, but the overall approach looks great!

dartsim/src/Base.hh
srmainwaring: Updated in af9f7cca.

dartsim/src/EntityManagementFeatures.cc
srmainwaring: Updated in 9553dea7 (using the internally generated fully scoped joint name).

dartsim/src/SDFFeatures_TEST.cc
srmainwaring: Update formatting in b03b284c.

dartsim/src/SDFFeatures_TEST.cc
srmainwaring: Updated in 539e8be6.

srmainwaring: @azeey thanks for the feedback. I've made the changes in a series of commits in case we want to revert any of them.

I've left some debug helpers in Base.h with \todo removes against them. I found them very useful for tracing what was going on in the various maps. Let me know if you'd like them removed or left (in which case I'll update the comment).

Also, in JointFeatures.cc I've left three \todo's about consolidating some common code (generating fully scoped names for joints). I was going to add a method in an anonymously namespace but the function to generate identities are protected (so I can't pass the this pointer to a non-class function to do the work). Can leave as is or consolidate in a member function if you think that would be better.

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -436,25 +442,41 @@ Identity EntityManagementFeatures::GetJoint(
...
-    return this->GenerateInvalidId();
+    auto joint = it->second;
+    if (this->joints.HasEntity(joint))
+    {
+      const std::size_t jointID = this->joints.IdentityOf(joint);
+      return this->GenerateIdentity(jointID, this->joints.at(jointID));
+    }
   }
+
+  // TODO(anyone) It's not clear what to do when `GetLink` is called on a
```
azeey, line 472: nit: `GetLink` should be `GetJoint`.

dartsim/src/SDFFeatures_TEST.cc
azeey: Since these classes aren't directly exposed to the user, i.e, the plugin implementation is an internal detail, we don't have a lot of constraints as to what should be a free function vs a member function, so please go ahead and consolidate the common code in a member function.

dartsim/src/EntityManagementFeatures.cc
srmainwaring: Fixed in b8b6842d

dartsim/src/SDFFeatures_TEST.cc
srmainwaring: If you're ok with leaving helper functions, then I'll remove the `\todo` comments from

https://github.com/gazebosim/gz-physics/blob/b8b6842d3ed990e849c2379f118584c053ee2db2/dartsim/src/Base.hh#L644-L668

etc. and leave them in place as they are very useful for debugging (I find it's difficult to inspect the maps directly in gdb / lldb). I'll add a helper for the commonly used code creating the fully scoped name and place it where it can be accessed as required.

srmainwaring:
Hi @azeey - I think this PR now includes all the changes you requested and has been tidied up. Is there anything else you'd like completed to get this merged?

dartsim/src/EntityManagementFeatures.cc
```diff
@@ -436,25 +442,41 @@ Identity EntityManagementFeatures::GetJoint(
...
+          << "] could not be found for joint [" << _jointName
+          << "]\n";
+    return this->GenerateInvalidId();
   }
-  else
+
+  auto world = this->worlds.at(worldID);
+  const std::string fullJointName = ::sdf::JoinName(
+      world->getName(),
+      ::sdf::JoinName(modelInfo->model->getName(), _jointName));
```
azeey, line 459: Can this use `FullyScopedJointName` instead?

dartsim/src/Base.hh
```diff
@@ -587,6 +607,49 @@ class Base : public Implements3d<FeatureList<Feature>>
...
+
+  /// \brief Create a fully (world) scoped joint name.
+  /// \param _modelID Identity of the parent model of the joint's child link.
+  /// \param _childID Identity of the joint's child link.
+  /// \param _name The unscoped joint name.
+  /// \return The fully (world) scoped joint name, or an empty string
+  /// if a world cannot be resolved.
+  public: inline std::string FullyScopedJointName(
+    const Identity &_modelID,
+    const Identity &_childID,
```
azeey, line 631: `_childID` is not being used.

azeey (approved):
Just a couple of small comments. Otherwise, LGTM!

dartsim/src/Base.hh
srmainwaring: You're right - removed in 8137d90a

dartsim/src/EntityManagementFeatures.cc
srmainwaring: Good spot - updated in 47b46868

PR #475, opened 2023-01-23
Do not install common tests

j-rivero:
# 🦟 Bug fix

## Summary

Nightlies are complaining about the missing installation of some test files that were added to the `gz-phsyisc6` branch. I assume that we don't want to install them but please let me know otherwise.

## Checklist
- [ ] Added tests

### DIFF
--- a/test/common_test/CMakeLists.txt
+++ b/test/common_test/CMakeLists.txt
@@ -15,8 +15,6 @@ set(tests
   world_features
 )
 
-set(TEST_INSTALL_DIR ${CMAKE_INSTALL_LIBEXECDIR}/gz/${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}/)
-
 # Find the Python interpreter for running the
 # check_test_ran.py script
 include(GzPython)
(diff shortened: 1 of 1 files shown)


ahcorde:
maybe @scpeters might have a different opinion

scpeters:
yes, we do want to install these tests so that they can by used by gz-physics developers who are implementing 3rd-party physics plugins. We are planning to use these installed tests with https://github.com/gazebosim/gz-chrono for example

I think we just need to decide in which debian package we want to install the tests

mjcarroll:
> I think we just need to decide in which debian package we want to install the tests

Since the testing is largely contained to people creating new physics implementations (developing), I think it makes sense to land them in `-dev`.

j-rivero:
> yes, we do want to install these tests so that they can by used by gz-physics developers who are implementing 3rd-party physics plugins.

Umm interesting, so they are both test and tools at the same time, if I understand correctly. For the installation, the path in libexec seems fine if we expect them to be used as "auxiliary tools" although the name of `COMMON_TEST` is a bit unusual but we can live with that if the tools are in fact testing something.

For the package to put them, unless we want to create a custom one called -tools, probably -dev is the best place indeed.

mjcarroll:
Maybe we should update the installed binary name @scpeters and then install with -dev seems the best approach.

PR #484, opened 2023-03-13
Place virtual added mass at link origin

mjcarroll:


scpeters:
is the goal here to keep the fluid_added_mass inertial properties distinct from the rigid body inertial properties for introspection purposes? It sounds from the title of the PR that this changes the effective 3D point where the inertial parameters are expressed

PR #502, opened 2023-04-03
Disable failing bullet(-featherstone) tests on windows

azeey:
# 🦟 Bug fix

## Summary
We've recently accidentally started testing the Bullet and Bullet-featherstone plugins on Windows because one of our Jenkins nodes had Bullet installed. This created a few failures that looked like regressions, but were never meant to be tested. Instead of removing Bullet from the Jenkins node, we've decided to disable the failing tests so it's more explicit and so it'd be less steps to take in case we wanted to fix them down the road.

Also in this PR, I noticed that some of the tests were loading worlds from all plugins simultaneously instead of loading each plugin and world one by one. ~Perhaps some of the test failures are caused by that~, but in any case, I think we want to load the plugins one at a time, so I've made changes accordingly.
 
## Checklist
- [ ] Added tests

### DIFF
--- a/test/common_test/TestLibLoader.hh
+++ b/test/common_test/TestLibLoader.hh
@@ -19,6 +19,7 @@
 
 #include <iostream>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include <gz/common/Filesystem.hh>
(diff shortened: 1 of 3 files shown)


mjcarroll:
Maybe worth adding something like https://github.com/gazebosim/gz-rendering/blob/gz-rendering7/test/common_test/CommonRenderingTest.hh#L127-L138  to check for supported/unsupported?

azeey:
> Maybe worth adding something like https://github.com/gazebosim/gz-rendering/blob/gz-rendering7/test/common_test/CommonRenderingTest.hh#L127-L138 to check for supported/unsupported?

Done in https://github.com/gazebosim/gz-physics/pull/502/commits/0a86b2c4d63285e981a75b54d7d45740b15e86ce

azeey:
The ABI checker is failing due to what seems the node running out of memory. Since this is only disabling tests, there should be no change in the ABI. Merging!

Issue #497, opened 2023-03-24
:farmer: Segfaults in Homebrew CI cause test failures


Crola1702:
## Environment
* OS Version: Mac OS bigsur
* Source Garden build

## Description
* Expected behavior: `COMMON_TEST_simulation_features_bullet-featherstone` should pass
* Actual behavior: `COMMON_TEST_simulation_features_bullet-featherstone` and `INTEGRATION_DoublePendulum` are failing due to a Segfault

## Steps to reproduce
1. Run a build in [gz-physics6-homebrew](https://build.osrfoundation.org/view/ign-garden/job/ignition_physics-ci-gz-physics6-homebrew-amd64/)
2. See `COMMON_TEST_simulation_features_bullet-featherstone` fail.
3. If you enter the log, you can see `INTEGRATION_DoublePendulum` also failed due to a segfault.

## Output

Reference build: https://build.osrfoundation.org/view/ign-garden/job/ignition_physics-ci-gz-physics6-homebrew-amd64/55/

The following tests FAILED:
* 105 - COMMON_TEST_simulation_features_bullet-featherstone (SEGFAULT)
* 121 - INTEGRATION_DoublePendulum (SEGFAULT) (This one is failing since [build 37](https://build.osrfoundation.org/view/ign-garden/job/ignition_physics-ci-gz-physics6-homebrew-amd64/37/consoleFull), but for some reason, Jenkins doesn't show it on the report)

Log output:

COMMON_TEST_simulation_features_bullet-featherstone:
<details>

```
        Start 105: COMMON_TEST_simulation_features_bullet-featherstone

105: Test command: /Users/jenkins/jenkins-agent/workspace/ignition_physics-ci-gz-physics6-homebrew-amd64/build/bin/COMMON_TEST_simulation_features "/Users/jenkins/jenkins-agent/workspace/ignition_physics-ci-gz-physics6-homebrew-amd64/build/lib/libgz-physics6-bullet-featherstone-plugin.6.3.1.dylib" "--gtest_output=xml:/Users/jenkins/jenkins-agent/workspace/ignition_physics-ci-gz-physics6-homebrew-amd64/build/test_results/COMMON_TEST_simulation_features_bullet-featherstone.xml"
105: Working Directory: /Users/jenkins/jenkins-agent/workspace/ignition_physics-ci-gz-physics6-homebrew-amd64/build/test/common_test
105: Test timeout computed to be: 240
105: [==========] Running 9 tests from 6 test suites.
[...]

PR #504, opened 2023-04-06
Fix bullet related segfaults on macOS

azeey:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-physics/issues/497

## Summary
This fixes two tests with segfaults on macOS
1. `COMMON_TEST_simulation_features_bullet-featherstone`: The failure appears to be due to the destruction order of the `triangleMeshes` and `meshesGImpact` containers.  Per [this] (https://github.com/bulletphysics/bullet3/issues/567#issuecomment-186248424) comment, memory in bullet is sensitive to order of destruction. `meshesGImpact` objects contain pointers to items contained by`triangleMeshes`, so the solution is to ensure `meshesGImpact` is destructed before `triangleMeshes`.
2. `INTEGRATION_DoublePendulum`: This is a DART test, but the segfault seems to be related to Bullet. I found that even after removing all the code in `DARTDoublePendulum.cc`, the segfault still occurs as long as it's linked with `libdart-collision-bullet`. Since the test doesn't need the bullet collision detector, I've opted to just remove the library from the `DART_LIBRARIES` variable. I still don't know the root cause though, so this might affect the actual dartsim-plugin library.

## Checklist
- [ ] Added tests

### DIFF
--- a/bullet-featherstone/src/Base.hh
+++ b/bullet-featherstone/src/Base.hh
@@ -354,8 +354,10 @@ class Base : public Implements3d<FeatureList<Feature>>
   public: std::unordered_map<std::size_t, CollisionInfoPtr> collisions;
   public: std::unordered_map<std::size_t, JointInfoPtr> joints;
 
-  public: std::vector<std::unique_ptr<btGImpactMeshShape>> meshesGImpact;
+  // Note, the order of triangleMeshes and meshesGImpact is important. Reversing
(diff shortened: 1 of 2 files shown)


mjcarroll (approved):

azeey:
The windows warnings are coming from gz-math and I believe are fixed by https://github.com/gazebosim/gz-math/pull/529

PR #515, opened 2023-06-07
Minor cleanup of dartsim

shameekganguly:
- Change remaning ignerr to gzerr
- Clangtidy fixes
- Remove unused header

### DIFF
--- a/dartsim/src/ShapeFeatures.cc
+++ b/dartsim/src/ShapeFeatures.cc
@@ -252,7 +252,7 @@ Vector3d ShapeFeatures::GetEllipsoidShapeRadii(
 Identity ShapeFeatures::AttachEllipsoidShape(
     const Identity &_linkID,
     const std::string &_name,
-    const Vector3d &_radii,
+    const Vector3d _radii,
     const Pose3d &_pose)
 {
   common::MeshManager *meshMgr = common::MeshManager::Instance();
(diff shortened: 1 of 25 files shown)


dartsim/src/ShapeFeatures.hh
```diff
@@ -148,7 +148,7 @@ class ShapeFeatures :
   public: Identity AttachEllipsoidShape(
       const Identity &_linkID,
       const std::string &_name,
-      const Vector3d _radii,
+      Vector3d _radii,
```
iche033, line 151: I think it was meant to be `const Vector3d &_radii`

shameekganguly: Done

dartsim/src/ShapeFeatures.cc
```diff
@@ -252,7 +252,7 @@ Vector3d ShapeFeatures::GetEllipsoidShapeRadii(
 Identity ShapeFeatures::AttachEllipsoidShape(
     const Identity &_linkID,
     const std::string &_name,
-    const Vector3d _radii,
+    const Vector3d &_radii,
```
azeey, line 255: I think this is a breaking change, so we wouldn't want to do it in this branch. Instead, it should go into `main`. Also, to do this, we'd have to change the signatures in the `AttachEllipsoidShapeFeature` (https://github.com/gazebosim/gz-physics/blob/8ee8028861567303cdba2602c9ba79c34a5ab542/include/gz/physics/EllipsoidShape.hh)

scpeters: I've applied these API changes in https://github.com/gazebosim/gz-physics/pull/521, targeting `main`

scpeters: I reverted this change in https://github.com/gazebosim/gz-physics/pull/515/commits/39e475315769ed20a77e5c2e8ff1e3fb32795f30

iche033 (approved):

shameekganguly:
Thanks Steve, looks good!

PR #551, opened 2023-09-25
Make dartsim/World.hh header private

j-rivero:
Move `dartsim/World.hh` header out from public headers to be in `src/` as a private header as suggested long ago in https://github.com/gazebosim/gz-physics/pull/274#issuecomment-893991310

## Checklist
- [ ] Added tests


azeey (approved):

scpeters (approved):

azeey:
After this PR, would  the `libgz-physics7-dartsim-dev` package become obsolete?

scpeters:
> After this PR, would the `libgz-physics7-dartsim-dev` package become obsolete?

that still holds the cmake config files, so I think it's needed?

* https://github.com/gazebo-release/gz-physics7-release/blob/main/ubuntu/debian/libgz-physics-dartsim-dev.install

j-rivero:
> > After this PR, would the `libgz-physics7-dartsim-dev` package become obsolete?
> 
> that still holds the cmake config files, so I think it's needed?
> 
>     * https://github.com/gazebo-release/gz-physics7-release/blob/main/ubuntu/debian/libgz-physics-dartsim-dev.install 
> 
> that still holds the cmake config files, so I think it's needed?

This is a good point. I think that the main question is: is there a valid use for developers that would generate code from the dartsim plugin that has no public headers? Because we are generating the buildsystem helpers (CMake + pkgconfig) to facilitate that use case. If there is no use case, we either not packaging the helpers and remove the -dev package or change our buildsystem not to generate the helpers.

j-rivero:
Let's move the conversation to https://github.com/gazebo-release/gz-physics7-release/pull/3

PR #582, opened 2023-12-20
Support setting max contacts in dartsim's ODE collision detector

iche033:
# 🎉 New feature

## Summary
Support setting max number of contacts between collision pairs.

* Added a new `MaxContacts` feature. 
* Support for this is currently only implemented when the ode collision detector is used in dartsim (which is the default collision detector).
  * This is done by overriding the `collide` function in dartsim's `OdeCollisionDetector` class to post process the contact results. Not ideal as it involves copying contact data. It would be nice if dart provides some hooks / callback into their collision checking process e.g. [here](https://github.com/dartsim/dart/blob/b7f5dd1cee755fe0e7a43c8c83794f22452501c1/dart/collision/ode/OdeCollisionDetector.cpp#L283-L286) so we can avoid the copy.

Added test to make sure the max contact value does limit the no. of contacts.

See https://github.com/gazebosim/gz-sim/pull/2270 for testing with gz-sim.

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/GzOdeCollisionDetector.cc
+++ b/dartsim/src/GzOdeCollisionDetector.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Open Source Robotics Foundation
+ * Copyright (C) 2024 Open Source Robotics Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/dartsim/src/GzOdeCollisionDetector.cc
+++ b/dartsim/src/GzOdeCollisionDetector.cc
@@ -94,9 +94,6 @@ void GzOdeCollisionDetector::LimitMaxContacts(
 
   for (auto &contact : allContacts)
   {
-    //auto contactPair =
-    //  std::make_pair(contact.collisionObject1, contact.CollisionObject2);
-    //auto &count = contactMap[contactPair];
     auto &count = contactMap[contact.collisionObject1][contact.collisionObject2];
     count++;
     auto &otherCount = contactMap[contact.collisionObject2][contact.collisionObject1];
--- a/dartsim/src/GzOdeCollisionDetector.cc
+++ b/dartsim/src/GzOdeCollisionDetector.cc
@@ -52,7 +52,7 @@ bool GzOdeCollisionDetector::collide(
     CollisionResult *_result)
 {
   bool ret = OdeCollisionDetector::collide(_group, _option, _result);
-  this->LimitCollisionPairMaxTotalContacts(_result);
+  this->LimitCollisionPairMaxContacts(_result);
   return ret;
 }
 
@@ -64,25 +64,25 @@ bool GzOdeCollisionDetector::collide(
     CollisionResult *_result)
 {
   bool ret = OdeCollisionDetector::collide(_group1, _group2, _option, _result);
-  this->LimitCollisionPairMaxTotalContacts(_result);
+  this->LimitCollisionPairMaxContacts(_result);
   return ret;
 }
 
 /////////////////////////////////////////////////
-void GzOdeCollisionDetector::SetCollisionPairMaxTotalContacts(
+void GzOdeCollisionDetector::SetCollisionPairMaxContacts(
     std::size_t _maxContacts)
 {
   this->maxCollisionPairContacts = _maxContacts;
 }
 
 /////////////////////////////////////////////////
-std::size_t GzOdeCollisionDetector::GetCollisionPairMaxTotalContacts() const
+std::size_t GzOdeCollisionDetector::GetCollisionPairMaxContacts() const
 {
   return this->maxCollisionPairContacts;
 }
 
 /////////////////////////////////////////////////
-void GzOdeCollisionDetector::LimitCollisionPairMaxTotalContacts(
+void GzOdeCollisionDetector::LimitCollisionPairMaxContacts(
     CollisionResult *_result)
 {
   if (this->maxCollisionPairContacts ==
--- a/dartsim/src/GzOdeCollisionDetector.cc
+++ b/dartsim/src/GzOdeCollisionDetector.cc
@@ -15,6 +15,9 @@
  *
 */
 
+#include <memory>
+#include <unordered_map>
+
 #include <dart/collision/CollisionObject.hpp>
 
 #include "GzOdeCollisionDetector.hh"
@@ -94,9 +97,11 @@ void GzOdeCollisionDetector::LimitMaxContacts(
 
   for (auto &contact : allContacts)
   {
-    auto &count = contactMap[contact.collisionObject1][contact.collisionObject2];
+    auto &count =
+      contactMap[contact.collisionObject1][contact.collisionObject2];
     count++;
-    auto &otherCount = contactMap[contact.collisionObject2][contact.collisionObject1];
+    auto &otherCount =
+      contactMap[contact.collisionObject2][contact.collisionObject1];
     std::size_t total =  count + otherCount;
     if (total <= this->maxCollisionPairContacts)
     {
--- a/dartsim/src/GzOdeCollisionDetector.cc
+++ b/dartsim/src/GzOdeCollisionDetector.cc
@@ -100,5 +100,4 @@ void GzOdeCollisionDetector::LimitMaxContacts(
     // std::cerr << "2 " << contact.collisionObject2 << std::endl;
     // std::cerr << "count " << count << std::endl;
   }
-  // std::cerr << "contacts " << _result->getNumContacts() << std::endl;
 }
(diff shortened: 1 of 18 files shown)


azeey:
I see the upstream Ode collision detector has the option to limit contacts https://github.com/dartsim/dart/blob/b7f5dd1cee755fe0e7a43c8c83794f22452501c1/dart/collision/ode/OdeCollisionDetector.cpp#L317. Is it not possible to use that?

iche033:
> I see the upstream Ode collision detector has the option to limit contacts https://github.com/dartsim/dart/blob/b7f5dd1cee755fe0e7a43c8c83794f22452501c1/dart/collision/ode/OdeCollisionDetector.cpp#L317. Is it not possible to use that?

That's the flag we're setting in [EntityManagementFeatures.cc](https://github.com/gazebosim/gz-physics/blob/780e2d0b08a373d586c6a05bf7445094e13daa1c/dartsim/src/EntityManagementFeatures.cc#L732). I noticed that it limits the max number of contacts for all entities instead of max contacts per collision pair.

azeey:
> That's the flag we're setting in [EntityManagementFeatures.cc](https://github.com/gazebosim/gz-physics/blob/780e2d0b08a373d586c6a05bf7445094e13daa1c/dartsim/src/EntityManagementFeatures.cc#L732). I noticed that it limits the max number of contacts for all entities instead of max contacts per collision pair.

Ah, I think I understand now. So, this PR makes it so that each collision pair will have the same number of max contacts. This as opposed to the total number of contacts in the world being limited to the max contact parameter. Is that right?

I think Gazebo-classic also had a way to set the max contact parameter for each `<collision>`. This is different from that right?

iche033:
> This as opposed to the total number of contacts in the world being limited to the max contact parameter. Is that right?

yes that's correct

> I think Gazebo-classic also had a way to set the max contact parameter for each <collision>. This is different from that right?

yeah that's a related feature which gives users finer control of max contacts for a specific collision.

azeey:
@azeey  to review.

scpeters:
> > This as opposed to the total number of contacts in the world being limited to the max contact parameter. Is that right?
> 
> yes that's correct
> 
> > I think Gazebo-classic also had a way to set the max contact parameter for each . This is different from that right?
> 
> yeah that's a related feature which gives users finer control of max contacts for a specific collision.

I think we should be careful about naming the features and APIs to make it clear which interpretation of max contacts is being used. Here's some suggested feature names:

* maximum total number of contacts allowed in a world: `gz::physics::WorldMaxTotalContacts`
* default maximum number of contacts for a collision pair: `gz::physics::CollisionPairMaxTotalContacts`
* maximum number of contacts for a specific collision, which overrides the `CollisionPairMaxTotalContacts` value, and requires some logic for merging the parameters for each collision shape in a pair if they both have it specified: `gz::physics::CollisionMaxTotalContacts`

iche033:
> I think we should be careful about naming the features and APIs to make it clear which interpretation of max contacts is being used. Here's some suggested feature names:

Great, thanks for the suggestions. For the feature in this PR, I renamed it from `MaxContacts` to `gz::physics::CollisionPairMaxTotalContacts`. [98b5e61](https://github.com/gazebosim/gz-physics/pull/582/commits/98b5e6108accc7b160500c83274666b2e16f007a)

scpeters:
> > I think we should be careful about naming the features and APIs to make it clear which interpretation of max contacts is being used. Here's some suggested feature names:
> 
> Great, thanks for the suggestions. For the feature in this PR, I renamed it from `MaxContacts` to `gz::physics::CollisionPairMaxTotalContacts`. [98b5e61](https://github.com/gazebosim/gz-physics/pull/582/commits/98b5e6108accc7b160500c83274666b2e16f007a)

thanks; I just wrote the first thing that came to mind; we can see if @azeey has any different naming suggestions. At risk of bike-shedding, I think I only meant to include `Total` in the `WorldMaxTotalContacts` feature name, so the feature in this PR could be shortened to `CollisionPairMaxContacts`, but I'll let @azeey make the call (and apologies for the churn)

dartsim/src/GzOdeCollisionDetector.hh
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (C) 2023 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2024 Open Source Robotics Foundation
```

iche033: [e4e51e1](https://github.com/gazebosim/gz-physics/pull/582/commits/e4e51e1ebcd041981e4047aeb75adaa0f771875c)

iche033:
> At risk of bike-shedding, I think I only meant to include Total in the WorldMaxTotalContacts feature name, so the feature in this PR could be shortened to CollisionPairMaxContacts, but I'll let @azeey make the call (and apologies for the churn)

ok no problem. @azeey any thoughts on naming? I'll make the change if Addisu agrees.

azeey:
`CollisionPairMaxContacts` sounds good to me. I think having "total" would add confusion since it might imply the total number of contacts in the world.

iche033:
> CollisionPairMaxContacts sounds good to me.

ok updated! [b140329](https://github.com/gazebosim/gz-physics/pull/582/commits/b140329abacec3cf1961b1dc7924d4985fed556f)

include/gz/physics/World.hh
```diff
@@ -130,6 +130,45 @@ namespace gz
...
+      /// \brief The World API for getting and setting max contacts.
+      public: template <typename PolicyT, typename FeaturesT>
+      class World : public virtual Feature::World<PolicyT, FeaturesT>
+      {
+        /// \brief Set the maximum number of contacts allowed between two
+        /// entities.
+        /// \param[in] _maxContacts Maximum number of contacts.
+        public: void SetCollisionPairMaxContacts(std::size_t _maxContacts);
+
+        /// \brief Set the maximum number of contacts allowed between two
```
scpeters, line 146:
```suggestion
        /// \brief Get the maximum number of contacts allowed between two
```

include/gz/physics/World.hh
```diff
@@ -130,6 +130,45 @@ namespace gz
       };
     };
 
+    /////////////////////////////////////////////////
+    class GZ_PHYSICS_VISIBLE CollisionPairMaxContacts:
+        public virtual Feature
+    {
+      /// \brief The World API for getting and setting max contacts.
```
scpeters, line 137:
```suggestion
      /// \brief The World API for getting and setting the maximum
      number of contacts between two entities.
```

include/gz/physics/World.hh
```diff
@@ -130,6 +130,45 @@ namespace gz
...
+        /// contacts between two entities.
+        /// \param[in] _id Identity of the world.
+        /// \param[in] _maxContacts Maximum number of contacts.
+        public: virtual void SetWorldCollisionPairMaxContacts(
+            const Identity &_id, std::size_t _maxContacts) = 0;
+
+        /// \brief Implementation API for getting the maximum number of
+        /// contacts between two entities.
+        /// \param[in] _id Identity of the world.
+        /// \param[in] _maxContacts Maximum number of contacts.
```
scpeters, line 166:
```suggestion
        /// \return Maximum number of contacts.
```

dartsim/src/GzOdeCollisionDetector.cc
```diff
@@ -0,0 +1,112 @@
...
+
+  for (auto &contact : allContacts)
+  {
+    auto &count =
+      contactMap[contact.collisionObject1][contact.collisionObject2];
+    count++;
+    auto &otherCount =
+      contactMap[contact.collisionObject2][contact.collisionObject1];
+    std::size_t total =  count + otherCount;
+    if (total <= this->maxCollisionPairContacts)
```
scpeters, line 107: my only other thought is about how to choose between multiple contact points if we need to drop some, but I see that is what #584 is for, so I'll go look at that now

scpeters (commented):
just a few places where the documentation could be cleaned up

include/gz/physics/World.hh
iche033: [89d6347](https://github.com/gazebosim/gz-physics/pull/582/commits/89d63477103ce082d334ebd2ea92f31109c6500b)

include/gz/physics/World.hh
iche033: [89d6347](https://github.com/gazebosim/gz-physics/pull/582/commits/89d63477103ce082d334ebd2ea92f31109c6500b)

include/gz/physics/World.hh
iche033: [89d6347](https://github.com/gazebosim/gz-physics/pull/582/commits/89d63477103ce082d334ebd2ea92f31109c6500b)

dartsim/src/GzOdeCollisionDetector.cc
iche033: responded to comments in https://github.com/gazebosim/gz-physics/pull/584

scpeters (approved):

PR #584, opened 2023-12-27
Minor optimization for picking contact points in dartsim's ODE collision detector 

iche033:
# 🎉 New feature

## Summary

Depends on https://github.com/gazebosim/gz-physics/pull/582.

This PR implements a minor optimization for picking contact points when too many contacts are generated. The logic implemented is based on the one used in gazebo classic, specifically this [piece of code](https://github.com/gazebosim/gazebo-classic/blob/7ccef40e24831eb5cd974b489ee279fc064eacc4/gazebo/physics/ode/ODEPhysics.cc#L1082-L1097). 

My understanding is that gazebo-classic will pick the first `n-1` contact points  between a pair of collisions that were generated, where `n` is the max allowed contact points. It then selects the last contact point by comparing all remaining contact points and picks the one with the largest penetration depth.

## Test it

You can test this together with gz-sim using https://github.com/gazebosim/gz-sim/pull/2270

Launching the [contacts.sdf](https://gist.github.com/iche033/3f5f080bc53c7196f2ebdfeebb54a0f8) world will show the following output:

**Before:**
Contact points located at upper half of the capsule:

max_contacts

**After:**
Contact points located at upper half of the capsule + 1 last contact point at the bottom end of the capsule:

contact_point_optimize

## Checklist
- [ ] Added tests


dartsim/src/GzOdeCollisionDetector.cc
```diff
@@ -92,21 +93,45 @@ void GzOdeCollisionDetector::LimitCollisionPairMaxContacts(
   auto allContacts = _result->getContacts();
   _result->clear();
 
+
+  if (this->maxCollisionPairContacts == 0u)
+    return;
+
+  // A map of collision pairs and their their contact info
```
scpeters, line 100:
```suggestion
  // A map of collision pairs and their contact info
```

dartsim/src/GzOdeCollisionDetector.cc
```diff
@@ -92,21 +93,45 @@ void GzOdeCollisionDetector::LimitCollisionPairMaxContacts(
...
+        otherLastContactIdx = lastContactIdx;
+      }
       _result->addContact(contact);
     }
+    else
+    {
+      // If too many contacts were generated, replace the last contact point
+      // of the collision pair with one that has a larger penetration depth
+      auto &c = _result->getContact(lastContactIdx);
+      if (contact.penetrationDepth > c.penetrationDepth)
```
scpeters, line 131: this does match what gazebo-classic was doing, so I think it's a reasonable place to start, but reading this code raises questions for me about what should be done.

One approach could be to sort contacts by penetration depth and choose the points with the maximum depth values, but then you could still have to choose between points if their depth values were identical. Then again, the computational cost must also be considered

scpeters (commented):
> You can test this together with gz-sim using [gazebosim/gz-sim#2270](https://github.com/gazebosim/gz-sim/pull/2270)
> 
> Launching the [contacts.sdf](https://gist.github.com/iche033/3f5f080bc53c7196f2ebdfeebb54a0f8) world will show the following output:
> 
> **Before:** Contact points located at upper half of the capsule:
> ...
> **After:** Contact points located at upper half of the capsule + 1 last contact point at the bottom end of the capsule:
> ...

Is it possible to make even a rudimentary test to confirm the changes proposed here? For example, if we put two objects in collision for a single time step with `CollisionPairMaxContacts == 1`, we should expect the only contact point to be the one with maximum depth

dartsim/src/GzOdeCollisionDetector.cc
iche033: [20ccc51](https://github.com/gazebosim/gz-physics/pull/584/commits/20ccc51a3088fddb8f1e95a18123ca77c916c845)

iche033:
> Is it possible to make even a rudimentary test to confirm the changes proposed here? For example, if we put two objects in collision for a single time step with CollisionPairMaxContacts == 1, we should expect the only contact point to be the one with maximum depth

yep added test in [20ccc51](https://github.com/gazebosim/gz-physics/pull/584/commits/20ccc51a3088fddb8f1e95a18123ca77c916c845) to verify that the contact with max depth is picked.

dartsim/src/GzOdeCollisionDetector.cc
iche033: yeah I think that's a good approach and see if how much extra sorting costs. We can consider this as a follow up improvement?

scpeters: > yeah I think that's a good approach and see if how that much extra sorting logic costs. We can consider this as a follow up improvement?

👍

test/common_test/simulation_features.cc
```diff
@@ -275,6 +275,62 @@ TYPED_TEST(SimulationFeaturesCollisionPairMaxContactsTest,
...
+      ASSERT_NE(nullptr, extraContactData);
+      if (extraContactData->depth > maxDepth)
+        maxDepth = extraContactData->depth;
+    }
+    EXPECT_GT(maxDepth, 0.0);
+
+    // Set max contact between collision pairs to be 1
+    world->SetCollisionPairMaxContacts(1u);
+    EXPECT_EQ(1u, world->GetCollisionPairMaxContacts());
+    checkedOutput = StepWorld<FeaturesCollisionPairMaxContacts>(
```
scpeters, line 315: this test assumes that the objects will be at the same position for both time steps, which would be the case if the `collide_without_contact` parameter was supported and set to `true`, but otherwise the contact could cause the objects to move. I think it may work for the contact parameters we have set in dartsim, but it may not work in general.

Maybe it would be more repeatable in general if we explicitly reset the object poses to the same value before each step?

iche033: ok makes sense.  I updated the test to reset the objects to their initial pose between steps. [115b541](https://github.com/gazebosim/gz-physics/pull/584/commits/115b541a8e22bf966150930d0aa568e794ebdbe9)

iche033: @scpeters friendly ping, does this look good to you?

scpeters (approved):

scpeters:
conflicts

scpeters:
> conflicts

never mind

PR #598, opened 2024-02-06
bazel: Updates for garden (minus ABI break)

mjcarroll:
Rework #513 to not break ABI.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -7,6 +7,10 @@ load(
     "gz_export_header",
     "gz_include_header",
 )
+load(
+    "@gz//bazel/lint:lint.bzl",
+    "add_lint_tests",
+)
 load("@rules_license//rules:license.bzl", "license")
 
 package(
@@ -132,3 +136,5 @@ test_sources = glob(["src/*_TEST.cc"])
         "@gtest//:gtest_main",
     ],
 ) for src in test_sources]
+
+add_lint_tests()
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -7,18 +7,25 @@ load(
     "gz_export_header",
     "gz_include_header",
 )
+load("@rules_license//rules:license.bzl", "license")
 
 package(
+    default_applicable_licenses = [GZ_ROOT + "physics:license"],
     default_visibility = GZ_VISIBILITY,
     features = GZ_FEATURES,
 )
 
-licenses(["notice"])  # Apache-2.0
+license(
+    name = "license",
+    package_name = "gz-physics",
+)
+
+licenses(["notice"])
 
 exports_files(["LICENSE"])
 
 gz_configure_header(
-    name = "physics_config_hh",
+    name = "config",
     src = "include/gz/physics/config.hh.in",
     cmakelists = ["CMakeLists.txt"],
     defines = {
@@ -49,7 +56,7 @@ sources = glob(
 )
 
 gz_include_header(
-    name = "physics_hh_genrule",
+    name = "physicshh_genrule",
     out = "include/gz/physics.hh",
     hdrs = public_headers_no_gen + [
         "include/gz/physics/config.hh",
@@ -83,10 +90,9 @@ cc_library(
 
 cc_library(
     name = "heightmap",
-    hdrs = [
-        "heightmap/include/gz/physics/heightmap/HeightmapShape.hh",
-        "heightmap/include/gz/physics/heightmap/detail/HeightmapShape.hh",
-    ],
+    hdrs = glob([
+        "heightmap/include/gz/physics/heightmap/**/*.hh",
+    ]),
     includes = ["heightmap/include"],
 )
 
@@ -110,9 +116,7 @@ cc_library(
     ],
 )
 
-test_sources = glob(
-    include = ["src/*_TEST.cc"],
-)
+test_sources = glob(["src/*_TEST.cc"])
 
 [cc_test(
     name = src.replace("/", "_").replace(".cc", "").replace("src_", ""),
@@ -123,7 +127,7 @@ test_sources = glob(
     },
     deps = [
         ":physics",
-        GZ_ROOT + "physics/test:test_headers",
+        GZ_ROOT + "physics/test:common_test",
         "@gtest",
         "@gtest//:gtest_main",
     ],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,130 @@
+load(
+    "@gz//bazel/skylark:build_defs.bzl",
+    "GZ_FEATURES",
+    "GZ_ROOT",
+    "GZ_VISIBILITY",
+    "gz_configure_header",
+    "gz_export_header",
+    "gz_include_header",
+)
+
+package(
+    default_visibility = GZ_VISIBILITY,
+    features = GZ_FEATURES,
+)
+
+licenses(["notice"])  # Apache-2.0
+
+exports_files(["LICENSE"])
+
+gz_configure_header(
+    name = "physics_config_hh",
+    src = "include/gz/physics/config.hh.in",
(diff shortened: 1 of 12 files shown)


dartsim/src/plugin.cc
```diff
@@ -62,6 +62,7 @@ class Plugin :
     public virtual SimulationFeatures,
     public virtual WorldFeatures { };
 
+#ifndef GZ_PHYSICS_BAZEL_BUILD
```
mjcarroll, line 65: FYI @azeey this is what breaks when dartsim is statically linked via bazel.  Removing this section causes the cashes to go away.

mjcarroll:
Tested here: https://github.com/gazebosim/gz-bazel/pull/70

dartsim/src/plugin.cc
scpeters: 💸

src/InstallationDirectories.cc
```diff
@@ -28,27 +28,28 @@ namespace gz
 {
 namespace physics
 {
-inline namespace GZ_PHYSICS_VERSION_NAMESPACE {
```
scpeters, line 31: given that this is still in the header file, is it ok to change it here?

https://github.com/gazebosim/gz-physics/blob/mjcarroll/garden_bazel/include/gz/physics/InstallationDirectories.hh#L30

mjcarroll: I just moved it a little further down (https://github.com/gazebosim/gz-physics/blob/ddcd72b79b6341bf69d6f27da97ccca7b840a10f/src/InstallationDirectories.cc#L149), and put the rest in an anonymous namespace since it's not part of the public API

scpeters: 👍

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+    "gz_export_header",
+    "gz_include_header",
+)
+
+package(
+    default_visibility = GZ_VISIBILITY,
+    features = GZ_FEATURES,
+)
+
+licenses(["notice"])  # Apache-2.0
```
shameekganguly, line 16: I don't think the comment is necessary here (has to be stripped in copybara).

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+    "GZ_FEATURES",
+    "GZ_ROOT",
+    "GZ_VISIBILITY",
+    "gz_configure_header",
+    "gz_export_header",
+    "gz_include_header",
+)
+
+package(
+    default_visibility = GZ_VISIBILITY,
```
shameekganguly, line 12: Consider adding 
`default_applicable_licenses = [GZ_ROOT + "physics:license"],` similar to https://github.com/gazebosim/gz-utils/blob/gz-utils2/BUILD.bazel

BUILD.bazel
shameekganguly: Also, consider adding `package_name = "gz-physics"`

BUILD.bazel
shameekganguly: You'd also add
```
license(
    name = "license",
    package_name = "gz-physics",
)
```

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+    features = GZ_FEATURES,
+)
+
+licenses(["notice"])  # Apache-2.0
+
+exports_files(["LICENSE"])
+
+gz_configure_header(
+    name = "physics_config_hh",
```
shameekganguly, line 21: Consider simplifying rule name to just "config"

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+    name = "include/gz/physics/Export.hh",
+    export_base = "GZ_PHYSICS",
+    lib_name = "gz-physics",
+    visibility = ["//visibility:private"],
+)
+
+public_headers_no_gen = glob([
+    "include/gz/physics/*.hh",
+    "include/gz/physics/detail/*.hh",
+])
```
shameekganguly, line 42: In Garden, we probably want to include the `include/ignition` headers as well?

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+        "sdf/include/gz/physics/sdf/*.hh",
+    ]),
+    includes = ["sdf/include"],
+    deps = [
+        GZ_ROOT + "sdformat",
+    ],
+)
+
+test_sources = glob(
+    include = ["src/*_TEST.cc"],
```
shameekganguly, line 114: nit: we can drop the `include` argument name to be consistent with other uses of glob in the file

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+
+sources = glob(
+    ["src/*.cc"],
+    exclude = [
+        "src/*_TEST.cc",
+    ],
+)
+
+gz_include_header(
+    name = "physics_hh_genrule",
```
shameekganguly, line 52: nit: do we want to use `physics_hh_genrule` or `physicshh_genrule`? We're using `utilshh_genrule` in https://github.com/gazebosim/gz-utils/blob/gz-utils2/BUILD.bazel

BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+        "GZ_BAZEL": "1",
+        "GZ_BAZEL_PATH": "physics",
+    },
+    deps = [
+        ":physics",
+        GZ_ROOT + "physics/test:test_headers",
+        "@gtest",
+        "@gtest//:gtest_main",
+    ],
+) for src in test_sources]
```
shameekganguly, line 130: nit: consider exporting `resources`

bullet-featherstone/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
+load(
+    "@gz//bazel/skylark:build_defs.bzl",
+    "GZ_FEATURES",
```
shameekganguly, line 3: GZ_FEATURES seems to be unused. Did you mean to add a package rule?
```
package(
    default_visibility = GZ_VISIBILITY,
    features = GZ_FEATURES,
)
```

shameekganguly (commented):
Thanks!

BUILD.bazel
shameekganguly: Nope, lets exclude them :)

dartsim/src/plugin.cc
azeey: I found that loading the plugins with `RTLD_NODELETE` and removing this section fixes the issue on macOS and Linux. We do that for Gazebo plugins already, so I think we should do the same for physics engines.

dartsim/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+    srcs = [
+        "src/plugin.cc",
+    ],
+    linkshared = True,
+    visibility = GZ_VISIBILITY,
+    deps = [
+        ":dartsim",
+        GZ_ROOT + "plugin:register",
+    ],
+)
```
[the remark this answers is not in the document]
mjcarroll: This ended up being a bit more complicated than this loop as different tests have different linking requirements for using the dartsim plugin without any static initialization errors. PTAL

mjcarroll:

dartsim/src/plugin.cc
```diff
@@ -62,31 +62,6 @@ class Plugin :
...
-struct UnregisterCollisionDetectors
-{
-  ~UnregisterCollisionDetectors()
-  {
-    dart::collision::CollisionDetector::getFactory()->unregisterAllCreators();
-  }
-};
-
-UnregisterCollisionDetectors unregisterAtUnload;
-}
```
scpeters, line 88: I think this removal is the source of the macOS test failures

azeey:
@azeey to take a look at macOS failures.

azeey:
Closing since Garden is EOL.

PR #600, opened 2024-02-23
bullet-featherstone: Improve mesh collision stability

iche033:
# 🦟 Bug fix

## Summary

Mesh collisions are quite unstable in our bullet-featherstone implementation. Sometimes they explode and fly away on contact. The bullet-featherstone's multibody constraint solve applies a large impulse (uncapped) when penetration occurs.  To reduce this effect, we tune 2 parameters:
* collision margin - increased from 0.001 to 0.01
    * 0.01 is actually the [default margin](https://github.com/bulletphysics/bullet3/blob/6bb8d1123d8a55d407b19fd3357c724d0f5c9d3c/src/BulletCollision/Gimpact/btGImpactShape.h#L530) for GImpact meshes. The new value is also consistent with gazebo-classic which does not have margin set for GImpact meshes (so uses default value of 0.01)
* [erp2](https://github.com/bulletphysics/bullet3/blob/6bb8d1123d8a55d407b19fd3357c724d0f5c9d3c/src/BulletDynamics/ConstraintSolver/btContactSolverInfo.h#L48) (contact constraint error reduction param) - Reduces from default value of 0.2 to 0.02. This [affects](https://github.com/bulletphysics/bullet3/blob/6bb8d1123d8a55d407b19fd3357c724d0f5c9d3c/src/BulletDynamics/Featherstone/btMultiBodyConstraintSolver.cpp#L803) the contact penetration impulse applied.

Here's original behavior:
erp = 0.2, margin = 0.001
Box meshes explode on contact

bullet-1mm

Here's testing with just increasing the margin and leaving the erp untouched:
erp = 0.2, margin = 0.01
Stability improved but there are still small bounces due to penetration

bullet-1cm

Here's the new behavior with improved stability even when the top box is dropped from higher z:
erp = 0.002, margin = 0.01

bullet-002erp-1cm

The downside of having a larger margin is that the mesh collision has a small gap when resting on top of one another.
bullet-1cm_margin

This will need to be addressed

## To test

To reproduce the tests above, first you'll need these 2 PRs that allow bullet-featherstone in gz-phsyics to load meshes from relative paths:
* https://github.com/gazebosim/gz-sim/pull/2323
* https://github.com/gazebosim/sdformat/pull/1373
 
You can download the world and model files for testing:

```
mkdir -p box_mesh_test/box
wget https://gist.githubusercontent.com/iche033/9943a9bcf6ed9eede40ee3e9dbb4f8ad/raw/2f837c5221c3d56ee13f6191de4c0a449c36b0db/box_mesh_test.sdf
wget https://gist.githubusercontent.com/iche033/d13a8d8e0c05308f98ce1e5e105500ce/raw/681a2be4a2d53d36c9f0e415640258674ef977e5/model.sdf -O box_mesh_test/box/model.sdf
wget https://gist.githubusercontent.com/iche033/8c57b8882085dbce615419d5d472530c/raw/754f620ab034d218c0cda02453c4664968fe5935/box.obj -O box_mesh_test/box/box.obj
wget https://gist.githubusercontent.com/iche033/e59504e1b67f37f2ad30588d16fadb1b/raw/9a4dd30007345f017ca44f7da671eca6cddb9843/model.config -O box_mesh_test/box/model.config
```

Run gz-sim with bullet-featherstone plugin:

```
gz sim -v 4 box_mesh_test.sdf --physics-engine gz-physics-bullet-featherstone-plugin
```

## Checklist
- [ ] Added tests

### DIFF
--- a/bullet-featherstone/src/Base.cc
+++ b/bullet-featherstone/src/Base.cc
@@ -43,9 +43,9 @@ WorldInfo::WorldInfo(std::string name_)
   this->world->getSolverInfo().m_jointFeedbackInJointFrame = true;
   this->world->getSolverInfo().m_jointFeedbackInWorldSpace = false;
 
-  // By default a large impulse is is applied when collisions penetrate
+  // By default a large impulse is applied when collisions penetrate
   // which causes unstable behavior. Bullet featherstone does not support
-  // configuring split impulse and penetration threhold parameters. Instead the
+  // configuring split impulse and penetration threshold parameters. Instead the
   // penentration impulse depends on the erp2 parameter so set to a small value
   // (default is 0.2).
   this->world->getSolverInfo().m_erp2 = 0.002;
--- a/bullet-featherstone/src/Base.cc
+++ b/bullet-featherstone/src/Base.cc
@@ -42,6 +42,13 @@ WorldInfo::WorldInfo(std::string name_)
   // Needed for force-torque sensor
   this->world->getSolverInfo().m_jointFeedbackInJointFrame = true;
   this->world->getSolverInfo().m_jointFeedbackInWorldSpace = false;
+
+  // By default a large impulse is is applied when collisions penetrate
+  // which causes unstable behavior. Bullet featherstone does not support
+  // configuring split impulse and penetration threhold parameters. Instead the
+  // penentration impulse depends on the erp2 parameter so set to a small value
+  // (default is 0.2).
+  this->world->getSolverInfo().m_erp2 = 0.002;
 }
 
 }  // namespace bullet_featherstone
--- a/test/common_test/joint_features.cc
+++ b/test/common_test/joint_features.cc
@@ -1138,7 +1138,7 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachDetach)
     }
     frameDataModel2Body = model2Body->FrameDataRelativeToWorld();
 
-    EXPECT_NEAR(0.0, frameDataModel2Body.linearVelocity.z(), 1e-3);
+    EXPECT_NEAR(0.0, frameDataModel2Body.linearVelocity.z(), 2e-3);
   }
 }
 
@@ -1266,9 +1266,9 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachMultiple)
           gz::math::eigen3::convert(frameDataModel2Body.linearVelocity);
       gz::math::Vector3d body3LinearVelocity =
           gz::math::eigen3::convert(frameDataModel3Body.linearVelocity);
-      EXPECT_NEAR(0.0, body1LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(0.0, body1LinearVelocity.Z(), 3e-3);
       EXPECT_NEAR(0.0, body2LinearVelocity.Z(), 1e-3);
-      EXPECT_NEAR(0.0, body3LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(0.0, body3LinearVelocity.Z(), 3e-3);
     }
 
     // Detach the joints. M1 and M3 should fall as there is now nothing stopping
@@ -1296,9 +1296,9 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachMultiple)
       gz::math::Vector3d body3LinearVelocity =
           gz::math::eigen3::convert(frameDataModel3Body.linearVelocity);
 
-      EXPECT_NEAR(dt * (numSteps) * -10, body1LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(dt * (numSteps) * -10, body1LinearVelocity.Z(), 3e-3);
       EXPECT_NEAR(0.0, body2LinearVelocity.Z(), 1e-3);
-      EXPECT_NEAR(dt * (numSteps) * -10, body3LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(dt * (numSteps) * -10, body3LinearVelocity.Z(), 3e-3);
     }
   }
 }
--- a/test/common_test/joint_features.cc
+++ b/test/common_test/joint_features.cc
@@ -1132,13 +1132,22 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachDetach)
     }
 
     // After a while, body2 should reach the ground and come to a stop
-    for (std::size_t i = 0; i < 1000; ++i)
+    std::size_t stepCount = 0u;
+    const std::size_t maxNumSteps = 1000u;
+    while (stepCount++ < maxNumSteps)
     {
       world->Step(output, state, input);
+      frameDataModel2Body = model2Body->FrameDataRelativeToWorld();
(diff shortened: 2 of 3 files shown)


iche033:
the reduced `erp2` values caused a few vel checks in `joint_features` to fail and I had to increase tolerance for the tests to pass.  [bae4277](https://github.com/gazebosim/gz-physics/pull/600/commits/bae42771f496529ee68c6b712606f72308eea99e). It seems like the small `erp2` value causes objects to be slight more unstable when two objects are attached by a joint.

bullet-featherstone/src/Base.cc
```diff
@@ -42,6 +42,13 @@ WorldInfo::WorldInfo(std::string name_)
   // Needed for force-torque sensor
   this->world->getSolverInfo().m_jointFeedbackInJointFrame = true;
   this->world->getSolverInfo().m_jointFeedbackInWorldSpace = false;
+
+  // By default a large impulse is is applied when collisions penetrate
+  // which causes unstable behavior. Bullet featherstone does not support
+  // configuring split impulse and penetration threhold parameters. Instead the
```
shameekganguly, line 48: nit: threhold -> threshold

test/common_test/joint_features.cc
```diff
@@ -1138,7 +1138,7 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachDetach)
     }
     frameDataModel2Body = model2Body->FrameDataRelativeToWorld();
 
-    EXPECT_NEAR(0.0, frameDataModel2Body.linearVelocity.z(), 1e-3);
+    EXPECT_NEAR(0.0, frameDataModel2Body.linearVelocity.z(), 2e-3);
```
shameekganguly, line 1141: nit: this check seems a bit flaky, if we update erp2 in future, we'll have to tune this threshold again. Given that this test has nothing to do with collisions, I'd recommend checking for a combination of position and velocity thresholds instead inside the loop above, i.e.

```
while (true)
{
  world->Step(output, state, input);
  frameDataModel2Body = model2Body->FrameDataRelativeToWorld();
  // Expected Z height of model2 is 0.75 when both boxes are stacked on top of each other since each is 0.5 high.
  if (fabs(frameDataModel2Body.pose.z() - 0.75) < 2e-2 && fabs(frameDataModel2Body.linearVelocity.z()) < 1e-3)
  {
    break;
  }
}
```
Ideally, the position threshold of 1e-2 above can be set based on the configured margin for both objects, but it should be ok to use a large value.

test/common_test/joint_features.cc
```diff
@@ -1266,9 +1266,9 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachMultiple)
           gz::math::eigen3::convert(frameDataModel2Body.linearVelocity);
       gz::math::Vector3d body3LinearVelocity =
           gz::math::eigen3::convert(frameDataModel3Body.linearVelocity);
-      EXPECT_NEAR(0.0, body1LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(0.0, body1LinearVelocity.Z(), 3e-3);
```
shameekganguly, line 1269: nit: I'm slightly surprised that we need to increase the LinearVelocity.Z() thresholds for both body1 and body3 but not body2 with this change. I guess this is due to the weld joint constraint being imperfect in bullet? You could consider using a position and velocity based threshold here as well as I mentioned in my previous comment to prevent test flakiness.

test/common_test/joint_features.cc
```diff
@@ -1296,9 +1296,9 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachMultiple)
       gz::math::Vector3d body3LinearVelocity =
           gz::math::eigen3::convert(frameDataModel3Body.linearVelocity);
 
-      EXPECT_NEAR(dt * (numSteps) * -10, body1LinearVelocity.Z(), 1e-3);
+      EXPECT_NEAR(dt * (numSteps) * -10, body1LinearVelocity.Z(), 3e-3);
```
shameekganguly, line 1299: nit: you can make these checks relative to the velocity before stepping to make it more robust:
```
    frameDataModel1Body = model1Body->FrameDataRelativeToWorld();
    const double preStepBody1LinearVelocityZ = frameDataModel1Body.linearVelocity.z();
    ...
    for (std::size_t i = 0; i < numSteps; ++i)
    {
      world->Step(output, state, input);
    }
    {
      frameDataModel1Body = model1Body->FrameDataRelativeToWorld();
      ...
      // Expect the middle box to be still as it is already at rest.
      // Expect the two side boxes to fall away.

      EXPECT_NEAR(preStepBody1LinearVelocityZ + dt * (numSteps) * -10, frameDataModel1Body.linearVelocity.z(), 1e-3);
```
That way, you should be able to keep using the original conservative threshold of 1e-3 for this check.

test/common_test/joint_features.cc
```diff
@@ -1296,9 +1296,9 @@ TYPED_TEST(JointFeaturesAttachDetachTest, JointAttachMultiple)
       gz::math::Vector3d body3LinearVelocity =
```
shameekganguly, line 1296: nit: why do we need to convert to eigen types first? Doesn't seem like we are using a custom matcher in any case. Would it be enough to check with `frameDataModel1Body.linearVelocity.z()` instead of `body1LinearVelocity.Z()`?

shameekganguly (approved):

test/common_test/joint_features.cc
iche033: thanks for the suggestions! I updated to use while loop with max step count to make sure the loop eventually exits. [8ea3e67](https://github.com/gazebosim/gz-physics/pull/600/commits/8ea3e67e90177a28a5e6ed31da02dbec7bb93415)

bullet-featherstone/src/Base.cc
iche033: fixed. [8ea3e67](https://github.com/gazebosim/gz-physics/pull/600/commits/8ea3e67e90177a28a5e6ed31da02dbec7bb93415)

test/common_test/joint_features.cc
iche033: updated checks to take into account the pre-step body velocity. [8ea3e67](https://github.com/gazebosim/gz-physics/pull/600/commits/8ea3e67e90177a28a5e6ed31da02dbec7bb93415)

test/common_test/joint_features.cc
iche033: we shouldn't need to. I removed these conversions. [8ea3e67](https://github.com/gazebosim/gz-physics/pull/600/commits/8ea3e67e90177a28a5e6ed31da02dbec7bb93415)

test/common_test/joint_features.cc
iche033: yeah I'm not sure why the constraints are more unstable. I tried using the while loop approach with position and vel thresholds but the checks are always satisfied (and loop exists) on the 1st iteration right after the joints are created since this is when the objects are most stable. Given that the idea is to check that the objects are still at rest and stable after a few iterations, I reverted and kept the test the same as before. I just added position checks to make sure the boxes didn't move.

PR #603, opened 2024-03-13
bullet-featherstone: Support convex decomposition for meshes

iche033:
# 🎉 New feature

Depends on 
* https://github.com/gazebosim/sdformat/pull/1380
* https://github.com/gazebosim/gz-common/pull/583

## Summary

Supports convex decomposition on meshes. Bullet-featherstone implementation will parse the new mesh simplification attribute introduced in https://github.com/gazebosim/sdformat/pull/1380, decompses the mesh into convex meshes, and builds `btConvexHullShape` collision shapes.

Compared to `btGImpactMeshes` (which is currently used for all meshes), the convex hulls seems to be more stable, do not have gaps between meshes (collision margins can be set to 1mm instead of 1cm), and some manual testing shows potentially faster performance (dependent on the number of submeshes generated)

Added test to verify that convex decomposition flag is parsed and valid collisions are generated.

Other changes:
* the `m_erp2` value is now only set when there are meshes in the scene. The param was reduced to improve stability in https://github.com/gazebosim/gz-physics/pull/600. I think we do not need to do it globally when there are no meshes in the world as the instability seems to only come from meshes. The value is also bumped up a little to prevent to much penetration - This will likely need more tuning.

## To Test

Together with https://github.com/gazebosim/gz-sim/pull/2331, you can visualize the decomposed collision meshes. 

Run gz sim with bullet-featherstone plugin and a world that has a model (e.g. [Cordless Drill Simplified](https://app.gazebosim.org/iche033/fuel/models/Cordless%20Drill%20Simplified)) that uses mesh decomposition:

```
gz sim -v 4 your_test_world.sdf --physics-engine gz-physics-bullet-featherstone-plugin
```

## Checklist
- [x] Added tests

### DIFF
--- a/bullet-featherstone/src/SDFFeatures.cc
+++ b/bullet-featherstone/src/SDFFeatures.cc
@@ -1034,6 +1034,14 @@ bool SDFFeatures::AddSdfCollision(
   }
   else if (const auto *meshSdf = geom->MeshShape())
   {
+    // By default a large impulse is applied when mesh collisions penetrate
+    // which causes unstable behavior. Bullet featherstone does not support
+    // configuring split impulse and penetration threshold parameters. Instead
+    // the penentration impulse depends on the erp2 parameter so set to a small
+    // value (default is 0.2).
+    auto *world = this->ReferenceInterface<WorldInfo>(model->world);
+    world->world->getSolverInfo().m_erp2 = btScalar(0.02);
+
     auto &meshManager = *gz::common::MeshManager::Instance();
     auto *mesh = meshManager.Load(meshSdf->Uri());
     const btVector3 scale = convertVec(meshSdf->Scale());
--- a/bullet-featherstone/src/SDFFeatures.cc
+++ b/bullet-featherstone/src/SDFFeatures.cc
@@ -1054,86 +1054,45 @@ bool SDFFeatures::AddSdfCollision(
       bool meshCreated = false;
       if (meshSdf->Simplification() == "convex_decomposition")
       {
-        auto decomposed = meshManager.ConvexDecomposition(s.get());
+        std::vector<common::SubMesh> decomposed;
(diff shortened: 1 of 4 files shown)


bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -1051,34 +1059,77 @@ bool SDFFeatures::AddSdfCollision(
          ++submeshIdx)
     {
       auto s = mesh->SubMeshByIndex(submeshIdx).lock();
-      auto vertexCount = s->VertexCount();
-      auto indexCount = s->IndexCount();
-      btAlignedObjectArray<btVector3> convertedVerts;
-      convertedVerts.reserve(static_cast<int>(vertexCount));
-      for (unsigned int i = 0; i < vertexCount; i++)
+      bool meshCreated = false;
+      if (meshSdf->Simplification() == "convex_decomposition")
```
scpeters, line 1063: would it be possible to also handle the `convex_hull` case in this pull request? I would think that would be simpler than the `convex_decomposition` case, but would still be useful in cases where users know that a mesh is already convex

iche033: yep I was going to add it as a followup PR but I'll bundle it together.

iche033: addressed in new PR https://github.com/gazebosim/gz-physics/pull/606 that targets main. Supported by telling MeshManager to generate 1 convex hull. 

Alternative is to feed all vertices to `btConvexHullShape` and let bullet do the convex hull generation (intead of MeshManager) but that mean it'll be difficult for gz-sim visualize the generated collision. So I opted to let MeshManager create the convex hull instead so we can use the same function in the gui plugin to generate the collision visuals

iche033:
replaced by #606  - > `main`

PR #606, opened 2024-03-14
bullet-featherstone: Support convex decomposition for meshes

iche033:
# 🎉 New feature

Replaces #603 

Depends on 
* https://github.com/gazebosim/sdformat/pull/1382
* https://github.com/gazebosim/sdformat/pull/1386
* https://github.com/gazebosim/gz-common/pull/585
* https://github.com/gazebosim/gz-common/pull/588

## Summary

Supports convex decomposition on meshes. Bullet-featherstone implementation will parse the new mesh ~simplification~ optimization attribute introduced in https://github.com/gazebosim/sdformat/pull/1380, decompses the mesh into convex meshes, and builds `btConvexHullShape` collision shapes.

Compared to `btGImpactMeshes` (which is currently used for all meshes), the convex hulls seems to be more stable, do not have gaps between meshes (collision margins can be set to 1mm instead of 1cm), and some manual testing shows potentially faster performance (dependent on the number of submeshes generated)

Added test to verify that convex decomposition flag is parsed and valid collisions are generated.

Other changes:
* the `m_erp2` ~value is now only set when there are meshes in the scene. The param was reduced to improve stability in https://github.com/gazebosim/gz-physics/pull/600. I think we do not need to do it globally when there are no meshes in the world as the instability seems to only come from meshes.~ The value is also bumped up a little too prevent to much penetration - This will likely need more tuning as we do more testing with bullet.

## To Test

Run gz sim with bullet-featherstone plugin and a world that has a model (e.g. [Cordless Drill Simplified](https://app.gazebosim.org/iche033/fuel/models/Cordless%20Drill%20Simplified)) that uses mesh decomposition:

```
gz sim -v 4 your_test_world.sdf --physics-engine gz-physics-bullet-featherstone-plugin
```

## Checklist
- [x] Added tests


bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -1051,34 +1051,92 @@ bool SDFFeatures::AddSdfCollision(
...
+        }
+        else if (meshSdf->ConvexDecomposition())
+        {
+          // limit max number of convex hulls to generate
+          maxConvexHulls = meshSdf->ConvexDecomposition()->MaxConvexHulls();
+        }
+        std::vector<common::SubMesh> decomposed =
+          std::move(meshManager.ConvexDecomposition(*s.get(), maxConvexHulls));
+
+        gzdbg << "Optimizing mesh using convex decomposition. " << std::endl;
```
scpeters, line 1074: we may want to remove this, but it has been very helpful for testing

bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -1051,34 +1051,92 @@ bool SDFFeatures::AddSdfCollision(
...
-      auto vertexCount = s->VertexCount();
-      auto indexCount = s->IndexCount();
-      btAlignedObjectArray<btVector3> convertedVerts;
-      convertedVerts.reserve(static_cast<int>(vertexCount));
-      for (unsigned int i = 0; i < vertexCount; i++)
+      bool meshCreated = false;
+      if (meshSdf->Optimization() ==
+          ::sdf::MeshOptimization::CONVEX_DECOMPOSITION ||
+          meshSdf->Optimization() ==
+          ::sdf::MeshOptimization::CONVEX_HULL)
```
scpeters, line 1058: consider moving this logic outside of the SubMesh `for` loop

test/common_test/collisions.cc
```diff
@@ -141,6 +146,106 @@ TYPED_TEST(CollisionTest, MeshAndPlane)
...
+  // Load an optimized mesh, drop it from some height,
+  // and verify it collides with the ground plane
+  std::string modelOptimizedStr = R"(
+  <sdf version="1.11">
+    <model name="model_optimized">
+      <pose>0 0 2.0 0 0 0</pose>
+      <link name="body">
+        <collision name="collision">
+          <geometry>
+            <mesh optimization="convex_decomposition">
```
scpeters, line 176: this test could be generalized to include other values of `//mesh/@optimization` (both `convex_hull` and the default `""` case)

bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -1051,34 +1051,92 @@ bool SDFFeatures::AddSdfCollision(
...
+          const btVector3& v2 = convertedVerts[s->Index(i*3 + 2)];
+          this->triangleMeshes.back()->addTriangle(v0, v1, v2);
+        }
+
+        this->meshesGImpact.push_back(
+          std::make_unique<btGImpactMeshShape>(
+            this->triangleMeshes.back().get()));
+        this->meshesGImpact.back()->updateBound();
+        this->meshesGImpact.back()->setMargin(btScalar(0.01));
+        compoundShape->addChildShape(btTransform::getIdentity(),
```
scpeters, line 1137: this currently uses `btCompoundShape` to join submeshes

scpeters (commented):
I was tinkering with this by changing the `//mesh/@optimization` value in the test to both an empty string and `convex_hull` to for comparison's sake, and I realized from the console output that we are computing convex decompositions of all the submeshes and then adding them all to a `btCompoundShape`. For comparison, when using bullet collisions in dartsim (which stores mesh information with submeshes in an Assimp Scene), all the submeshes are merged together into a single mesh:

* https://github.com/dartsim/dart/blob/f33fd04448141554eab8f4de378e1d40c1d0d212/dart/collision/bullet/BulletCollisionDetector.cpp#L879-L911

I think `convex_decomposition` and `convex_hull` operations should apply to the fusion of all submeshes, unless only a single submesh is specified, in which case it would act only on that submesh. In the unoptimized case, we could also consider merging to a single mesh shape instead of using `btCompoundShape` (this can happen in separate pull requests)

iche033:
> I think convex_decomposition and convex_hull operations should apply to the fusion of all submeshes, unless only a single submesh is specified, in which case it would act only on that submesh. In the unoptimized case, we could also consider merging to a single mesh shape instead of using btCompoundShape (this can happen in separate pull requests)

hmm ok to make this happen, we'll need to:
* update [sdf description](https://github.com/gazebosim/sdformat/blob/main/sdf/1.11/mesh_shape.sdf#L13) - currently it explicitly states decomposition is for each submesh
* add a new function to merge multiple `common::SubMesh`es into one (before calling `gz::common::ConvexDecomposition` with that one combined submesh)

I'll work on these changes

bullet-featherstone/src/SDFFeatures.cc
iche033: moved out of `for` loop. [1f2aaa7](https://github.com/gazebosim/gz-physics/pull/606/commits/1f2aaa7a24756c51f56b5d03ad56958926d38630)

bullet-featherstone/src/SDFFeatures.cc
iche033: shortened the debug msg to one line. [1f2aaa7](https://github.com/gazebosim/gz-physics/pull/606/commits/1f2aaa7a24756c51f56b5d03ad56958926d38630)

I can also remove this msg completely if preferred.

test/common_test/collisions.cc
iche033: generalized test to include other optimization methods. [1f2aaa7](https://github.com/gazebosim/gz-physics/pull/606/commits/1f2aaa7a24756c51f56b5d03ad56958926d38630)

iche033:
> hmm ok to make this happen, we'll need to:
>
> * update [sdf description](https://github.com/gazebosim/sdformat/blob/main/sdf/1.11/mesh_shape.sdf#L13) - currently it explicitly states decomposition is for each submesh
> * add a new function to merge multiple common::SubMeshes into one (before calling gz::common::ConvexDecomposition with that one combined submesh)
>
> I'll work on these changes

https://github.com/gazebosim/sdformat/pull/1386
https://github.com/gazebosim/gz-common/pull/588

bullet-featherstone/src/SDFFeatures.cc
iche033: Updated to merge all submeses before decomposition. 

Note that the decomposed submeshes are then still joined together using `btCompoundShape`. Same for the unoptimized case. I can change this to using a single mesh in a follow-up PR.

test/common_test/collisions.cc
```diff
@@ -141,6 +147,133 @@ TYPED_TEST(CollisionTest, MeshAndPlane)
...
+      frameDataModelOptimizedBody =
+          modelOptimizedBody->FrameDataRelativeToWorld();
+
+      // convex decomposition gives more accurate results
+      double tol = (optimizationStr == "convex_decomposition") ? 1e-3 : 1e-2;
+      EXPECT_NEAR(0.1,
+                  frameDataModelOptimizedBody.pose.translation().z(), tol);
+      EXPECT_NEAR(0.0, frameDataModelOptimizedBody.linearVelocity.z(), tol);
+
+      initialModelPose.Pos() += gz::math::Vector3d(0, 2, 0);
```
azeey, line 272: In general, I thought testing with a V or U shaped object would really show the difference in behavior between `convex_hull` and `convex_decomposition`, but
I'm getting some unexpected penetration with `convex_decomposition`. I've attached the .sdf files.
image

Maybe we can use this in the test?

[test_conv_decomp.zip](https://github.com/gazebosim/gz-physics/files/14928199/test_conv_decomp.zip)

azeey (commented):
The code LGTM! but I did see a weird behavior while testing.

bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -1036,49 +1036,140 @@ bool SDFFeatures::AddSdfCollision(
...
-        const btVector3& v0 = convertedVerts[s->Index(i*3)];
-        const btVector3& v1 = convertedVerts[s->Index(i*3 + 1)];
-        const btVector3& v2 = convertedVerts[s->Index(i*3 + 2)];
-        this->triangleMeshes.back()->addTriangle(v0, v1, v2);
+        // Merge meshes before convex decomposition
+        auto mergedMesh = gz::common::MeshManager::MergeSubMeshes(*mesh);
+        if (mergedMesh && mergedMesh->SubMeshCount() == 1u)
+        {
+          // Decompose and add mesh to MeshManager
+          auto s = mergedMesh->SubMeshByIndex(0u).lock();
```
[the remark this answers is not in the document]
iche033: renamed to `mergedSubmesh`.  [2099054](https://github.com/gazebosim/gz-physics/pull/606/commits/2099054eef9a92bf473f595042954f1d99593af6)

test/common_test/collisions.cc
```diff
@@ -141,6 +147,133 @@ TYPED_TEST(CollisionTest, MeshAndPlane)
...
+    gz::math::Pose3d initialModelPose(0, 0, 2, 0, 0, 0);
+    // Test all optimization methods
+    for (const auto &optimizationStr : optimizations)
+    {
+      // create the chassis model
+      const std::string modelOptimizedName = "model_optimized_" + optimizationStr;
+      sdf::Root root;
+      sdf::Errors errors = root.LoadSdfString(getModelOptimizedStr(
+          optimizationStr, modelOptimizedName, initialModelPose));
+      ASSERT_TRUE(errors.empty()) << errors.front();
```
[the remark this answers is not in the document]
iche033: updated. [2099054](https://github.com/gazebosim/gz-physics/pull/606/commits/2099054eef9a92bf473f595042954f1d99593af6)

test/common_test/collisions.cc
iche033: I added an integration test for this. In the test I set `<max_convex_hulls>` to 64 which prevents this penetration. [2099054](https://github.com/gazebosim/gz-physics/pull/606/commits/2099054eef9a92bf473f595042954f1d99593af6)

The generated convex hulls looked fine with either 16 or 64 `max_convex_hulls` so something else is not right. This is something to look into next.

iche033:
> I'm getting some unexpected penetration with convex_decomposition. I've attached the .sdf files.

I found that by setting a small `m_globalCfm` value, it's able to prevent this penetration, e.g.

```
diff --git a/bullet-featherstone/src/Base.cc b/bullet-featherstone/src/Base.cc
index ac66f574..80a04ced 100644
--- a/bullet-featherstone/src/Base.cc
+++ b/bullet-featherstone/src/Base.cc
@@ -49,6 +49,7 @@ WorldInfo::WorldInfo(std::string name_)
   // the penentration impulse depends on the erp2 parameter so set to a small
   // value (default in bullet is 0.2).
   this->world->getSolverInfo().m_erp2 = btScalar(0.02);
+  this->world->getSolverInfo().m_globalCfm = btScalar(0.001);
```

by default `m_globalCfm` is 0

PR #608, opened 2024-03-18
Add package.xml

azeey:
# 🎉 New feature

## Summary

Add `package.xml` based on dependencies found in `.github/ci/packages`. 

This is to test out using vendor packages to provide Gazebo packages to ROS users (see https://github.com/gazebo-tooling/release-tools/issues/1117).


package.xml
```diff
@@ -0,0 +1,27 @@
...
+  <name>gz-physics7</name>
+  <version>7.2.0</version>
+  <description>Gazebo Physics : Physics classes and functions for robot applications</description>
+  <maintainer email="scpeters@openrobotics.org">Steve Peters</maintainer>
+  <license>Apache License 2.0</license>
+  <url type="website">https://github.com/gazebosim/gz-physics</url>
+
+  <buildtool_depend>cmake</buildtool_depend>
+
+  <build_depend>gz-cmake</build_depend>
```
scpeters, line 12:
```suggestion
  <build_depend>gz-cmake3</build_depend>
```

package.xml
```diff
@@ -0,0 +1,27 @@
...
+
+  <depend>benchmark</depend>
+  <depend>bullet</depend>
+  <depend>dartsim</depend>
```
scpeters, line 16: I know dartsim uses this name in its `package.xml` file, but I had added a [colcon.pkg file](https://github.com/dartsim/dart/blob/main/colcon.pkg) to change its package name for colcon's purposes to `DART` so that `colcon-cmake` would recognize `find_package(DART)` (see https://github.com/dartsim/dart/pull/1389)

I haven't yet tried to build a workspace with gz-physics, dartsim, bullet and ODE, and we don't need to block on this, but I'm not sure building all those from source in the same workspace will work properly in this current form

azeey: You're right. I used `colcon graph` and saw that `gz-physics` didn't have a dependency on DART, so I've changed it in https://github.com/gazebosim/gz-physics/pull/608/commits/6fda9a7ffeb88a8b0f4e71179534ec0a08265244

**Before**
```
DART            +
gz-cmake3        +********* *****
gz-tools2         +  * ***..*****
gz-utils2          +***.******.*.
gz-math7            + ***********
gz-plugin2           +    ** *.**
gz-common5            +  *** ****
gz-msgs10              + *  *****
sdformat14              + *   ***
gz-fuel_tools9           +     *.
gz-physics7               +    **
gz-rendering8              + ****
gz-transport13              +****
gz-gui8                      + **
gz-sensors8                   +**
gz-sim8                        +*
gz-launch7                      +
```

**After**
```
DART            +         *    ..
gz-cmake3        +********* *****
gz-tools2         +  * ***..*****
gz-utils2          +***.******.*.
gz-math7            + ***********
gz-plugin2           +    ** *.**
gz-common5            +  *** ****
gz-msgs10              + *  *****
sdformat14              + *   ***
gz-fuel_tools9           +     *.
gz-physics7               +    **
gz-rendering8              + ****
gz-transport13              +****
gz-gui8                      + **
gz-sensors8                   +**
gz-sim8                        +*
gz-launch7                      +

 ```

package.xml
azeey: Fixed in https://github.com/gazebosim/gz-physics/pull/608/commits/df390df0f936bc5cacb5bfd15d52e49226763eb8

scpeters (approved):

PR #630, opened 2024-04-22
bullet-featurestore: Enable auto deactivation

iche033:
# 🎉 New feature

## Summary

bullet supports auto deactivation of bodies after certain period if there's no motion (velocity < some threshold). This feature should be enabled by default by bullet but it looks like we were overriding this by looping through all models and setting them to active after every step. I think this may have been done because free group / joint cmds were not working after objects were deactivated.

This PR removes that logic for activating all bodies after every step. Instead, an explicit call to wake up the bodies when setting free group / joint commands.

## Test it

The effect is most obvious with mesh-to-mesh collisions which are less stable. When 2 meshes collide and come to a rest, they should now stop drifting.

I tested with the box meshes from https://github.com/gazebosim/gz-physics/pull/600 (under To Test section). Expand the Component Inspector to see more decimal points, let the 2 box meshes collide. After a couple of seconds, the box pose values should settle and stop updating. Before these changes, the pose values would continue to drift over time.
 
bullet_auto_disable

## Checklist
- [ ] Added tests


ahcorde (approved):

iche033:
> Do we need to do the same for applying external forces to links? You can test this with the "Mouse Drag" GUI plugin in gz-sim.

oh yes, missed LinkFeatures. Added in [457578d](https://github.com/gazebosim/gz-physics/pull/630/commits/457578dbf18b879688324176c158597a5411b7ac).

PR #639, opened 2024-05-15
Add Cone as a collision shape

bperseghetti:
# 🦟 Bug fix

## Summary
This helps add the missing cone geometry for primitive/basic parametric shapes:

conetopple
cone

And is also valuable for visualizations of emitters/source that typically have conic-based spread as seen in this acoustic attack on an IMU by showing the affected area:

drone_attack

Associated PRs:
- https://github.com/gazebosim/gz-gui/pull/621
- https://github.com/gazebosim/gz-math/pull/594
- https://github.com/gazebosim/gz-msgs/pull/442
- https://github.com/gazebosim/gz-physics/pull/639
- https://github.com/gazebosim/gz-rendering/pull/1003
- https://github.com/gazebosim/gz-sim/pull/2410
- https://github.com/gazebosim/sdformat/pull/1418

## Checklist
- [x] Added tests
- [x] Updated documentation (as needed)
- [x] Updated migration guide (as needed)
- [x] Consider updating Python bindings (if the library has them)
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

### DIFF
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -20,6 +20,7 @@
 #include <cmath>
 #include <limits>
 #include <memory>
+#include <string>
 #include <utility>
 
 #include <dart/constraint/ConstraintSolver.hpp>
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -26,6 +26,7 @@
 #include <dart/dynamics/BallJoint.hpp>
 #include <dart/dynamics/BoxShape.hpp>
 #include <dart/dynamics/CapsuleShape.hpp>
+#include <dart/dynamics/ConeShape.hpp>
 #include <dart/dynamics/CylinderShape.hpp>
 #include <dart/dynamics/EllipsoidShape.hpp>
 #include <dart/dynamics/FreeJoint.hpp>
@@ -49,6 +50,7 @@
 #include <sdf/Box.hh>
 #include <sdf/Collision.hh>
 #include <sdf/Capsule.hh>
+#include <sdf/Cone.hh>
 #include <sdf/Cylinder.hh>
 #include <sdf/Ellipsoid.hh>
 #include <sdf/Geometry.hh>
@@ -257,6 +259,14 @@ static ShapeAndTransform ConstructBox(
 }
 
 /////////////////////////////////////////////////
+static ShapeAndTransform ConstructCone(
+    const ::sdf::Cone &_cone)
+{
+  return {std::make_shared<dart::dynamics::ConeShape>(
+        _cone.Radius(), _cone.Length())};
+}
+
+/////////////////////////////////////////////////
 static ShapeAndTransform ConstructCylinder(
     const ::sdf::Cylinder &_cylinder)
 {
@@ -340,8 +350,14 @@ static ShapeAndTransform ConstructGeometry(
   {
     return ConstructCapsule(*_geometry.CapsuleShape());
   }
+  else if (_geometry.ConeShape())
+  {
+    return ConstructCone(*_geometry.ConeShape());
+  }
   else if (_geometry.CylinderShape())
+  {
     return ConstructCylinder(*_geometry.CylinderShape());
+  }
   else if (_geometry.EllipsoidShape())
   {
     // TODO(anyone): Replace this code when Ellipsoid is supported by DART
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -259,14 +259,6 @@ static ShapeAndTransform ConstructBox(
 }
 
 /////////////////////////////////////////////////
-static ShapeAndTransform ConstructCone(
-    const ::sdf::Cone &_cone)
-{
-  return {std::make_shared<dart::dynamics::ConeShape>(
-        _cone.Radius(), _cone.Length())};
-}
-
-/////////////////////////////////////////////////
 static ShapeAndTransform ConstructCylinder(
     const ::sdf::Cylinder &_cylinder)
 {
@@ -352,7 +344,25 @@ static ShapeAndTransform ConstructGeometry(
   }
   else if (_geometry.ConeShape())
   {
-    return ConstructCone(*_geometry.ConeShape());
+    // TODO(anyone): Replace this code when Cone is supported by DART
+    gzwarn << "DART: Cone is not a supported collision geomerty"
+           << " primitive, using generated mesh of a cone instead"
+           << std::endl;
+    common::MeshManager *meshMgr = common::MeshManager::Instance();
+    std::string coneMeshName = std::string("cone_mesh")
+      + "_" + std::to_string(_geometry.ConeShape()->Radius())
+      + "_" + std::to_string(_geometry.ConeShape()->Length());
+    meshMgr->CreateCone(
+      coneMeshName,
+      _geometry.ConeShape()->Radius(),
+      _geometry.ConeShape()->Length(),
+      3, 40);
+    const gz::common::Mesh * _mesh =
+      meshMgr->MeshByName(coneMeshName);
+
+    auto mesh = std::make_shared<CustomMeshShape>(*_mesh, Vector3d(1, 1, 1));
+    auto mesh2 = std::dynamic_pointer_cast<dart::dynamics::MeshShape>(mesh);
+    return {mesh2};
   }
   else if (_geometry.CylinderShape())
   {
@@ -361,6 +371,9 @@ static ShapeAndTransform ConstructGeometry(
   else if (_geometry.EllipsoidShape())
   {
     // TODO(anyone): Replace this code when Ellipsoid is supported by DART
+    gzwarn << "DART: Ellipsoid is not a supported collision geomerty"
+           << " primitive, using generated mesh of an ellipsoid instead"
+           << std::endl;
     common::MeshManager *meshMgr = common::MeshManager::Instance();
     std::string ellipsoidMeshName = std::string("ellipsoid_mesh")
       + "_" + std::to_string(_geometry.EllipsoidShape()->Radii().X())
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -68,6 +68,7 @@
 #include <sdf/World.hh>
 
 #include "AddedMassFeatures.hh"
+#include "CustomConeMeshShape.hh"
 #include "CustomMeshShape.hh"
 
 namespace gz {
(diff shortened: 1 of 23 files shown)


dartsim/src/SDFFeatures.cc
```diff
@@ -340,11 +342,38 @@ static ShapeAndTransform ConstructGeometry(
...
     return ConstructCapsule(*_geometry.CapsuleShape());
   }
+  else if (_geometry.ConeShape())
+  {
+    // TODO(anyone): Replace this code when Cone is supported by DART
+    gzwarn << "DART: Cone is not a supported collision geomerty"
+           << " primitive, using generated mesh of a cone instead"
+           << std::endl;
+    common::MeshManager *meshMgr = common::MeshManager::Instance();
+    std::string coneMeshName = std::string("cone_mesh")
```
ahcorde, line 352: include `<string>`

dartsim/src/ShapeFeatures.cc
```diff
@@ -161,6 +162,76 @@ Identity ShapeFeatures::AttachCapsuleShape(
...
+      static_cast<dart::dynamics::ConeShape *>(
+          shapeInfo->node->getShape().get());
+
+  return cone->getHeight();
+}
+
+/////////////////////////////////////////////////
+Identity ShapeFeatures::AttachConeShape(
+    const Identity &_linkID,
+    const std::string &_name,
```
ahcorde, line 206: include `<string>`

dartsim/src/ShapeFeatures.cc
```diff
@@ -161,6 +162,76 @@ Identity ShapeFeatures::AttachCapsuleShape(
...
+  auto mesh = std::make_shared<CustomMeshShape>(*_mesh, Vector3d(1, 1, 1));
+
+  DartBodyNode *bn = this->ReferenceInterface<LinkInfo>(_linkID)->link.get();
+  dart::dynamics::ShapeNode *sn =
+      bn->createShapeNodeWith<dart::dynamics::CollisionAspect,
+                              dart::dynamics::DynamicsAspect>(
+          mesh, bn->getName() + ":" + _name);
+
+  sn->setRelativeTransform(_pose);
+  const std::size_t shapeID = this->AddShape({sn, _name});
```
ahcorde, line 231: include `<cstddef>`

ahcorde (dismissed):

dartsim/src/SDFFeatures.cc
```diff
@@ -340,11 +343,38 @@ static ShapeAndTransform ConstructGeometry(
...
+      + "_" + std::to_string(_geometry.ConeShape()->Radius())
+      + "_" + std::to_string(_geometry.ConeShape()->Length());
+    meshMgr->CreateCone(
+      coneMeshName,
+      _geometry.ConeShape()->Radius(),
+      _geometry.ConeShape()->Length(),
+      3, 40);
```
bperseghetti, line 360: @azeey @scpeters @ahcorde I set these values as they seemed to be "reasonable" tradeoff between increased computation and avoiding "lumpy" cones that can't roll. I think the problem really stems from the implementation in gz-common though, it would be nice if segments were actually a function of arc length or surface area so you get consistent segment sizes. IE the bounded area of each segment by the top vertex ring set is much much much smaller than the area bounded by the ringset adjacent to the bottom cap. I think maybe changing that here to be calculated from a segment area and number of rings instead of number of segments and number of rings would make a lot more sense: https://github.com/gazebosim/gz-common/blob/27f7017c5c1b1fd2ba9c603e92b694f98417175d/graphics/src/MeshManager.cc#L1242-L1313

azeey: Can the number of rings be set to 0? I would think the rolling issue is affected by the number of segments, but not rings. 

Regarding the idea of making segments be a function of the arc length, I think we can do that here---if the `rings` can be set to 0, `segments` could be a simple function of the radius.

bperseghetti: Correct, and even for visualization I was thinking it might not be bad to have less segments with the same logic but just implemented in the gz-common I showed earlier? I guess, is there ever a case where we would desire more than one ring? I would feel like having the input to the mesh function just be number of segments might be enough for all cases?

azeey: > is there ever a case where we would desire more than one ring? 

I can't think of one. The `unit_cone` which is created in  `gz::common::MeshManager` is used in `gz::rendering` for the cone in the transform tool. That's the only place I see it used. And for that, I don't think we need more than one ring (0 if you don't count the base circle). Maybe someone  That being said, is changing the API worth the hassle?

scpeters (commented):
I've updated some shape tests in https://github.com/gazebosim/gz-physics/commit/71d883eafcfaa843714b5c8f7d67a2d37f2a7c99 (feel free to cherry-pick) to include the cone (just by duplicating cylinder and search/replace `s/ylinder/one/`), and I'm seeing a test failure in dart related to casting to the ConeShape. I commented on the dartsim ShapeFeatures.cc file where I think the problem comes from.

Perhaps we could create a `CustomConeShape` that inherits from `CustomMeshShape`, so we can cast to `CustomConeShape` instead of `dart::dynamics::ConeShape`?

bperseghetti:
> I've updated some shape tests in [71d883e](https://github.com/gazebosim/gz-physics/commit/71d883eafcfaa843714b5c8f7d67a2d37f2a7c99) (feel free to cherry-pick) to include the cone (just by duplicating cylinder and search/replace `s/ylinder/one/`), and I'm seeing a test failure in dart related to casting to the ConeShape. I commented on the dartsim ShapeFeatures.cc file where I think the problem comes from.
> 
> Perhaps we could create a `CustomConeShape` that inherits from `CustomMeshShape`, so we can cast to `CustomConeShape` instead of `dart::dynamics::ConeShape`?

Cherry-picked and pushed.
You have a test implementation of that method?

scpeters:
> Cherry-picked and pushed. You have a test implementation of that method?

not yet; I was just tinkering with it and since the `CustomMeshShape` constructor wants a `gz::common::Mesh`, I think we would need a factory function to return a `gz::common::Mesh` to use in the `CustomMeshShape` initializer. I'll tinker a bit more

scpeters:
> > Cherry-picked and pushed. You have a test implementation of that method?
> 
> not yet; I was just tinkering with it and since the `CustomMeshShape` constructor wants a `gz::common::Mesh`, I think we would need a factory function to return a `gz::common::Mesh` to use in the `CustomMeshShape` initializer. I'll tinker a bit more

ok, I think I've gotten most of the tests fixed, except for bullet-featherstone, which complains about the bounding box size being larger due to the collision margin parameter, which we should discuss directly

* fix for tpe test: https://github.com/scpeters/ign-physics/commit/79e38ff610c3189067405ee4bece1edd1fb74e7e
* fix for dartsim tests by adding CustomConeMeshShape: https://github.com/scpeters/ign-physics/commit/2053f150c305444f99a74c1419e5eff7403f4a36

bperseghetti:
> > > Cherry-picked and pushed. You have a test implementation of that method?
> > 
> > 
> > not yet; I was just tinkering with it and since the `CustomMeshShape` constructor wants a `gz::common::Mesh`, I think we would need a factory function to return a `gz::common::Mesh` to use in the `CustomMeshShape` initializer. I'll tinker a bit more
> 
> ok, I think I've gotten most of the tests fixed, except for bullet-featherstone, which complains about the bounding box size being larger due to the collision margin parameter, which we should discuss directly
> 
> * fix for tpe test: [scpeters@79e38ff](https://github.com/scpeters/ign-physics/commit/79e38ff610c3189067405ee4bece1edd1fb74e7e)
> * fix for dartsim tests by adding CustomConeMeshShape: [scpeters@2053f15](https://github.com/scpeters/ign-physics/commit/2053f150c305444f99a74c1419e5eff7403f4a36)

Thanks, cherry-picked!

scpeters:
the following patch will fix tests by reducing the bullet-featherstone margin to 0, but I'm not sure if that will cause problems with contact simulation

~~~
diff --git a/bullet-featherstone/src/SDFFeatures.cc b/bullet-featherstone/src/SDFFeatures.cc
index 82fe6744..3ad3b3f5 100644
--- a/bullet-featherstone/src/SDFFeatures.cc
+++ b/bullet-featherstone/src/SDFFeatures.cc
@@ -970,6 +970,7 @@ bool SDFFeatures::AddSdfCollision(
     const auto height = static_cast<btScalar>(cone->Length());
     shape =
       std::make_unique<btConeShapeZ>(radius, height);
+    shape->setMargin(0.0);
   }
   else if (const auto *cylinder = geom->CylinderShape())
   {
diff --git a/bullet-featherstone/src/ShapeFeatures.cc b/bullet-featherstone/src/ShapeFeatures.cc
index c4702764..038ad835 100644
--- a/bullet-featherstone/src/ShapeFeatures.cc
+++ b/bullet-featherstone/src/ShapeFeatures.cc
@@ -247,6 +247,7 @@ Identity ShapeFeatures::AttachConeShape(
   const auto height = static_cast<btScalar>(_height);
   auto shape =
     std::make_unique<btConeShapeZ>(radius, height);
+  shape->setMargin(0.0);
 
   auto identity = this->AddCollision(
     CollisionInfo{
~~~

bperseghetti:
> the following patch will fix tests by reducing the bullet-featherstone margin to 0, but I'm not sure if that will cause problems with contact simulation
> 
> ```
> diff --git a/bullet-featherstone/src/SDFFeatures.cc b/bullet-featherstone/src/SDFFeatures.cc
> index 82fe6744..3ad3b3f5 100644
> --- a/bullet-featherstone/src/SDFFeatures.cc
> +++ b/bullet-featherstone/src/SDFFeatures.cc
> @@ -970,6 +970,7 @@ bool SDFFeatures::AddSdfCollision(
>      const auto height = static_cast<btScalar>(cone->Length());
>      shape =
>        std::make_unique<btConeShapeZ>(radius, height);
> +    shape->setMargin(0.0);
>    }
>    else if (const auto *cylinder = geom->CylinderShape())
>    {
> diff --git a/bullet-featherstone/src/ShapeFeatures.cc b/bullet-featherstone/src/ShapeFeatures.cc
> index c4702764..038ad835 100644
> --- a/bullet-featherstone/src/ShapeFeatures.cc
> +++ b/bullet-featherstone/src/ShapeFeatures.cc
> @@ -247,6 +247,7 @@ Identity ShapeFeatures::AttachConeShape(
>    const auto height = static_cast<btScalar>(_height);
>    auto shape =
>      std::make_unique<btConeShapeZ>(radius, height);
> +  shape->setMargin(0.0);
>  
>    auto identity = this->AddCollision(
>      CollisionInfo{
> ```

BTW, I do have your personal github as one of my remotes for that repo so you can always just push to the branch and I can fetch/cherry-pick anything in ;-P

bperseghetti:
@scpeters thanks once again, I applied your patch, I think it just needs CI approval.

dartsim/src/SDFFeatures_TEST.cc
```diff
@@ -972,10 +972,10 @@ TEST_P(SDFFeatures_TEST, Shapes)
   auto dartWorld = world->GetDartsimWorld();
   ASSERT_NE(nullptr, dartWorld);
 
-  ASSERT_EQ(5u, dartWorld->getNumSkeletons());
+  ASSERT_EQ(6u, dartWorld->getNumSkeletons());
 
   int count{0};
-  for (auto name : {"sphere", "box", "cylinder", "capsule", "ellipsoid"})
+  for (auto name : {"sphere", "box", "cylinder", "capsule", "ellipsoid", "cone"})
```
scpeters, line 978:
```suggestion
  for (auto name : {"sphere", "box", "cylinder", "capsule", "ellipsoid",
                    "cone"})
```
linter complains about line length from one of my patches

bperseghetti: @scpeters no worries, I also changed the number of overall segments in https://github.com/gazebosim/gz-physics/pull/639/commits/642692058a5bc51e2cc681caeb701197a9349ba3 to what we had discussed in https://github.com/gazebosim/gz-physics/pull/639#discussion_r1605419363

scpeters:
rebuilding nightly debs:

* https://build.osrfoundation.org/view/gz-ionic/job/gz-physics8-debbuilder/376/
* https://build.osrfoundation.org/view/gz-ionic/job/gz-physics8-debbuilder/377/

PR #647, opened 2024-06-03
bullet-featherstone: Fix bounding box for collisions with pose offset

iche033:
# 🦟 Bug fix

hide whitespace changes for better diff:
https://github.com/gazebosim/gz-physics/pull/647/files?w=1

## Summary

Currently if a collision has a pose offset, the bullet-featherstone plugin return an incorrect axis aligned bounding box for the link containing the collision. This is found to be caused by incorrect frame data pose computed for collisions - It currently ignores the collision pose offset. 

This PR makes a few changes to `KinematicsFeatures::FrameDataRelativeToWorld` in bullet-featherstone plugin:
* fixed getting collision pose relative to world
* fixed getting model pose relative to world
* minor refactoring

The link_features's bounding box test was previously skipped for bullet-featherstone plugin. This PR enables this test for bullet-featherstone by requiring only a minimal set of features needed to run the test.

## Checklist
- [x] Added tests

### DIFF
--- a/bullet-featherstone/src/KinematicsFeatures.cc
+++ b/bullet-featherstone/src/KinematicsFeatures.cc
@@ -22,30 +22,40 @@ namespace gz {
 namespace physics {
 namespace bullet_featherstone {
 
+
+FrameData3d getNonBaseLinkFrameData(const ModelInfo *_modelInfo,
+                                    const LinkInfo *_linkInfo)
+{
+  const auto index = _linkInfo->indexInModel.value();
+  FrameData3d data;
+  data.pose = GetWorldTransformOfLink(*_modelInfo, *_linkInfo);
+  const auto &link = _modelInfo->body->getLink(index);
+  data.linearVelocity = convert(link.m_absFrameTotVelocity.getLinear());
+  data.angularVelocity = convert(link.m_absFrameTotVelocity.getAngular());
+  return data;
+}
+
 /////////////////////////////////////////////////
 FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
     const FrameID &_id) const
 {
-  const auto linkIt = this->links.find(_id.ID());
+  bool isModel = false;
+  bool isCollision = false;
   const ModelInfo *model = nullptr;
+  Eigen::Isometry3d collisionPoseOffset = Eigen::Isometry3d();
+
+  const auto linkIt = this->links.find(_id.ID());
   if (linkIt != this->links.end())
   {
     const auto &linkInfo = linkIt->second;
-    const auto indexOpt = linkInfo->indexInModel;
     model = this->ReferenceInterface<ModelInfo>(linkInfo->model);
 
-    if (indexOpt.has_value())
+    if (linkInfo->indexInModel.has_value())
     {
-      const auto index = *indexOpt;
-      FrameData data;
-      data.pose = GetWorldTransformOfLink(*model, *linkInfo);
-      const auto &link = model->body->getLink(index);
-      data.linearVelocity = convert(link.m_absFrameTotVelocity.getLinear());
-      data.angularVelocity = convert(link.m_absFrameTotVelocity.getAngular());
-      return data;
+      return getNonBaseLinkFrameData(model, linkInfo.get());
     }
 
-    // If indexOpt is nullopt then the link is the base link which will be
+    // If indexInModel is nullopt then the link is the base link which will be
     // calculated below.
   }
   else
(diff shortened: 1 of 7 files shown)


bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -59,60 +69,55 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
   FrameData data;
-  if(model && model->body)
+  if (model && model->body)
   {
-    data.pose = convert(model->body->getBaseWorldTransform())
-      * model->baseInertiaToLinkFrame;
+    data.pose = convert(model->body->getBaseWorldTransform());
+    if (!isModel)
+      data.pose = data.pose * model->baseInertiaToLinkFrame;
+    if (isCollision)
```
azeey, line 119: I'm not sure why we need this here. Aren't collisions handled above starting from line 81?

bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -59,60 +69,55 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
-          data.angularVelocity = convert(
-            link.m_absFrameTotVelocity.getAngular());
-          return data;
+          const auto &linkInfo2 = linkIt2->second;
+          model = this->ReferenceInterface<ModelInfo>(linkInfo2->model);
+          collisionPoseOffset = collisionInfo->linkToCollision;
+          if (linkInfo2->indexInModel.has_value())
+          {
+            auto data = getNonBaseLinkFrameData(model, linkInfo2.get());
+            data.pose = data.pose * collisionPoseOffset;
```
azeey, line 96: This works for poses, but not for velocities since the offset needs to be taken into account. Do you mind creating an issue?

azeey (commented):
Mostly looks good. I just have one question.

bullet-featherstone/src/KinematicsFeatures.cc
iche033: This function could probably be refactored better. The logic starting in line 81 handles collisions for 'non-base' links,  i.e. if `linkInfo2->indexInModel.has_value()` is true, it's a non-base link. There is a similar check in many places of this function - if this check fails, it all comes down to this part of code (starting line 114) that computes frame data using the base link.

bullet-featherstone/src/KinematicsFeatures.cc
iche033: yes created #648

bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -59,60 +69,55 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
-      model = this->FrameInterface<ModelInfo>(_id);
   }
 
   FrameData data;
-  if(model && model->body)
+  if (model && model->body)
   {
-    data.pose = convert(model->body->getBaseWorldTransform())
-      * model->baseInertiaToLinkFrame;
+    data.pose = convert(model->body->getBaseWorldTransform());
```
azeey, line 116: Don't we have to account for 
- The offset from the model to the base link, i.e, the links pose in SDF? Do we have a test that covers this?
- The offset of the link frame to the center of mass? I thought that's what `model->baseInertiaToLinkFrame` was there for .

iche033: This line computes the pose for the model. The next 2 lines takes into account the pose of the link (and inertia offset if any) which uses `model->baseInertiaToLinkFrame`. I think that's correct? I added a test in [5c6ec0f](https://github.com/gazebosim/gz-physics/pull/647/commits/5c6ec0ff7adc80245b0653580c96cf306a48a7bc) that checks the frame data for a link and a collision that has a pose offset.

iche033: expanded test to cover both base and non-base links. [84054fa](https://github.com/gazebosim/gz-physics/pull/647/commits/84054faf625898319f469f3f6eff0a11d18f238d)

azeey: It looks like this test is not testing model frame semantics at all. Fortunately, we currently don't use it in gz-sim, but it would be good to fix (not in this PR). I've added a failing test in the `azeey/kinematic_feature_failing_test`  branch. Evidently, dartsim fails the test as well.

azeey (approved):
The fix for Collision looks good to me. I think there's still an issue with model FrameData calculation, but I think it would be best to address in a separate PR.

iche033:
Thanks for adding the test. I'm tracking all frame data issues in https://github.com/gazebosim/gz-physics/issues/648

iche033:
> I think there's still an issue with model FrameData calculation, but I think it would be best to address in a separate PR.

Follow-up PR: #649

Issue #648, opened 2024-06-04
bullet-featherstone: Velocities FrameData for collisions do not take into account collision pose offset. 


iche033:
## Environment
* OS Version: Ubuntu 20.04
* Source or binary build? source build, gz-physics7 branch

## Description
In bullet-featherstone plugin, The frame data computation is done in the [KinematicsFeatures::FrameDataRelativeToWorld](https://github.com/gazebosim/gz-physics/blob/74c4560e7b263e46b61dacff7ee30fff9efd5a19/bullet-featherstone/src/KinematicsFeatures.cc#L26). The Pose FrameData for collisions was updated to take into account collision pose offset in #647. However, the velocities FrameData still do not incorporate the collision pose offset.

iche033:
The JointFrameSemantics may also be incorrect in bullet-featherstone plugin. After fixing / enabling [these frame data checks](https://github.com/gazebosim/gz-physics/blob/74c4560e7b263e46b61dacff7ee30fff9efd5a19/test/common_test/kinematic_features.cc#L150-L155), the test fails:

```
97: Expected equality of these values:
97:   F_WCexpected.pose.translation()
97:     Which is: { 0, 1.23942, -0.569664 }
97:   childLinkFrameData.pose.translation()
97:     Which is: { 0, 0.619709, 0.215168 }
```

The `F_WCexpected.pose.translation()` calculated using joint frame semantics and the resulting values are incorrect, while the values of `childLinkFrameData.pose.translation` are correct.

iche033:
ModelFrame data is also incorrect, see https://github.com/gazebosim/gz-physics/pull/647#discussion_r1626874571.

This commit adds a failing test: https://github.com/gazebosim/gz-physics/commit/27deae998d00387e8dbffc3c10cfeaf42f22d21a

Issue #617, opened 2024-04-05
bullet-featherstone: Non-base links return zero velocities


iche033:
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build?
source build, `gz-physics7` and `main` branches

## Description

* Expected behavior: A moving non-base link should return correct linear and angular velocity values. 
* Actual behavior: Zero velocities are returned

In the bullet-featherstone plugin, non-base links currently returns zero linear and angular linear velocities. The relevant code for generating velocities for non-base links are located here:
https://github.com/gazebosim/gz-physics/blob/2bc19ad0d7ca34f2150c765e8cce5d712ba117bd/bullet-featherstone/src/KinematicsFeatures.cc#L43-L44

A closer look at the bullet source code reveals that the `btMultiBodyLink`'s `m_absFrameTotVelocity` variables are never filled unless `btMultiBody::useGlobalVelcoties()` is called and set to `true`

PR #649, opened 2024-06-05
bullet-featherstone: Fix model frame data pose

iche033:
# 🦟 Bug fix

Related issue #648

## Summary

Fix bullet-featherstone's model frame data pose.
Also expanded test to cover nested model / link / collision

More info:

I added a couple of fields in `ModelInfo` (`modelToRootLinkTf` and `nestedModelFromRootModelTf`) to store transforms that are computed in SDFFeatures. These transforms are then used in KinematicFeatures for computing model and nested model frame data.

## Checklist
- [x] Added tests


azeey:
@scpeters to review

bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -113,10 +113,17 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
+        * model->baseInertiaToLinkFrame;
+    if (isModel)
+    {
+      data.pose = model->modelToRootLinkTf.inverse() * data.pose;
+      if (model->isNestedModel)
+      {
+        data.pose = data.pose * model->nestedModelFromRootModelTf;
+      }
+    }
+    else if (isCollision)
```
azeey, line 126: This not really part of this PR, but I was wondering why whe need to handle `isCollision` here. Isn't it handled above?

bullet-featherstone/src/SDFFeatures.cc
```diff
@@ -517,8 +547,22 @@ Identity SDFFeatures::ConstructSdfModelImpl(
...
+          auto errors = _sdfModel.SemanticPose().Resolve(modelPose,
+              modelScopedName);
+          if (!errors.empty())
+            this->GenerateInvalidId();
+          auto modelToNestedModel = math::eigen3::convert(modelPose).inverse();
+
           return this->AddNestedModel(
-              _model->Name(), modelIdentity, worldIdentity, rootInertialToLink,
+              _model->Name(), modelIdentity, worldIdentity,
+              *modelToRootLink, rootInertialToLink,
```
azeey, line 564: What do you think about using applying the extra transform from the root link to the nested model here so that we don't have to apply a special case for nested models in `KinemtacisFeatures`?

azeey (commented):
I think we can avoid having a special case for nested models. Otherwise, LGTM!

azeey:
One additional thing I noticed is that the linear velocity is not properly calculated for things that have an offset from the root link. We need to have add an `omega x r` term to what we have now. Maybe just open an issue for now?

Also saw https://github.com/gazebosim/gz-physics/issues/617 after writing this, so maybe not relevant right now.

bullet-featherstone/src/SDFFeatures.cc
iche033: ok I think I understood what you mean now. I refactored the code in [77e8a6e](https://github.com/gazebosim/gz-physics/pull/649/commits/77e8a6ef9d7cf675f79a31b16541ff81c93759d5). I was not able to do `structure.rootLink->SemanticPose()->Resolve` because I think `structure.rootLink` does not have a view of the whole SDF tree containing the nested model so I computed the transform manually.

bullet-featherstone/src/Base.hh
```diff
@@ -132,9 +132,12 @@ struct ModelInfo
   std::string name;
   Identity world;
   int indexInWorld;
+  Eigen::Isometry3d rootLinkToModelTf;
   Eigen::Isometry3d baseInertiaToLinkFrame;
   std::shared_ptr<GzMultiBody> body;
 
+  bool isNestedModel = false;
```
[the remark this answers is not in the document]
iche033: removed. [3d4c840](https://github.com/gazebosim/gz-physics/pull/649/commits/3d4c840acde2814a83e52614fcce23c5aa1a1b16)

bullet-featherstone/src/KinematicsFeatures.cc
```diff
@@ -110,13 +112,16 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
   if (model && model->body)
   {
-    data.pose = convert(model->body->getBaseWorldTransform());
-    if (!isModel)
-      data.pose = data.pose * model->baseInertiaToLinkFrame;
-    if (isCollision)
+    data.pose = convert(model->body->getBaseWorldTransform())
+        * model->baseInertiaToLinkFrame;
+    if (isModel)
+      data.pose = model->rootLinkToModelTf * data.pose;
```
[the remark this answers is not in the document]
iche033: ah yes good catch. Added pose rotatation to the nested model in test and confirms that it fails. Fixed in [3d4c840](https://github.com/gazebosim/gz-physics/pull/649/commits/3d4c840acde2814a83e52614fcce23c5aa1a1b16)

PR #664, opened 2024-07-03
bullet-featherstone: Use a single contact manifold for each convex decomposed mesh collision

iche033:
# 🦟 Bug fix

## Summary

Contact point optimization for convex decomposed mesh collisions in bullet-featherstone plugin. This PR reduces the number of contact points generated for convex decomposed mesh collision by using a single contact manifold.

Previously when a mesh collision is decomposed into multiple convex hull shape, each convex hull shape has a corresponding contact manifold which can generate up to 4 contact points (4 is the default limit in bullet). This slows down sim especially when querying contact points (via `GetContactsFromLastStep`). This PR overrides the `btCollisionDispatcher` class in order to update the contact manifolds stored in bullet. It post processes the contact points, generates a single contact manifold for each collision associated with convex hull shapes, and clears the exiting contact manifolds. 

Note: it would be more efficient if we can do this inside bullet to avoid generating all the contact manifolds in the first place.

Before: many points can be generated from one convex decomposed mesh collision.
bullet_mesh_contacts

After: limit contact points to <= 4 
bullet_mesh_contacts_single_manifold

Here're remotery graphs showing before and after the fix for a scene with multiple convex decomposed meshes. Note objects were auto-deactivated when these graphs were captured.

Before: >200 contact points
bullet_deactivated_profile_cropped

After: 20 contact points
bullet_deactivated_with_fix_profile_cropped

Notable speedups are:
* `PhysicsPrivate::Step`: 0.153ms -> 0.097ms
* `PhysicsPrivate::UpdateCollision`:  0.528ms -> 0.135ms

## Checklist
- [x] Added tests


azeey:
Looks like there are several test failures in windows and I don't see any in the stable branch (https://build.osrfoundation.org/view/gz-harmonic/job/gz_physics-7-win/)

iche033:
> Looks like there are several test failures in windows and I don't see any in the stable branch (https://build.osrfoundation.org/view/gz-harmonic/job/gz_physics-7-win/)

Fixed and this PR is ready for review again. The test failures were due to  RTTI access violation when doing a `dynamic_cast` from `btCollisionShape` to `btCompoundShape`. I saw a similar [issue](https://github.com/dartsim/dart/issues/1522) reported in dartsim and [fixed](https://github.com/dartsim/dart/pull/1541) by adding pragmas in base classes but that's not viable here so I switched to `static_cast`.

azeey:
@azeey to review.

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+
+      if (!colShape0 || !colShape1 ||
+         (!this->HasConvexHullChildShapes(colShape0) &&
+          !this->HasConvexHullChildShapes(colShape1)))
+      {
+        this->manifoldsToKeep.insert(contactManifold);
+        continue;
+      }
+
+      btPersistentManifold* colManifold =
```
azeey, line 196: When we remove manifolds, we don't do anything with the pointer in `colPairManifolds`, so we it might not be a nullptr even if the manifold has been removed. I suggest checking if it's in `customManifolds` instead of just a `nullptr` check.

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+  }
+}
+
+/////////////////////////////////////////////////
+void GzCollisionDispatcher::releaseManifold(btPersistentManifold *_manifold)
+{
+  auto manifoldIt = this->manifoldsToKeep.find(_manifold);
+  if (manifoldIt != this->manifoldsToKeep.end())
+    this->manifoldsToKeep.erase(manifoldIt);
+
```
azeey, line 224: Do we need to also remove it from `customManifolds`?

bullet-featherstone/src/Base.hh
```diff
@@ -551,6 +606,10 @@ class Base : public Implements3d<FeatureList<Feature>>
       if (link->collider)
       {
         world->world->removeCollisionObject(link->collider.get());
+        GzCollisionDispatcher *dispatcher =
+            dynamic_cast<GzCollisionDispatcher *>(
+            world->world->getDispatcher());
+        dispatcher->RemoveManifoldByCollisionObject(link->collider.get());
```
azeey, line 612: Should we remove the object from the `colPairManifolds` map too? Otherwise, that map will grow indefinitely, which might become a problem if objects are added and removed frequently.

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+    const btMultiBodyLinkCollider* ob1 =
+        dynamic_cast<const btMultiBodyLinkCollider *>(
+        contactManifold->getBody1());
+
+    // if it's a custom manifold, just refresh contacts, no need to
+    // loop through and check them.
+    if (this->customManifolds.find(contactManifold) !=
+        this->customManifolds.end())
+    {
+      contactManifold->refreshContactPoints(ob0->getWorldTransform(),
```
azeey, line 163: Is this necessary? Do you know that Bullet itself won't be doing this?

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+  if (childCount > 0)
+  {
+    if (_childIndex >= 0 && _childIndex < childCount)
+    {
+      // todo(iche033) We do not have sufficient info to determine which
+      // child shape is the collision if the link has convex decomposed mesh
+      // collisions alongside of other collisions. See following example:
+      // parentLink -> boxShape0
+      //            -> boxShape1
+      //            -> comopundShape -> convexShape0
```
azeey, line 103:
```suggestion
      //            -> compoundShape -> convexShape0
```

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+      //                             -> convexShape1
+      // A _childIndex of 1 is ambiguous as it could refer to either
+      // boxShape1 or convexShape1
```
azeey, line 106: Could we use the `getUserIndex` or `getUserPointer` to help with this somehow?

azeey (commented):
Well, this was not a walk in the park :), but I think I finally understand how things are working. I don't see any major problems, but it would be good to test what happens when objects are removed. Another point is about the ambiguity with the collision index. I'm wondering if we can find a way around that. No worries if you don't have time to tackle that right now though.

bullet-featherstone/src/Base.hh
```diff
@@ -53,6 +53,61 @@ namespace gz {
 namespace physics {
 namespace bullet_featherstone {
 
+/// \brief Custom gz collision dispatcher
+class GzCollisionDispatcher : public btCollisionDispatcher
+{
+  /// \brief Constructor
+  public: GzCollisionDispatcher(
```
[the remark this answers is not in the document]
iche033: [15d4ede](https://github.com/gazebosim/gz-physics/pull/664/commits/15d4ede77239f9a1b35971a13a9c3fe4ccaa1bca)

bullet-featherstone/src/Base.cc
```diff
@@ -26,14 +26,213 @@ namespace gz {
...
+  {
+    btCollisionDispatcher::releaseManifold(manifold);
+  }
+
+  this->customManifolds.clear();
+  this->colPairManifolds.clear();
+}
+
+/////////////////////////////////////////////////
+void GzCollisionDispatcher::RemoveManifoldByCollisionObject(
```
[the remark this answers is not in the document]
iche033: updated `MeshContact` test in `collisions.cc` test to remove the model with convex decomposed mesh collision. It should trigger this call

bullet-featherstone/src/Base.cc
iche033: I think this is only be called for contact manifolds owned by bullet, and not the custom ones created by gz-physics. I tested manually by removing models with convex decomposed collisions and did not notice that custom manifolds were passed to this function.

bullet-featherstone/src/Base.hh
iche033: yes removed them from `colPairManifolds`. [15d4ede](https://github.com/gazebosim/gz-physics/pull/664/commits/15d4ede77239f9a1b35971a13a9c3fe4ccaa1bca)

bullet-featherstone/src/Base.cc
iche033: I remember I had to keep this call for the multi body link transform to be updated. As a test, I dropped a model with convex decomposed collisions on the ground -> this worked fine, but as soon as I moved it to a different pose, it's transforms are not updated.

bullet-featherstone/src/Base.cc
iche033: I updated the code to remove the pointers in `colPairManifolds` when manifolds are removed. So the `colPairManifolds` map should always be up-to-date now. [15d4ede](https://github.com/gazebosim/gz-physics/pull/664/commits/15d4ede77239f9a1b35971a13a9c3fe4ccaa1bca)

iche033:
I addressed most of the comments but not all. I still need to figure out how to resolve the issue with ambiguous collision index and then add a test.

caguero:
@iche033 , some conflicts here.

PR #694, opened 2024-09-27
Contact, wheel slip, and friction documentation

scpeters:
# 🎉 New feature

## Summary

This adds the following documentation of the following topics:

* Simulation of contact, slip, and friction in Gazebo
* Special considerations for modeling wheel contact / slip / friction

I'm not sure if a gz-physics tutorial is the right place for this; I'm open to moving it elsewhere. It still needs a few figures, but it collects a lot of knowledge about contact simulation in general and specifically for wheel-terrain contact. It includes both Gazebo Classic and gz-sim.

## Checklist
- [ ] Added tests


iche033 (commented):
This tutorial has really great info and I think we should try to get this in! 

One thing I noticed is that it refers to ODE and Gazebo-classic's code / unmerged PRs in many places to describe in detail how some features are implemented there. I am guessing some of these features are not available in the modern gazebo yet. I think it's fine to point to classic if it helps to illustrate various concepts presented but it would good to clarify whether which of these features are available in the modern Gazebo yet. Focusing more on the modern gazebo would be nice so that the tutorial does not seem like it's written for classic.

azeey:
> I'm not sure if a gz-physics tutorial is the right place for this; I'm open to moving it elsewhere.

We don't have clear policy about this, but the way I'm thinking about it right now is that if the tutorial describes how to use the API in the package or is written to show how to use a specific feature/plugin in the package, it should be kept with the package. Otherwise, it should go to https://github.com/gazebosim/docs. In this case, since most of the content is conceptual, I'd vote for moving it to `gazebosim/docs`.  Thoughts?

PR #700, opened 2024-11-01
bullet-featherstone: Fix finding free group for a body with fixed base

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-sim/issues/2663

## Summary

Finding free group for a body that has a fixed base results an invalid ID, which causes set world pose from gz-sim to fail, see https://github.com/gazebosim/gz-sim/issues/2663. This PR removes the fixed base body check and I verified that set world pose works for a body with fixed base.  

Another minor change - I noticed that finding free group for links has the same check. Currently free groups only works for models in bullet-featherstone so I updated the function to just call `FindFreeGroupForModel`.

## Checklist
- [x] Added tests

### DIFF
--- a/bullet-featherstone/src/FreeGroupFeatures.cc
+++ b/bullet-featherstone/src/FreeGroupFeatures.cc
@@ -85,13 +85,24 @@ Identity FreeGroupFeatures::FindFreeGroupForModel(
 {
   const auto *model = this->ReferenceInterface<ModelInfo>(_modelID);
 
-  // Also reject if the model is a child of a fixed constraint
-  // (detachable joint)
+  std::cerr <<  "joints " << this->joints.size() << std::endl;
   for (const auto & joint : this->joints)
   {
+    // Also reject if the model is a child of a fixed constraint
+    // (detachable joint)
     if (joint.second->fixedConstraint)
     {
       if (joint.second->fixedConstraint->getMultiBodyB() == model->body.get())
+      {
+        return this->GenerateInvalidId();
+      }
+    }
+    // Reject if the model has a world joint
+    if (std::size_t(joint.second->model) == std::size_t(_modelID))
+    {
+      const auto *identifier =
+          std::get_if<RootJoint>(&joint.second->identifier);
+      if (identifier)
       {
         return this->GenerateInvalidId();
       }
--- a/bullet-featherstone/src/FreeGroupFeatures.cc
+++ b/bullet-featherstone/src/FreeGroupFeatures.cc
@@ -85,7 +85,6 @@ Identity FreeGroupFeatures::FindFreeGroupForModel(
 {
   const auto *model = this->ReferenceInterface<ModelInfo>(_modelID);
 
-  std::cerr <<  "joints " << this->joints.size() << std::endl;
   for (const auto & joint : this->joints)
   {
     // Also reject if the model is a child of a fixed constraint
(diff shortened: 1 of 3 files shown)


test/common_test/free_joint_features.cc
```diff
@@ -316,6 +316,71 @@ TEST_F(FreeGroupFeaturesTest, FreeGroupSetWorldPosePrincipalAxesOffset)
   }
 }
 
+TEST_F(FreeGroupFeaturesTest, FreeGroupSetWorldPoseStaticModel)
```
azeey, line 319: Do we need to check how it behaves with non-static models attached to the world with a fixed joint?

iche033: updated code to make sure non-static model with a world joint can not be moved, also added a test a7e6eb5

bullet-featherstone/src/FreeGroupFeatures.cc
```diff
@@ -85,37 +85,39 @@ Identity FreeGroupFeatures::FindFreeGroupForModel(
...
+    // (detachable joint)
     if (joint.second->fixedConstraint)
     {
       if (joint.second->fixedConstraint->getMultiBodyB() == model->body.get())
       {
         return this->GenerateInvalidId();
       }
     }
+    // Reject if the model has a world joint
+    if (std::size_t(joint.second->model) == std::size_t(_modelID))
```
azeey, line 100: I think this is good for matching the behavior of DART, but we might want to consider the allowing models with world joints to move. I think it would be more consistent with also allowing static models to move. I won't block on this, but I'd like to hear @scpeters thoughts on it.

azeey (approved):

test/common_test/free_joint_features.cc
```diff
@@ -316,6 +316,121 @@ TEST_F(FreeGroupFeaturesTest, FreeGroupSetWorldPosePrincipalAxesOffset)
   }
 }
 
+TEST_F(FreeGroupFeaturesTest, FreeGroupSetWorldPoseStaticAndFixedModel)
+{
+  const std::string modelStr = R"(
```
scpeters, line 321: nit: rename to `modelStaticStr`

bullet-featherstone/src/FreeGroupFeatures.cc
scpeters: according to our documentation of `FreeGroup` in [FreeGroup.hh](https://github.com/gazebosim/gz-physics/blob/gz-physics8_8.0.0/include/gz/physics/FreeGroup.hh#L33-L34):

> the FreeGroup class, which represents a group of links that are not connected to the world with any kinematic constraints

I think it could be a nice user feature in gz-sim to be able to move objects that have a joint connecting them to the world, but by the definition of a FreeGroup, I don't think we can support that with this gz-physics API. I think we would need a gz-sim feature that detaches a joint, moves, and then reattaches the joint.

scpeters (approved):

test/common_test/free_joint_features.cc
iche033: [ded1e0a](https://github.com/gazebosim/gz-physics/pull/700/commits/ded1e0a7efcdb9c61d69e917a5c242a4aa423bbc)

bullet-featherstone/src/FreeGroupFeatures.cc
iche033: ok sounds like a new feature to add  so I'll keep this PR as is.

iche033:
@mergifyio backport gz-physics7

Issue #712, opened 2025-01-13
bullet-featherstone: support joint damping and friction parameters


scpeters:
Part of https://github.com/gazebosim/gz-physics/issues/545.

## Desired behavior

The [//joint/axis*/dynamics/damping](http://sdformat.org/spec?ver=1.11&elem=joint#dynamics_damping) and [//joint/axis*/dynamics/friction](http://sdformat.org/spec?ver=1.11&elem=joint#dynamics_friction) parameters are [read from SDFormat by the bullet-featherstone SDFFeatures::ConstructSdfModelImpl method](https://github.com/gazebosim/gz-physics/blob/gz-physics8/bullet-featherstone/src/SDFFeatures.cc#L788-L791) and written to the `m_jointDamping` and `m_jointFriction` member variables of [btMultibodyLink](https://github.com/bulletphysics/bullet3/blob/master/src/BulletDynamics/Featherstone/btMultiBodyLink.h), but a [todo code comment in the btMultibodyLink.h](https://github.com/bulletphysics/bullet3/blob/master/src/BulletDynamics/Featherstone/btMultiBodyLink.h#L147-L148) indicates that those variables are not yet used.

We either need to implement support for these parameters in upstream bullet or implement an explicit joint damping workaround (see [gazebo::physics::ODEJoint::ApplyExplicitStiffnessDamping](https://github.com/gazebosim/gazebo-classic/blob/b22c6e15e52299865b31093b8feebc9ca19e26e8/gazebo/physics/ode/ODEJoint.cc#L1192-L1210) from gazebo-classic for reference).

Sukhvansh2004:
Hi, can I work on this task?

iche033:
I think we'll need to first confirm that built-in support for joint damping / friction is  available in [bullet-featherstone implementation upstream](https://github.com/bulletphysics/bullet3/tree/master/src/BulletDynamics/Featherstone) or not. If not, I found this old post for doing it using [btMultibodyJointMotor](https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=10942). Note that in our ode, we currently only [create a joint motor](https://github.com/gazebosim/gz-physics/blob/6b41b750f60d606a422a1e41c6e408fe37f42aa4/bullet-featherstone/src/JointFeatures.cc#L440-L454) when users set joint velocities.

PR #738, opened 2025-04-23
Add joint damping support in bullet-featherstone plugin

shameekganguly:
# 🦟 Bug fix

Fixes missing joint damping effect when `//joint/axis/dynamics/damping` is specified in the model sdf in the bullet-featherstone plugin. Also adds a new test.

Partially addresses https://github.com/gazebosim/gz-physics/issues/712

## Summary
Added a method in the `GzMultiBody` wrapper class to add joint damping torques computed based on the last updated joint velocity. This method is called for each joint before stepping the Bullet world in `ForwardStep`.

In the process, I also found that we were using the wrong identifier (`InternalJoint`) for detachable joints added through the `AddFixedJoint` feature. Switched to a new `FixedJointConstraint` identifier to fix this. Otherwise, we'd get a segfault when adding the joint damping torque.

Also, previously we were setting the joint damping value in `body->link[i]->m_jointDamping`. I removed this and saved the set damping value in the `JointInfo` struct instead so that once the upstream feature request for joint damping (https://github.com/bulletphysics/bullet3/issues/4709) is addressed, we don't accidentally apply damping torques twice (once in the engine and again in the plugin). Currently, only homogeneous damping across all dofs of a particular joint will be supported.

## Checklist
- [x] Added tests

### DIFF
--- a/test/common_test/simulation_features.cc
+++ b/test/common_test/simulation_features.cc
@@ -567,7 +567,7 @@ TYPED_TEST(SimulationFeaturesDynamicsTest, JointDamping)
     // Assuming m, g, d and t denote mass, gravity, damping and time
     // respectively, the motions can be analytically computed as follows:
     //
-    // m * dot(v) + d * v = m * g
+    // m * dot(v) + d * v = -m * g
     // dot(z) = v
     // where v and z denote joint velocity and position respectively.
     //
--- a/test/common_test/simulation_features.cc
+++ b/test/common_test/simulation_features.cc
@@ -546,29 +546,58 @@ TYPED_TEST(SimulationFeaturesDynamicsTest, JointDamping)
       ASSERT_NE(nullptr, root.Model());
       world->ConstructModel(*root.Model());
     };
-    addModel("model1", {}, 5.0);
+    constexpr double kGravityAcc = 9.8;
+    constexpr double kMass = 2.0;
+    constexpr double kDamping = 5.0;
+    addModel("model1", {}, kDamping);
     addModel("model2", gz::math::Pose3d(1, 0, 0, 0, 0, 0), 0.0);
-
-    auto [checkedOutput, output] =
-      StepWorld<FeaturesDynamics>(world, true, 3000);
-    EXPECT_TRUE(checkedOutput);
 
     const auto jointWithDamping =
         world->GetModel("model1")->GetJoint("test_joint");
     const auto jointWithoutDamping =
         world->GetModel("model2")->GetJoint("test_joint");
 
+    // Step for a small amount of time (< 1s) and verify that the damped joint
+    // moved to a smaller position and velocity compared to the undamped joint.
+    auto [checkedOutput, output] =
+      StepWorld<FeaturesDynamics>(world, true, 500);
+    EXPECT_TRUE(checkedOutput);
+
+    // Default dt is 0.001s.
+    double tElapsed = 0.5;
+
+    // Compute the expected difference in position between damped and undamped
+    // joints by taking the 4th order taylor series approximation. This is a
+    // slight under-estimation of the expected difference (5th order approx.
+    // would over estimate).
+    double expectedPosDiff =
+        kGravityAcc * kDamping / kMass * std::pow(tElapsed, 3) / 6.0 *
+        (1 - tElapsed * kDamping / kMass / 4.0);
+
     auto posWithDamping = jointWithDamping->GetPosition(0);
     auto posWithoutDamping = jointWithoutDamping->GetPosition(0);
-    EXPECT_GT(posWithDamping, posWithoutDamping);
+    EXPECT_GT(posWithDamping, expectedPosDiff + posWithoutDamping);
+
+    // Compute the expected difference in velocity between damped and undamped
+    // joints by taking the 3rd order taylor series approximation. This is a
+    // slight under-estimation of the expected difference (4th order approx.
+    // would over estimate).
+    double expectedVelDiff =
+        kGravityAcc * kDamping / kMass * std::pow(tElapsed, 2) / 2.0 *
+        (1 - tElapsed * kDamping / kMass / 3.0);
 
     auto velWithDamping = jointWithDamping->GetVelocity(0);
     auto velWithoutDamping = jointWithoutDamping->GetVelocity(0);
-    EXPECT_GT(velWithDamping, velWithoutDamping);
+    EXPECT_GT(velWithDamping, expectedVelDiff + velWithoutDamping);
+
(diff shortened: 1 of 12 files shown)


test/common_test/simulation_features.cc
```diff
@@ -469,6 +470,108 @@ TYPED_TEST(SimulationFeaturesFallingTest, Falling)
...
+    EXPECT_TRUE(checkedOutput);
+
+    const auto jointWithDamping =
+        world->GetModel("model1")->GetJoint("test_joint");
+    const auto jointWithoutDamping =
+        world->GetModel("model2")->GetJoint("test_joint");
+
+    auto posWithDamping = jointWithDamping->GetPosition(0);
+    auto posWithoutDamping = jointWithoutDamping->GetPosition(0);
+    EXPECT_GT(posWithDamping, posWithoutDamping);
```
scpeters, line 563: let's strengthen this expectation so it's more than just `>`. I haven't tested this specific suggestion; it's just an example, please find an appropriate constant

```suggestion
    EXPECT_GT(posWithDamping, 1.0 + posWithoutDamping);
```

test/common_test/simulation_features.cc
```diff
@@ -469,6 +470,108 @@ TYPED_TEST(SimulationFeaturesFallingTest, Falling)
...
+
+    auto velWithDamping = jointWithDamping->GetVelocity(0);
+    auto velWithoutDamping = jointWithoutDamping->GetVelocity(0);
+    EXPECT_GT(velWithDamping, velWithoutDamping);
```
scpeters, line 567: please strengthen this expectation as well

scpeters (commented):
the overall approach looks good, basically adding numerically explicit joint damping

just a few test expectations that should be strengthened

test/common_test/simulation_features.cc
shameekganguly: Done, see https://github.com/gazebosim/gz-physics/pull/738/commits/8813604329220184278e086946b3ee31f317e01c

iche033 (approved):
looks good to me.

test/common_test/simulation_features.cc
```diff
@@ -469,6 +470,137 @@ TYPED_TEST(SimulationFeaturesFallingTest, Falling)
...
+      StepWorld<FeaturesDynamics>(world, true, 500);
+    EXPECT_TRUE(checkedOutput);
+
+    // Default dt is 0.001s.
+    double tElapsed = 0.5;
+
+    // Compute the expected difference in position between damped and undamped
+    // joints by taking the 4th order taylor series approximation. This is a
+    // slight under-estimation of the expected difference (5th order approx.
+    // would over estimate).
```
scpeters, line 572: sorry to nitpick here, but I'd prefer that we show our work in the comments if we're going to use principled expectations here (for example the comments in the [joint wrench test](https://github.com/gazebosim/gz-physics/blob/cd8657514226cedb571d2b539c51c8ebed340d38/test/common_test/joint_features.cc#L954-L1031))

test/common_test/simulation_features.cc
```diff
@@ -469,6 +470,137 @@ TYPED_TEST(SimulationFeaturesFallingTest, Falling)
...
+    auto velWithoutDamping = jointWithoutDamping->GetVelocity(0);
+    EXPECT_GT(velWithDamping, expectedVelDiff + velWithoutDamping);
+
+    // Step again for a long time so that the damped joint achieves close to
+    // terminal velocity.
+    StepWorld<FeaturesDynamics>(world, false, 2500);
+
+    // Verify that velocity with joint damping matches terminal velocity
+    // = mg/d
+    double expectedTerminalVel = -kGravityAcc*kMass/kDamping;
```
scpeters, line 599: this looks right to me

scpeters (commented):
regarding the test expectations updated in 8813604329220184278e086946b3ee31f317e01c, I think we can find an analytical solution

each model has a prismatic joint whose axis is aligned with gravity, so each link has a downward force due to gravity, and the damped link has a linear damping force. initial position and velocity are 0

constants `m,g,d` for mass, gravity, and damping

time `t`

undamped is trivial: `z_undamped = -0.5 g * t^2`

for the damped link with `z,v,a` representing position, velocity, and acceleration:

~~~
m a = - m g - d v
# equivalent to
a + d/m v = - g
# or
\dot{v} + d/m v = -g
~~~

this is a damped first order system with

~~~
v(t) = - m g / d * (1 - exp(-d / m * t))
~~~

integrating to get position

~~~
z(t) = - (m g / d) * (t + m / d * exp(-d / m * t))
~~~

if that seems right to you, we could add this as comments to the expectations

test/common_test/simulation_features.cc
shameekganguly: Added comments in https://github.com/gazebosim/gz-physics/pull/738/commits/a4d93ccf25d9cdcf10073bc50ee6284af9b95049
I also extended the test to 
- check the computed position and velocity of the damped and undamped joints directly (instead of the taylor series approximation of the difference. I originally used that approach to make the test more robust to numerical errors, but this is a lot more readable.)
- step a few times and perform the above checks at different elapsed times

Please take another look.

test/common_test/simulation_features.cc
```diff
@@ -469,6 +470,173 @@ TYPED_TEST(SimulationFeaturesFallingTest, Falling)
...
+    // and without damping are computed correctly upto integration errors.
+    //
+    // Each model has a prismatic joint whose axis is aligned with gravity, so
+    // each link has a downward force due to gravity, and the damped link also
+    // has a linear damping force. Initial position and velocity are 0 for both
+    // joints.
+    // Assuming m, g, d and t denote mass, gravity, damping and time
+    // respectively, the motions can be analytically computed as follows:
+    //
+    // m * dot(v) + d * v = m * g
```
scpeters, line 570:
```suggestion
    // m * dot(v) + d * v = -m * g
```

scpeters (commented):
amazing! I just see one place that looks like a sign error, but it doesn't affect the test expectations

test/common_test/simulation_features.cc
scpeters: thanks so much for adding the documentation and cleaning up the errors in my initial write-up! The Taylor series approximation sounds like a clever approach, but I couldn't immediately see how you got there. I do think this is more readable now

💜

test/common_test/simulation_features.cc
shameekganguly: Yep, fixed.

scpeters (approved):

scpeters:
https://github.com/Mergifyio backport gz-physics7

PR #834, opened 2025-12-27
Prevent crash by removing constrains which in deleted body

XINJIANGMO:
# 🦟 Bug fix

Fixes [gz-sim#3243](https://github.com/gazebosim/gz-sim/issues/3243)

## Summary

Although the object is removed, its associated `btMultiBodyConstraints` (e.g., mimic joints) are not cleaned up and remain in the physics world's constraint list. And accessing it in the next step will trigger the crash. 

In `gz-physics/bullet-featherstone/src/SimulationFeatures.cc`, inside `WorldForwardStep`, we should:
 1. Identify valid bodies.
 2. Remove any constraints referencing invalid (deleted) bodies.

## Steps to reproduce

1. gz sim -r -v 4 mimic_fast_slow_pendulums_world.sdf --physics-engine gz-physics-bullet-featherstone-plugin
2. run the service command
`gz service --timeout 10000 -s /world/default/remove --reptype gz.msgs.Boolean --reqtype gz.msgs.Entity --req 'id: 26
name: "pendulum_with_base_mimic_slow_follows_fast"`
3. crash

## Checklist
- [ ] Added tests


XINJIANGMO:
> Thanks for the fix @XINJIANGMO. Would you mind adding a test that would fail without this fix?

@azeey  I tried to add a test which will crash without the fix. But the test can not reproduce this `use-after-free` issues without this fix in current gtest-based setup and we also don’t have any public or testing API at this level to directly construct the dangling Bullet multibody constraint and force the use-after-free . I think it doesn’t  crash in this simpler test environment. Maybe run with `asan` wiil help.

PR #839, opened 2026-01-14
Prevent DART crash by adding check for negative damping

XINJIANGMO:
# 🦟 Bug fix

Fixes [#3266](https://github.com/gazebosim/gz-sim/issues/3266)

## Summary
Loading an SDF model with a negative joint damping coefficient (e.g., <damping>-5.0</damping>) causes a crash when using the gz-physics DART plugin in Coverage build modes.
This PR added a check for negative damping to avoid crash.

## Checklist
- [ ] Added tests

### DIFF
--- a/dartsim/src/SDFFeatures.cc
+++ b/dartsim/src/SDFFeatures.cc
@@ -220,11 +220,11 @@ static JointType *ConstructSingleAxisJoint(
     // Check for negative damping
     if (properties.mDampingCoefficients[0] < 0.0)
     {
-      gzerr << "Detected negative damping coefficient (" 
-             << properties.mDampingCoefficients[0] 
-             << ") for joint [" << _sdfJoint.Name() 
-             << "]. Damping must be non-negative. "
-             << "Setting damping to zero." << std::endl;
+      gzerr << "Detected negative damping coefficient ("
+            << properties.mDampingCoefficients[0]
+            << ") for joint [" << _sdfJoint.Name()
+            << "]. Damping must be non-negative. "
+            << "Setting damping to zero." << std::endl;
       properties.mDampingCoefficients[0] = 0.0;
     }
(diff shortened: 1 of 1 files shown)


dartsim/src/SDFFeatures.cc
```diff
@@ -217,6 +217,16 @@ static JointType *ConstructSingleAxisJoint(
   {
     properties.mAxis = ConvertJointAxis(sdfAxis, _modelInfo, _T_joint);
     CopyStandardJointAxisProperties(0, properties, sdfAxis);
+    // Check for negative damping
+    if (properties.mDampingCoefficients[0] < 0.0)
```
scpeters: I see a [dart assertion](https://github.com/dartsim/dart/blob/v6.15.0/dart/dynamics/detail/GenericJoint.hpp#L1425) in `setDampingCoefficient`, so I think it makes sense to check for this.

dartsim/src/SDFFeatures.cc
```diff
@@ -217,6 +217,16 @@ static JointType *ConstructSingleAxisJoint(
...
+    {
+      gzerr << "Detected negative damping coefficient ("
+            << properties.mDampingCoefficients[0]
+            << ") for joint [" << _sdfJoint.Name()
+            << "]. Damping must be non-negative. "
+            << "Setting damping to zero." << std::endl;
+      properties.mDampingCoefficients[0] = 0.0;
```
scpeters: I'm not opposed to clamping to `0.0`, but I just want to acknowledge that there are other actions we could take:

* fail to load the model (a bit extreme, I would suggest adding linter complaints about negative damping coefficients in SDFormat before taking this step)
* use the absolute value of damping instead of `0.0`

scpeters (approved):

PR #879, opened 2026-02-18
Add collisionShapeId to RayIntersectionT for raycast hit entity ident…

khalidbourr:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-sim/issues/3314

## Summary

DART's `RayHit` already contains `mCollisionObject` but it was being dropped in `GetRayIntersectionFromLastStep()`. This PR adds a `collisionShapeId` field to `RayIntersectionT` and populates it using the same `CollisionObject → getShapeFrame() → asShapeNode() → shapes.IdentityOf()` mapping that `convertContact()` already uses.

## Changes

- `include/gz/physics/GetRayIntersection.hh` 
- `include/gz/physics/detail/GetRayIntersection.hh`
- `dartsim/src/SimulationFeatures.cc`

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

## Note

Companion PR for gz-sim: https://github.com/gazebosim/gz-sim/pull/3336

PR #880, opened 2026-02-19
Add GetBatchRayIntersectionFromLastStepFeature for CPU lidar

apojomovsky:
# 🎉 New feature

This PR is 1 out of 3 that implement: https://github.com/gazebosim/gz-sensors/issues/26

## Summary

Add `GetBatchRayIntersectionFromLastStepFeature`, a new gz-physics feature that accepts all rays for a sensor in a single call and returns all results at once.

The dartsim implementation bypasses DART's `raycast()` wrapper, which allocates a temporary `btCollisionObject` + `btShape` per ray and calls `btCollisionWorld::rayTest()` directly.

To reach the underlying `btCollisionWorld*` without patching DART (where `BulletCollisionGroup::getBulletCollisionWorld()` is `protected`), a thin subclass `GzBulletCollisionGroup` bridges the protected method via inheritance, and `GzBulletCollisionDetector::createCollisionGroup()` is overridden to produce instances of it.

A batch API was chosen over modifying the existing per-ray `GetRayIntersectionFromLastStep` to avoid a silent behavior change to a public API, provide independent test coverage, and establish a clean boundary for future batching (e.g. GPU).

Only the Bullet collision detector supports raycasting in dartsim. Non-Bullet detectors (ODE, FCL, native DART) emit a one-time `gzwarn` and return NaN results.

## Test it

```bash
colcon build --merge-install --packages-select gz-physics \
  --cmake-args '-DBUILD_TESTING=ON'
colcon test --packages-select gz-physics \
  --ctest-args -R COMMON_TEST_simulation_features_dartsim
```

## Checklist
- [x] Added a screen capture or video to the PR description that demonstrates the feature
- [x] Added tests
- [x] Added example and/or tutorial
- [x] Updated Bazel files (if adding new files). Created an issue otherwise.

Generated-by: Claude Opus 4.6

Disclaimer: The code here was reviewed, tested, and profiled by hand by the author.


dartsim/src/SimulationFeatures.cc
```diff
@@ -238,6 +244,87 @@ SimulationFeatures::GetRayIntersectionFromLastStep(
...
+      results.push_back(std::move(intersection));
+    }
+
+    return results;
+  }
+
+  // Non-Bullet collision detectors do not support raycasting.
+  // Warn once per detector type and return NaN for all rays.
+  auto collisionDetector = world->getConstraintSolver()->getCollisionDetector();
+  static std::unordered_set<std::string> warnedDetectors;
```
j-rivero: Do we need to call `getConstraintSolver` here again after did it in line 263?

j-rivero (commented):
Thanks for the PR ! 

Not a big expert in physics but let me ask some questions and raise some proposals for the idea:

I somehow miss an API clean way of knowing if there was a hit or not in Ray, it is not documented anyway in the code. I assume that the user needs to check for NaN  probably in `fraction` but I think that NaN has its own problems for some floating point comparisons. 

With some help from Claude, how about something like?
https://gist.github.com/j-rivero/5a96bf40b33f0468366dbea82d431ffc

dartsim/src/GzCollisionDetector.hh
```diff
@@ -83,6 +84,17 @@ class GzOdeCollisionDetector :
...
 
+/// \brief Exposes BulletCollisionGroup::getBulletCollisionWorld() which
+/// is protected in the base class.
+class GzBulletCollisionGroup : public dart::collision::BulletCollisionGroup
+{
+  public: explicit GzBulletCollisionGroup(
+      const dart::collision::CollisionDetectorPtr &_detector);
+
+  /// \brief Return the underlying btCollisionWorld
+  public: btCollisionWorld *getCollisionWorld();
```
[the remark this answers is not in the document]
apojomovsky: Done, good catch. Updated both the declaration and implementation as suggested.

include/gz/physics/GetBatchRayIntersection.hh
```diff
@@ -0,0 +1,104 @@
...
+{
+
+/// \brief GetBatchRayIntersectionFromLastStepFeature is a feature for
+/// retrieving multiple ray intersections generated in the previous simulation
+/// step in a single call.
+class GZ_PHYSICS_VISIBLE GetBatchRayIntersectionFromLastStepFeature
+    : public virtual FeatureWithRequirements<ForwardStep>
+{
+  public: template <typename PolicyT>
+  struct RayIntersectionT
```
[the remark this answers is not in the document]
apojomovsky: This is a good question, and the answer is a bit long (sorry).

The batch feature bypasses DART's `collisionDetector->raycast()` and calls `btCollisionWorld::rayTest()` directly (that's where the speedup comes from), so it operates at a different abstraction level and needs its own types (`RayT`, and now `RayIntersectionT` with `bool` hit). The single-ray feature also wraps its result in `SpecifyData<RequireData<...>>`, so the return semantics differ. Unifying the structs would create a cross-feature dependency, which I wasn't sure fits the pattern where features are independently composable.

Open to suggestions if you have a preferred way to structure this.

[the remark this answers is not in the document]
jmachowinski: We are currently looking into implementing the CPU based lidar also for ODE. While implementing the raycast for GzOdeCollisionDetector I also noticed that calling raycast repeatedly in a row will be expensive. 

I wonder if we should add a batch raycast api to GzCollisionDetector or somewhere else were  we could sneak in a nice API.

The gold solution would be to add the interface directly into dart, but seeing that upstream dart has moved on an renamed all of its headers I don't see a shift to the latest version of dart before the next gazebo release.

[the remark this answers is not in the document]
jmachowinski: > you're suggesting we add something like a virtual BatchRaycast() on GzCollisionDetector so that each detector (Bullet, ODE, etc.) can provide its own optimized implementation, and SimulationFeatures.cc just calls through the virtual without needing to know which backend it's talking to?

@apojomovsky yes that would exactly be the idea.

dartsim/src/GzCollisionDetector.hh
```diff
@@ -42,6 +64,20 @@ class GzCollisionDetector
...
+  /// \brief Cast multiple rays against a collision group.
+  /// \param[in] _group The collision group to test against.
+  /// \param[in] _rays The rays to cast.
+  /// \param[out] _results One result per input ray, in the same order.
+  /// \return True if the detector supports batch raycasting, false otherwise.
+  ///   When false, _results is left empty and the caller should fall back.
+  public: virtual bool BatchRaycast(
+      CollisionGroup *_group,
+      const std::vector<GzRay> &_rays,
+      std::vector<GzRayResult> &_results) const;
```
jmachowinski:
```suggestion
  public: virtual std::optional<std::vector<GzRayResult>> BatchRaycast(
      CollisionGroup *_group,
      const std::vector<GzRay> &_rays) const;
```
Why not use optional ?

dartsim/src/GzCollisionDetector.hh
```diff
@@ -83,6 +119,17 @@ class GzOdeCollisionDetector :
...
+/// \brief Exposes BulletCollisionGroup::getBulletCollisionWorld() which
+/// is protected in the base class.
+class GzBulletCollisionGroup : public dart::collision::BulletCollisionGroup
+{
+  public: explicit GzBulletCollisionGroup(
+      const dart::collision::CollisionDetectorPtr &_detector);
+
+  /// \brief Return the underlying btCollisionWorld
+  public: const btCollisionWorld *getCollisionWorld() const;
+};
```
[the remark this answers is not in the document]
apojomovsky: Done

j-rivero (approved):
IMHO, is good to go technically. Will require the design and architecture overview from @scpeters or @azeey but it is a great work Alexis.

j-rivero (approved):
I said nothing, comment was already published sorry.

apojomovsky:
Gentle ping @azeey @scpeters . 
Please let me know if you have further comments for me to address.

dartsim/src/SimulationFeatures.cc
```diff
@@ -253,6 +256,47 @@ SimulationFeatures::GetRayIntersectionFromLastStep(
...
+  auto detector = solver->getCollisionDetector();
+  auto *gzDetector =
+    dynamic_cast<dart::collision::GzCollisionDetector *>(detector.get());
+
+  if (gzDetector)
+  {
+    std::vector<dart::collision::GzRay> gzRays;
+    gzRays.reserve(_rays.size());
+    for (const auto &ray : _rays)
+      gzRays.push_back({ray.origin, ray.target});
```
azeey: Was it not possible to just pass `_rays` to `BatchRaycast` and avoid this copy here?

include/gz/physics/GetBatchRayIntersection.hh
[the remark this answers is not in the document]
azeey: We could move the `RayIntersectionT` in `GetRayIntersectionFromLastStepFeature` to a more common header and use that it here to replace `point`, `fraction` and `normal` while keeping `hit`.

include/gz/physics/GetBatchRayIntersection.hh
```diff
@@ -0,0 +1,104 @@
...
+        const std::vector<RayQuery> &_rays) const;
+  };
+
+  public: template <typename PolicyT>
+  class Implementation : public virtual Feature::Implementation<PolicyT>
+  {
+    public: using RayIntersection = RayIntersectionT<PolicyT>;
+    public: using RayQuery = RayT<PolicyT>;
+
+    public: virtual std::vector<RayIntersection>
```
azeey: Consider wrapping this in `SpecifyData`. As we have found recently in #879, there might be additional information users want from ray casts in the future.

azeey (commented):
Thank you so much for this great contribution. I just have a couple of small comments. 

I do appreciate your patience. There's just a lot of PRs in the queue these days. If you could consider reviewing other [open PRs](https://github.com/search?q=org%3Agazebosim&type=pullrequests&state=open) in the mean time, that would be super helpful.

test/common_test/simulation_features.cc
```diff
@@ -2275,6 +2276,201 @@ TYPED_TEST(SimulationFeaturesRayIntersectionTest, UnsupportedRayIntersections)
...
+      EXPECT_TRUE(hit.hit);
+      double epsilon = 1e-3;
+      EXPECT_TRUE(hit.point.isApprox(Eigen::Vector3d(-1, 0, 2), epsilon))
+        << "hit point: " << hit.point.transpose();
+      EXPECT_TRUE(hit.normal.isApprox(Eigen::Vector3d(-1, 0, 0), epsilon))
+        << "hit normal: " << hit.normal.transpose();
+      EXPECT_DOUBLE_EQ(0.25, hit.fraction);
+    }
+
+    // Ray 1 — misses; hit must be false, numeric fields NaN (REP-117)
```
scpeters: by my reading of [REP-117](https://reps.openrobotics.org/rep-0117/), a `NaN` is returned when there is an erroneous measurement, but the comment on this line refers to a correct measurement of no objects in range, so I would expect these APIs to return `+INF`

include/gz/physics/GetBatchRayIntersection.hh
```diff
@@ -0,0 +1,104 @@
...
+{
+  public: template <typename PolicyT>
+  struct RayIntersectionT
+  {
+    public: using Scalar = typename PolicyT::Scalar;
+    public: using VectorType =
+      typename FromPolicy<PolicyT>::template Use<LinearVector>;
+
+    /// \brief True if the ray intersected an object; false on a miss.
+    /// When false, point, normal, and fraction carry no meaningful value.
```
scpeters: there is a reference to [REP-117](https://reps.openrobotics.org/rep-0117/) in the comments of the test and numerous uses of `NaN` in the implementation. As noted in a separate comment, I think `+INF` is more appropriate than `NaN` and recommend referencing the REP and this choice in this API documentation

include/gz/physics/GetBatchRayIntersection.hh
apojomovsky: Happy to revisit if needed, but I'd like to push back a bit on wrapping batch results in `SpecifyData`. The batch API is designed specifically to reduce per-ray dispatch overhead, and `CompositeData` involves a heap allocation per result object, so a `vector<RayIntersectionData>` would mean O(N) allocations per step. For the single-ray API this is totally fine, but it works against the point of the batch call. If per-ray extra data is needed later, I'd rather add it directly to `RayIntersectionT` or a separate opt-in API. Let me know if you disagree though.

include/gz/physics/GetBatchRayIntersection.hh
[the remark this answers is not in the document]
apojomovsky: Done, `RayIntersectionT` now lives in `include/gz/physics/RayIntersection.hh` and both features alias it. On `bool hit`: we intentionally left it out. Based on @scpeters' suggestion we adopted the REP-117 convention where misses return `+INF` for `fraction`, so `hit` is just `!std::isinf(fraction)` and adds no information. Happy to add it back if you feel it helps ergonomics, just wanted to flag the redundancy first.

test/common_test/simulation_features.cc
apojomovsky: Done, thanks!

include/gz/physics/GetBatchRayIntersection.hh
apojomovsky: Done, thanks!

dartsim/src/SimulationFeatures.cc
apojomovsky: Done, thanks!

include/gz/physics/GetBatchRayIntersection.hh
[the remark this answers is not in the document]
j-rivero: > `hit` is just `!std::isinf(fraction)` and adds no information. Happy to add it back if you feel it helps ergonomics, just wanted to flag the redundancy first.

Does this mean that the consumer code needs to check hits comparing against `!std::isinf(fraction)` ?  Seems error-prone to my eyes if that is the case.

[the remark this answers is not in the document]
apojomovsky: > > `hit` is just `!std::isinf(fraction)` and adds no information. Happy to add it back if you feel it helps ergonomics, just wanted to flag the redundancy first.
> 
> Does this mean that the consumer code needs to check hits comparing against `!std::isinf(fraction)` ? Seems error-prone to my eyes if that is the case.

If I understand correctly, the concern is that checking hits with `!std::isinf(fraction)` is error-prone? From what I can tell, REP-117 covers this case: "out of range detections will be represented by +Inf" and "erroneous detections shall be represented by quiet NaNs", so `std::isfinite(fraction)` should be the right check here. That said, I can see how it feels a bit unexpected if you're coming from the physics API side rather than the sensor side.

One option I was thinking about is adding a small helper to the struct:

```cpp
bool IsHit() const { return std::isfinite(fraction); }
```

That way there's no risk of `hit` and `fraction` diverging. What do you think?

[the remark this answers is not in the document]
j-rivero: > If I understand correctly, the concern is that checking hits with `!std::isinf(fraction)` is error-prone?

If that is the kind of comparison that a consumer code needs to perform I think that we would in risk of doing a different unsafe comparison instead.



> o `std::isfinite(fraction)` should be the right check here. That said, I can see how it feels a bit unexpected if you're coming from the physics API side rather than the sensor side.
> 
> One option I was thinking about is adding a small helper to the struct:
> 
> ```c++
> bool IsHit() const { return std::isfinite(fraction); }
> ```
> 
> That way there's no risk of `hit` and `fraction` diverging. What do you think?

+1 . The helper function provides an standard way to check for hits encapsulating the details of that check.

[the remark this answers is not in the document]
apojomovsky: Thanks for the feedback, your suggestion was addressed in https://github.com/gazebosim/gz-physics/pull/880/commits/3c9d32ec0381e42b7b3117b69910928c339520f0

j-rivero:
Thanks @apojomovsky ! good again to my eyes

include/gz/physics/GetBatchRayIntersection.hh
azeey: That's a good observation. However, we can still make it work. The type can be `using "BatchedRayIntersectionData = SpecifyData<RequiredData<std::vector<RayIntersection>>>`. That way, it's just one allocation, not O(N). Also, with the current API, we are allocating the std::vector in every call. I think using an output parameter would allow the caller to preallocate and reuse an output parameter

so, we would have
```c++
    public: bool
      GetBatchRayIntersectionFromLastStep(
        const std::vector<RayQuery> &_rays, BatchedRayIntersectionData &_output) const;
```

I know the implementation right now in gz-sim requires copying the data over to a gz-sim native data type, but we could store the `BatchedRayIntersectionData` from one call in a custom gz-sim component and reuse it in the next call.

apojomovsky: My apologies for the delay here @azeey . The notification got lost in my inbox for some reason.

Switched to the output parameter approach with `bool` return, thanks for the suggestion. I used plain `std::vector<RayIntersection> &` for the output type rather than `BatchedRayIntersectionData`, since `RayIntersectionT` is a plain struct and the vector is already a single allocation.

Please let me know if you'd prefer the `SpecifyData` wrapper anyway, for example if you have the component storage use case in mind as a near-term follow-up.

Changes: gz-physics https://github.com/apojomovsky/gz-physics/commit/f1ba9285baabca21cdd0188fa755c399b16e1afe, gz-sim https://github.com/apojomovsky/gz-sim/commit/525d4ad19b21a40f24af630974b884e9976d8bd9.

dartsim/src/SimulationFeatures.cc
```diff
@@ -242,17 +245,53 @@ SimulationFeatures::GetRayIntersectionFromLastStep(
...
+  auto *gzDetector =
+    dynamic_cast<dart::collision::GzCollisionDetector *>(detector.get());
+
+  if (gzDetector)
+  {
+    auto gzResults = gzDetector->BatchRaycast(
+        solver->getCollisionGroup().get(), _rays);
+    if (gzResults)
+    {
+      _output = std::move(*gzResults);
```
azeey: This will still incur allocations in each step. Can we pass `_output` into `BatchRaycast` instead?

include/gz/physics/GetBatchRayIntersection.hh
azeey: I'd still prefer if we used `SpecifyData` unless we have a strong reason not to. It gives us thea ability to add more data to the output in the future without breaking API/ABI.

azeey (commented):
Thanks for addressing the feedback. However, I think there's still one more optimization we can make.

dartsim/src/SimulationFeatures.cc
apojomovsky: Both points are addressed in the updated commit. Matching changes were accordingly added for the gz-sim branch too.

**Output parameter**: `BatchRaycast` now takes `std::vector<GzRayResult> &_output` directly. `SimulationFeatures.cc` passes `_output` straight in, so the vector is cleared and filled in place with no intermediate allocation or move.

**SpecifyData** (addresses [this](https://github.com/gazebosim/gz-physics/pull/880#discussion_r3156505246) comment): The public `World<>` API now uses `BatchedRayIntersectionData = SpecifyData<RequireData<std::vector<RayIntersection>>>` as the output type, keeping the extensibility you described.

include/gz/physics/GetBatchRayIntersection.hh
```diff
@@ -0,0 +1,94 @@
...
+  public: template <typename PolicyT>
+  class Implementation : public virtual Feature::Implementation<PolicyT>
+  {
+    public: using RayIntersection = RayIntersectionT<PolicyT>;
+    public: using RayQuery = RayT<PolicyT>;
+
+    public: virtual bool GetBatchRayIntersectionFromLastStep(
+        const Identity &_worldID,
+        const std::vector<RayQuery> &_rays,
+        std::vector<RayIntersection> &_output) const = 0;
```
azeey: `GetBatchRayIntersectionFromLastStep` is now part of the public API from the physics engine plugin implementer's perspective. We should use `BatchedRayIntersectionData` here as well for `_output`.

apojomovsky: My apologies, I had pushed before completing this last round of changes. The updated commits are now pushed.

In `Implementation<>`, `GetBatchRayIntersectionFromLastStep` now takes `BatchedRayIntersectionData &_output` instead of a plain `std::vector<RayIntersection> &`. The type alias was lifted to the feature level (`BatchedRayIntersectionDataT<PolicyT>`) so both `World<>` and `Implementation<>` share the same definition. The dispatch header now passes `_output` directly without unwrapping.

dartsim/src/GzCollisionDetector.cc
```diff
@@ -193,3 +246,58 @@ bool GzBulletCollisionDetector::collide(
...
+      static_cast<btScalar>(ray.origin.z()));
+    const btVector3 btTo(
+      static_cast<btScalar>(ray.target.x()),
+      static_cast<btScalar>(ray.target.y()),
+      static_cast<btScalar>(ray.target.z()));
+
+    btCollisionWorld::ClosestRayResultCallback rayCallback(btFrom, btTo);
+    btWorld->rayTest(btFrom, btTo, rayCallback);
+
+    GzRayResult result;
```
azeey: A small optimization (avoids an extra copy), but this is in the hot loop, so it might be significant:

 Change this to `GzRayResult &result = _output.emplace_back();` and remove the 
`_output.push_back(std::move(result));` below.

azeey: https://github.com/gazebosim/gz-physics/pull/880/commits/467de911a154216dee17b6c5297273e08f03811d

azeey (approved):
This is a great contribution! I really appreciate your patience in iterating on the feedback. From my initial test, this significantly outperforms the GPU lidar for simple shapes (haven't tested meshes yet), so it's a very valuable contribution.

j-rivero:
Great work @apojomovsky !

azeey:
I'd like to backport this and #976 to older versions, but as documented in #986, this has a breaking change due to the `nan`->`inf` update. @apojomovsky could you please create a version of this PR on gz-physics9 that does not have the breaking change? cc @jmachowinski

apojomovsky:
> I'd like to backport this and #976 to older versions, but as documented in #986, this has a breaking change due to the `nan`->`inf` update. @apojomovsky could you please create a version of this PR on gz-physics9 that does not have the breaking change? cc @jmachowinski

Hey @azeey ! I just created 3 PRs with the respective backports of the pieces needed for this feature on Jetty, please find them below:

- https://github.com/gazebosim/gz-physics/pull/1027
- https://github.com/gazebosim/gz-sensors/pull/637
- https://github.com/gazebosim/gz-sim/pull/3854/

Also worth noting that https://github.com/gazebosim/gz-physics/pull/1027 already includes the backport of https://github.com/gazebosim/gz-physics/pull/976 . FYI @jmachowinski

azeey:
Awesome! I'll take a look at those

iche033:
The description of this PR seems to have changed to say this PR is a backport, targeted at the Jetty branch. I think that's not intended?

apojomovsky:
> The description of this PR seems to have changed to say this PR is a backport, targeted at the Jetty branch. I think that's not intended?

My apologies @iche033 , looks like I edited the wrong PR description while I was in the process of opening the [actual backport PR](https://github.com/gazebosim/gz-physics/pull/1027) . Original description is restored now. I'm sorry for the noise.

PR #902, opened 2026-03-17
Use rotary aliases packages for main CI

j-rivero:
Part of https://github.com/gazebo-tooling/release-tools/issues/1446


azeey (changes requested):
The packages should start with lib

PR #931, opened 2026-03-31
mujoco: Add support for collision bitmask and category bitmask

### LINKED ISSUE #299: MuJoCo physics engine plugin

iche033:
# 🎉 New feature

## Summary

Adds support for category bitmask for the mujoco physics engine. MuJoco has built-in support for collision filtering based using a geom's `conaffinity` and `contype` parameters , which map well to `collide_bitmask` and `category_bitmask` respectively.

When the user only sets collide bitmasks and does not set the category bitmask, the category bitmasks will default to the same value as the collide bitmasks. This is consistent with the behavior in bullet-featherstone (#884) and dartsim (#885), which follow the strategy outlined in https://github.com/gazebosim/sdformat/pull/1630. 

## Test it

```
ctest -R simulation_features_mujoco
```

## Checklist
- [ ] Added tests

Generated-by:  Gemini 3


mujoco/src/EntityManagementFeatures.cc
```diff
@@ -339,6 +340,125 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+}
+
+/////////////////////////////////////////////////
+void EntityManagementFeatures::RemoveCategoryFilterMask(
+    const Identity &_shapeID)
+{
+  auto shapeInfo = this->ReferenceInterface<ShapeInfo>(_shapeID);
+  if (shapeInfo && shapeInfo->geom)
+  {
+    shapeInfo->categoryMask.reset();
```
j-rivero: This reset / set / reset sequence is a bit weird to my eyes. Can this first reset be removed?

mujoco/src/SDFFeatures.cc
```diff
@@ -330,14 +331,44 @@ struct ModelKinematicStructure
...
+        geom->conaffinity = static_cast<int>(conaffinity);
         mjs_setName(geom->element,
                     ::sdf::JoinName(body_name, collision->Name()).c_str());
         auto shapeInfo =
             std::make_shared<ShapeInfo>(_base.GetNextEntity(), linkInfo);
         shapeInfo->geom = geom;
         shapeInfo->name = collision->Name();
+        if (contypeOpt.has_value())
+        {
+          shapeInfo->categoryMask = *contypeOpt;
```
j-rivero: Agent hint from clang-tidy:
shapeInfo->categoryMask = *contypeOpt dereferences an optional only to re-wrap the value into another optional. Clang-tidy flags this as  bugprone-optional-value-conversion. Fix: shapeInfo->categoryMask = contypeOpt; (assign the optional directly).

mujoco/src/EntityManagementFeatures.hh
```diff
@@ -40,18 +40,34 @@ struct EntityManagementFeatureList : FeatureList<
...
+  CategoryFilterMaskFeature
 > { };
 
 class EntityManagementFeatures :
     public virtual Base,
     public virtual Implements3d<EntityManagementFeatureList>
 {
+  // ----- CollisionFilterMaskFeature -----
+  public: void SetCollisionFilterMask(
+      const Identity &_shapeID, uint16_t _mask) override;
```
j-rivero, line 56: static analysis report a missing cstdint include in this file for using uint16_t/uint32_t in this header file and SDFeatures.cc

j-rivero (commented):
Testing agent was complaining by the lack of tests for this feature. I asked it to add them to give a try but maybe worth including in this PR? https://github.com/gazebosim/gz-physics/pull/new/jrivero/mujoco_col_bitmask_tests .

mujoco/src/EntityManagementFeatures.hh
iche033: added include. [cf1bbb8](https://github.com/gazebosim/gz-physics/pull/931/commits/cf1bbb83fd9f2d72443f29a3c180c33657808460)

mujoco/src/SDFFeatures.cc
iche033: fixed. [cf1bbb8](https://github.com/gazebosim/gz-physics/pull/931/commits/cf1bbb83fd9f2d72443f29a3c180c33657808460)

mujoco/src/EntityManagementFeatures.cc
iche033: reset is not needed, removed. [cf1bbb8](https://github.com/gazebosim/gz-physics/pull/931/commits/cf1bbb83fd9f2d72443f29a3c180c33657808460)

iche033:
> Testing agent was complaining by the lack of tests for this feature. I asked it to add them to give a try but maybe worth including in this PR? https://github.com/gazebosim/gz-physics/pull/new/jrivero/mujoco_col_bitmask_tests .

Thanks! I adapted the unique test cases and added two new tests in `test/common_test/simulation_features.cc` so that they can be run for all engines.  [b6808db](https://github.com/gazebosim/gz-physics/pull/931/commits/b6808dbec5d1a9d15f3432ed4906194d5c248915)

mujoco/src/EntityManagementFeatures.cc
```diff
@@ -339,6 +340,124 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+  }
+  return std::numeric_limits<uint16_t>::max();
+}
+
+/////////////////////////////////////////////////
+void EntityManagementFeatures::RemoveCategoryFilterMask(
+    const Identity &_shapeID)
+{
+  auto shapeInfo = this->ReferenceInterface<ShapeInfo>(_shapeID);
+  if (shapeInfo && shapeInfo->geom)
```
j-rivero: This block seems a bit weird, set the category filter mask to just reset afterwards.

Maybe something like Bullet-feathrestone https://github.com/gazebosim/gz-physics/blob/50971dc461c3d62f59c33e37caca2ab7e3641fe1/bullet-featherstone/src/EntityManagementFeatures.cc#L269-L279 would be cleaner, if possible:

```c++
  void EntityManagementFeatures::RemoveCategoryFilterMask(const Identity &_shapeID) {
    auto shapeInfo = this->ReferenceInterface<ShapeInfo>(_shapeID);
    if (!shapeInfo || !shapeInfo->geom) return;
    shapeInfo->categoryMask.reset();
    shapeInfo->geom->contype = shapeInfo->geom->conaffinity;
    // mirror update to mjModelObj->geom_contype[geomId]
  }
```

mujoco/src/EntityManagementFeatures.cc
```diff
@@ -339,6 +340,124 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+/////////////////////////////////////////////////
+uint16_t EntityManagementFeatures::GetCollisionFilterMask(
+    const Identity &_shapeID) const
+{
+  auto shapeInfo = this->ReferenceInterface<ShapeInfo>(_shapeID);
+  if (shapeInfo && shapeInfo->geom)
+  {
+    return static_cast<uint16_t>(shapeInfo->geom->conaffinity);
+  }
+  return std::numeric_limits<uint16_t>::max();
```
j-rivero: Agent is complaing here about this return in the case of no shapeInfo:

```
  0xFFFF is a valid, commonly-used bitmask (it's the "no filtering" default). Callers cannot distinguish "lookup failed" from "no filtering". Combined with setters that silently do nothing on the same condition, a misidentified shape will look like a normal unfiltered shape. ReferenceInterface
   normally asserts, so the shapeInfo && shapeInfo->geom check is likely dead — either drop the check and let the assert fire, or at minimum log via gzerr so bugs are visible. Bullet-featherstone just dereferences directly (no guard), which is the correct pattern her
```

j-rivero (approved):
Minor suggestion on semantics. Looks good, maybe worth adding a changelog entry.

mujoco/src/EntityManagementFeatures.cc
iche033: removed check [08eb845](https://github.com/gazebosim/gz-physics/pull/931/commits/08eb84537566855d8f40458dc0d2b609ba71e20f) so it's consistent with other places in gz-physics.

mujoco/src/EntityManagementFeatures.cc
iche033: I added more comments to explain this.  Instead of just setting the `contype` to `conaffinity` (`shapeInfo->geom->contype = shapeInfo->geom->conaffinity;`) which only updates the spec, we also need to update the model. The test catches this and should fail if only the spec is set. So I called `SetCategoryBitmask` for convenience which updates both the spec and the model. This saves having some duplicate code in the function. The caveat is that we need to reset the optional `categoryMask ` variable afterwards. Comment added in [08eb845](https://github.com/gazebosim/gz-physics/pull/931/commits/08eb84537566855d8f40458dc0d2b609ba71e20f).

mujoco/src/EntityManagementFeatures.cc
```diff
@@ -406,6 +407,121 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+    // as well
+    if (!shapeInfo->categoryMask.has_value())
+    {
+      shapeInfo->geom->contype = static_cast<int>(_mask);
+    }
+
+    // Update the model directly for immediate effect
+    if (auto linkInfo = shapeInfo->linkInfo.lock())
+    {
+      if (linkInfo->worldInfo)
```
azeey: `worldInfo` is now stored in `shapeInfo`.

mujoco/src/EntityManagementFeatures.cc
```diff
@@ -406,6 +407,121 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+  auto shapeInfo = this->ReferenceInterface<ShapeInfo>(_shapeID);
+  if (shapeInfo && shapeInfo->geom)
+  {
+    shapeInfo->categoryMask = _mask;
+    shapeInfo->geom->contype = static_cast<int>(_mask);
+
+    // Update the model directly for immediate effect
+    if (auto linkInfo = shapeInfo->linkInfo.lock())
+    {
+      if (linkInfo->worldInfo)
```
azeey: `worldInfo` is now stored in `shapeInfo`.

mujoco/src/EntityManagementFeatures.cc
```diff
@@ -406,6 +407,121 @@ Identity EntityManagementFeatures::GetLinkOfShape(
...
+    {
+      shapeInfo->geom->contype = static_cast<int>(_mask);
+    }
+
+    // Update the model directly for immediate effect
+    if (auto linkInfo = shapeInfo->linkInfo.lock())
+    {
+      if (linkInfo->worldInfo)
+      {
+        linkInfo->worldInfo->specDirty = true;
```
azeey: Do we need to mark the spec dirty since we're able to set the model directly? I would think there's no harm in just updating the spec but requiring a recompilation.

mujoco/src/SDFFeatures.cc
```diff
@@ -504,8 +504,34 @@ struct ModelKinematicStructure
...
+              contypeOpt = static_cast<uint16_t>(
+                  contactElem->Get<uint32_t>("category_bitmask"));
+            }
+            if (contactElem->HasElement("collide_bitmask"))
+            {
+              conaffinity = static_cast<uint16_t>(
+                  contactElem->Get<uint32_t>("collide_bitmask"));
+            }
+          }
+        }
```
azeey: nit: Using `sdf::Element::FindElement` would be cleaner.

mujoco/src/SDFFeatures.cc
```diff
@@ -504,8 +504,34 @@ struct ModelKinematicStructure
...
+
+        uint16_t contype = contypeOpt.has_value() ? *contypeOpt : conaffinity;
```
azeey: nit: 
```suggestion
        uint16_t contype = contypeOpt.value_or(conaffinity);
```

mujoco/src/SDFFeatures.cc
iche033: updated to use FindElement. [c0b7a42](https://github.com/gazebosim/gz-physics/pull/931/commits/c0b7a42007be2d4863953b4b9930e6b0af6942c2)

mujoco/src/SDFFeatures.cc
iche033: [c0b7a42](https://github.com/gazebosim/gz-physics/pull/931/commits/c0b7a42007be2d4863953b4b9930e6b0af6942c2)

mujoco/src/EntityManagementFeatures.cc
iche033: updated to get `worldInfo` from `shapeInfo`. [c0b7a42](https://github.com/gazebosim/gz-physics/pull/931/commits/c0b7a42007be2d4863953b4b9930e6b0af6942c2)

mujoco/src/EntityManagementFeatures.cc
iche033: updated to get `worldInfo` from `shapeInfo`. [c0b7a42](https://github.com/gazebosim/gz-physics/pull/931/commits/c0b7a42007be2d4863953b4b9930e6b0af6942c2)

iche033: actually, removed these calls in [5e5dc6f](https://github.com/gazebosim/gz-physics/pull/931/commits/5e5dc6f9905665a2211fe3fbb243931d9efa2d64) in favor of `specDirty = true`

mujoco/src/EntityManagementFeatures.cc
iche033: actually, removed these calls in [5e5dc6f](https://github.com/gazebosim/gz-physics/pull/931/commits/5e5dc6f9905665a2211fe3fbb243931d9efa2d64) in favor of `specDirty = true`

mujoco/src/EntityManagementFeatures.cc
iche033: hmm ran into a test failure before if I didn't update the model right afterwards but I couldn't reproduce it any more. So updated code to just set `specDirty = true` and removed code for updating the model directly. [5e5dc6f](https://github.com/gazebosim/gz-physics/pull/931/commits/5e5dc6f9905665a2211fe3fbb243931d9efa2d64)

azeey: Oh, I'm sorry. I meant to say we would update the spec **and** the model, but leave `specDirty = false`. This way, the spec has the correct information, but wouldn't cause a recompilation, which is an expensive operation.

iche033: ah I see, reverted my last commit and it should now update the model and leave `specDirty` untouched [305f61e](https://github.com/gazebosim/gz-physics/pull/931/commits/305f61e3a4fa14ba95019c15ca5394efef343309)

azeey (approved):

PR #919, opened 2026-03-25
Model: Enable/disable: gravity and static

ahcorde:
# 🎉 New feature

## Summary

Replace https://github.com/gazebosim/gz-physics/pull/764

Allow to enable/disable gravity and static state

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Claude Sonnet 4.6

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

PR #970, opened 2026-05-01
Refactor Gravity feature

iche033:
# 🎉 New feature

Targets https://github.com/gazebosim/gz-physics/pull/919

## Summary

Refactored so that both `Model` and `Link` both have a `SetGravityEnabled` feature. The one in `Model` does not have an engine-specific implementation. Instead it just calls the functions from `Link`.

I also updated the test's comments to be less dart specific

## Test it

```
ctest -R model_features
```

## Checklist
- [ ] Added tests

Generated-by: Gemini 3

### DIFF
--- a/dartsim/src/ModelFeatures.cc
+++ b/dartsim/src/ModelFeatures.cc
@@ -62,6 +62,72 @@ bool ModelFeatures::GetModelStatic(const Identity &_id) const
   return !modelInfo->model->isMobile();
 }
 
+/////////////////////////////////////////////////
+void ModelFeatures::SetModelGravityEnabled(
+    const Identity &_id, bool _enabled)
+{
+  auto modelInfo = this->GetModelInfo(_id);
+  if (!modelInfo || !modelInfo->model)
+    return;
+
+  for (const auto &linkInfo : modelInfo->links)
+  {
+    if (!linkInfo || !linkInfo->link)
+      continue;
+
+    // Added-mass links must keep DART gravity disabled
+    if (linkInfo->inertial.has_value() &&
+        linkInfo->inertial->FluidAddedMass().has_value())
+    {
+      continue;
+    }
+
+    linkInfo->link->setGravityMode(_enabled);
+  }
+
+  // Also apply to nested models recursively.
+  for (const std::size_t nestedID : modelInfo->nestedModels)
+  {
+    this->SetModelGravityEnabled(this->GenerateIdentity(
+        nestedID, this->GetModelInfo(nestedID)), _enabled);
+  }
+}
+
+/////////////////////////////////////////////////
+bool ModelFeatures::GetModelGravityEnabled(const Identity &_id) const
+{
+  const auto modelInfo = this->GetModelInfo(_id);
+  if (!modelInfo || !modelInfo->model)
+    return false;
+
+  for (const auto &linkInfo : modelInfo->links)
+  {
+    if (!linkInfo || !linkInfo->link)
+      continue;
+
+    // Added-mass links have DART gravity forcibly disabled; exclude them
+    if (linkInfo->inertial.has_value() &&
+        linkInfo->inertial->FluidAddedMass().has_value())
+    {
+      continue;
+    }
+
+    if (!linkInfo->link->getGravityMode())
+      return false;
+  }
+
+  // Also check nested models.
+  for (const std::size_t nestedID : modelInfo->nestedModels)
+  {
+    if (!this->GetModelGravityEnabled(this->GenerateIdentity(
+        nestedID, this->GetModelInfo(nestedID))))
+    {
+      return false;
+    }
+  }
+
+  return true;
+}
 
 }
 }
--- a/dartsim/src/ModelFeatures.cc
+++ b/dartsim/src/ModelFeatures.cc
@@ -28,22 +28,16 @@ void ModelFeatures::SetModelStatic(const Identity &_id, bool _static)
   if (!modelInfo || !modelInfo->model)
     return;
   auto skeleton = modelInfo->model;
-  const bool wasStatic = !skeleton->isMobile();
 
   // In DART, isMobile=false means the skeleton is fixed (static).
   skeleton->setMobile(!_static);
 
-  // When transitioning from static (kinematic) to dynamic, zero out any
+  // When transitioning between states, zero out any
   // velocities and accelerations that DART may have stored on the skeleton
-  // while it was kinematic. Without this the stored state is released as
-  // momentum when the body becomes dynamic.
-  if (wasStatic && !_static)
-  {
-    skeleton->resetVelocities();
-    skeleton->resetAccelerations();
-    skeleton->clearExternalForces();
-    skeleton->clearInternalForces();
-  }
+  skeleton->resetVelocities();
+  skeleton->resetAccelerations();
+  skeleton->clearExternalForces();
+  skeleton->clearInternalForces();
 
   // Also apply to nested models recursively.
   for (const std::size_t nestedID : modelInfo->nestedModels)
--- a/dartsim/src/ModelFeatures.cc
+++ b/dartsim/src/ModelFeatures.cc
@@ -62,52 +62,6 @@ bool ModelFeatures::GetModelStatic(const Identity &_id) const
   return !modelInfo->model->isMobile();
 }
 
-/////////////////////////////////////////////////
-void ModelFeatures::SetModelGravityEnabled(const Identity &_id, bool _enabled)
-{
-  auto modelInfo = this->GetModelInfo(_id);
-  if (!modelInfo || !modelInfo->model)
-    return;
-  for (const auto &linkInfo : modelInfo->links)
-  {
-    if (!linkInfo || !linkInfo->link)
-      continue;
-    // Added-mass links must keep DART gravity disabled: gravity is applied
-    // manually as F=ma each step. Let SetLinkAddedMass own that flag.
-    if (linkInfo->inertial.has_value() &&
-        linkInfo->inertial->FluidAddedMass().has_value())
-      continue;
-    linkInfo->link->setGravityMode(_enabled);
-  }
-
-  // Also apply to nested models recursively.
-  for (const std::size_t nestedID : modelInfo->nestedModels)
-  {
-    this->SetModelGravityEnabled(this->GenerateIdentity(
-        nestedID, this->GetModelInfo(nestedID)), _enabled);
-  }
-}
-
-/////////////////////////////////////////////////
-bool ModelFeatures::GetModelGravityEnabled(const Identity &_id) const
-{
-  const auto modelInfo = this->GetModelInfo(_id);
-  if (!modelInfo || !modelInfo->model)
-    return true;
-  for (const auto &linkInfo : modelInfo->links)
-  {
-    if (!linkInfo || !linkInfo->link)
-      continue;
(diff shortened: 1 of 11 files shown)


include/gz/physics/Gravity.hh
```diff
@@ -0,0 +1,97 @@
...
+        public: virtual bool GetLinkGravityEnabled(
+            const Identity &_id) const = 0;
+      };
+    };
+
+    /////////////////////////////////////////////////
+    using ModelGravityEnabledRequiredFeatures = FeatureList<
+      LinkGravityEnabled,
+      GetLinkFromModel,
+      GetNestedModelFromModel>;
```
azeey: It's better to use a struct for faster compiler times

include/gz/physics/Gravity.hh
```diff
@@ -0,0 +1,97 @@
...
+    /// and all its child entities.
+    class GZ_PHYSICS_VISIBLE ModelGravityEnabled
+        : public virtual
+          FeatureWithRequirements<ModelGravityEnabledRequiredFeatures>
+    {
+      /// \brief The Model API for getting and setting gravity mode.
+      public: template <typename PolicyT, typename FeaturesT>
+      class Model
+          : public virtual GetLinkFromModel::Model<PolicyT, FeaturesT>,
+            public virtual GetNestedModelFromModel::Model<PolicyT, FeaturesT>
```
azeey: I don't think you need to inherit from these. I see a similar example in BoundingBox, but think it's better to rely on calling `Interface<ModelGravityEnabledRequiredFeatures>()` in the implementation functions to get acccess to the needed functions from `GetLinkFromModel` and `GetNestedModelFromModel`.

include/gz/physics/detail/Gravity.hh
```diff
@@ -0,0 +1,100 @@
...
+  return this->template Interface<LinkGravityEnabled>()
+      ->GetLinkGravityEnabled(this->identity);
+}
+
+/////////////////////////////////////////////////
+template <typename PolicyT, typename FeaturesT>
+void ModelGravityEnabled::Model<PolicyT, FeaturesT>::SetGravityEnabled(
+    bool _enabled)
+{
+  const std::size_t linkCount = this->GetLinkCount();
```
azeey: Per my previous comment, you could have
```suggestion
  auto * reqIface = this->template Interface<ModelGravityEnabledRequiredFeatures>();
  const std::size_t linkCount = reqIface->GetLinkCount();
```

include/gz/physics/detail/Gravity.hh
```diff
@@ -0,0 +1,100 @@
...
+  for (std::size_t i = 0; i < linkCount; ++i)
```
azeey: I'm actually not sure whether we want to do this here or let each physics engine handle the model API. For example, the engine could have a more efficient way to disable gravity on all links.

include/gz/physics/detail/Gravity.hh
iche033: updated in [3e2b686](https://github.com/gazebosim/gz-physics/pull/970/commits/3e2b686c45cfdeeb85da14e42a110eaf37003af7)

include/gz/physics/Gravity.hh
iche033: removed inheritance and updated implementations, [3e2b686](https://github.com/gazebosim/gz-physics/pull/970/commits/3e2b686c45cfdeeb85da14e42a110eaf37003af7)

include/gz/physics/Gravity.hh
iche033: done. [3e2b686](https://github.com/gazebosim/gz-physics/pull/970/commits/3e2b686c45cfdeeb85da14e42a110eaf37003af7)

include/gz/physics/detail/Gravity.hh
iche033: Ok one option is just to remove this logic here and just call the physics engine's implementation (like it was before). 

As for dartsim implementation, I do see a [skeleton->setGravity(..)](https://dart.readthedocs.io/en/latest/cpp-api/d3/d19/classdart_1_1dynamics_1_1Skeleton.html#a58a52b805d73dce9d2174889d86b0d8a) API. However, we have specific logic for handling added mass when setting gravity, and that requires looping through each link. So for dartsim implementation specifically, I think we'll still be required to iterate over the links. For other physics engines that dont' support added mass yet, it maybe more straight forward

azeey: Yes, that's what I was thinking. Just call the physics engine's implementation and let it handle it in the most efficient way specific to the engine. For dartsim, like you said, we have to loop through each link, but that may not be the case for others.

iche033: ok updated accordingly in [c20030e](https://github.com/gazebosim/gz-physics/pull/970/commits/c20030e30e41334ce32e30f02ec9af3c82970a56) by moving implementation back to the dartsim engine.

include/gz/physics/Gravity.hh
```diff
@@ -0,0 +1,103 @@
...
+#include <gz/physics/FeatureList.hh>
+#include <gz/physics/GetEntities.hh>
+
+namespace gz
+{
+  namespace physics
+  {
+    /////////////////////////////////////////////////
+    /// \brief Feature for getting and setting whether gravity affects a link.
+    class GZ_PHYSICS_VISIBLE LinkGravityEnabled : public virtual Feature
```
azeey: Do we need a separate feature? I was thinking the link and model APIs will be under one feature.
```c++
class GravityEnabled: public virtual Feature
{
      public: template <typename PolicyT, typename FeaturesT>
      class Model : public virtual Feature::Model<PolicyT, FeaturesT>
      {
        /// \brief Set whether gravity is enabled for this model and its links.
        /// \param[in] _enabled True to enable gravity, false to disable it.
        public: void SetGravityEnabled(bool _enabled);

        /// \brief Get whether gravity is enabled for this model.
        /// \return True if gravity is enabled for all entities, false
        /// otherwise.
        public: bool GetGravityEnabled() const;
      };

      public: template <typename PolicyT, typename FeaturesT>
      class Link : public virtual Feature::Link<PolicyT, FeaturesT>
      {
        /// \brief Set whether gravity is enabled for this link.
        /// \param[in] _enabled True to enable gravity, false to disable it.
        public: void SetGravityEnabled(bool _enabled);

        /// \brief Get whether gravity is enabled for this link.
        /// \return True if gravity is enabled, false otherwise.
        public: bool GetGravityEnabled() const;
      };
}
```

We could even include `World`.

iche033: consolidated the features in [ad624a9](https://github.com/gazebosim/gz-physics/pull/970/commits/ad624a91d379fe86173b597f43c58269d0d68b40).

> We could even include World.

sure let's do this in a separate PR.

ahcorde (approved):

dartsim/src/ModelFeatures.cc
```diff
@@ -63,20 +63,26 @@ bool ModelFeatures::GetModelStatic(const Identity &_id) const
...
+
     // Added-mass links must keep DART gravity disabled: gravity is applied
     // manually as F=ma each step. Let SetLinkAddedMass own that flag.
     if (linkInfo->inertial.has_value() &&
         linkInfo->inertial->FluidAddedMass().has_value())
+    {
       continue;
+    }
+
     linkInfo->link->setGravityMode(_enabled);
```
j-rivero: Instead of doing the real operation in link here with `link->setGravityMode` shouldn't we be using the `SetLinkGravityEnabled` function of each link to delegate the operation and reduce duplication (that method checks for link validation, handled the FluidAddedMass and set the gravity mode.

j-rivero (commented):
One minor suggestion to de-duplicate code

dartsim/src/ModelFeatures.cc
iche033: good idea, I refactored the code to a couple of helper functions for setting and getting gravity. [0edcd52](https://github.com/gazebosim/gz-physics/pull/970/commits/0edcd520f7c552cb4c9d6c755f23d50638d1efb3)

dartsim/src/ModelFeatures.cc
```diff
@@ -93,19 +99,35 @@ bool ModelFeatures::GetModelGravityEnabled(const Identity &_id) const
 {
   const auto modelInfo = this->GetModelInfo(_id);
   if (!modelInfo || !modelInfo->model)
-    return true;
+    return false;
```
azeey: I'm not sure if `false` is a better return value. Gravity is enabled by default, so `false` would indicate a deviation from the default.

include/gz/physics/Gravity.hh
```diff
@@ -0,0 +1,93 @@
...
+      /// \brief The Model API for getting and setting gravity mode.
+      public: template <typename PolicyT, typename FeaturesT>
+      class Model : public virtual Feature::Model<PolicyT, FeaturesT>
+      {
+        /// \brief Set whether gravity is enabled for this model and its links.
+        /// \param[in] _enabled True to enable gravity, false to disable it.
+        public: void SetGravityEnabled(bool _enabled);
+
+        /// \brief Get whether gravity is enabled for this model.
+        /// \return True if gravity is enabled for all entities, false
```
azeey:
```suggestion
        /// \return True if gravity is enabled for all child links and the links of nested models recursively, false
```

include/gz/physics/Gravity.hh
```diff
@@ -0,0 +1,93 @@
...
+        /// \brief Get whether gravity is enabled for this link.
+        /// \return True if gravity is enabled, false otherwise.
+        public: bool GetGravityEnabled() const;
+      };
+
+      /// \brief The Model API for getting and setting gravity mode.
+      public: template <typename PolicyT, typename FeaturesT>
+      class Model : public virtual Feature::Model<PolicyT, FeaturesT>
+      {
+        /// \brief Set whether gravity is enabled for this model and its links.
```
azeey: The semantics of this is a little confusing to me. The way it's currently implemented, gravity mode is not a state of a model (for example, like `<static>` is a state of a model), so it can not be set independently of the state of it's children. If I enable gravity on the model and then disable it on one of the linsk, `GetGravityEnabled` will return false. So, I'd recommend being explicit that it's just about the links. I was going to suggest changing the API to be `SetGravityEnabledOnAllLinks`, but that seems unwieldy.

```suggestion
        /// \brief Set whether gravity is enabled for all of this model's links and the links
                      of nested models recursively.
```

iche033: updated doc. [7640a92](https://github.com/gazebosim/gz-physics/pull/970/commits/7640a92cf2ed7b7fabca93e2bcb1c1d4c8edc4eb)

include/gz/physics/Gravity.hh
iche033: [7640a92](https://github.com/gazebosim/gz-physics/pull/970/commits/7640a92cf2ed7b7fabca93e2bcb1c1d4c8edc4eb)

dartsim/src/ModelFeatures.cc
iche033: ok I went back and forth on this one. We can make it consistent with default gravity setting. [7640a92](https://github.com/gazebosim/gz-physics/pull/970/commits/7640a92cf2ed7b7fabca93e2bcb1c1d4c8edc4eb)

j-rivero (approved):

azeey (approved):

PR #976, opened 2026-05-12
feat: Raycast implementation for ODE

jmachowinski:
# 🎉 New feature

Followup to https://github.com/gazebosim/gz-physics/pull/880

This adds support for ODE for the raycast implementation, and therefore to the CPU-Lidar implementation.

## Checklist
- [x ] Signed all commits for DCO
- [ ] Added tests
- [x ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

### DIFF
--- a/dartsim/src/GzCollisionDetector.cc
+++ b/dartsim/src/GzCollisionDetector.cc
@@ -244,7 +244,7 @@ void NearCallbackODE(void *_data, dGeomID _o1, dGeomID _o2)
     const double sqrDist = (*data->origin - contactPoint).squaredNorm();
     if(sqrDist > data->curContactDistSqr)
     {
-      // ignore nearer contacts
+      // ignore farther contacts
       return;
     }
 
@@ -257,6 +257,7 @@ void NearCallbackODE(void *_data, dGeomID _o1, dGeomID _o2)
     {
       setResult(data->result->mRayHits.front());
     }
+    data->curContactDistSqr = sqrDist;
   }
 }
 
--- a/dartsim/src/GzCollisionDetector.cc
+++ b/dartsim/src/GzCollisionDetector.cc
@@ -319,7 +319,7 @@ bool GzOdeCollisionDetector::raycast(
   auto odeGroup = static_cast<GzOdeCollisionGroup *>(group);
   const dSpaceID spaceId = odeGroup->getOdeSpaceId();
 
-  const dGeomID rayId = dCreateRay(spaceId, 1.0);
+  const dGeomID rayId = dCreateRay(nullptr, 1.0);
   dGeomRaySetClosestHit(rayId, 1);
 
   doSingleRaycastODE(from, to, result, rayId, spaceId);
@@ -338,7 +338,7 @@ bool GzOdeCollisionDetector::BatchRaycast(
   auto odeGroup = static_cast<GzOdeCollisionGroup *>(_group);
   const dSpaceID spaceId = odeGroup->getOdeSpaceId();
 
-  const dGeomID rayId = dCreateRay(spaceId, 1.0);
+  const dGeomID rayId = dCreateRay(nullptr, 1.0);
   dGeomRaySetClosestHit(rayId, 1);
 
   _results.clear();
--- a/dartsim/src/GzCollisionDetector.cc
+++ b/dartsim/src/GzCollisionDetector.cc
@@ -124,7 +124,6 @@ bool GzCollisionDetector::BatchRaycast(
 GzOdeCollisionDetector::GzOdeCollisionDetector()
   : OdeCollisionDetector(), GzCollisionDetector()
 {
-
 }
 
 /////////////////////////////////////////////////
@@ -186,6 +185,13 @@ bool GzOdeCollisionDetector::collide(
   return ret;
 }
 
+struct ODERayCastData
+{
+  RaycastResult* result;
+  const Eigen::Vector3d *origin;
+  double curContactDistSqr = std::numeric_limits<double>::max();
+};
+
 void NearCallbackODE(void *_data, dGeomID _o1, dGeomID _o2)
 {
   // Check space
@@ -218,10 +224,11 @@ void NearCallbackODE(void *_data, dGeomID _o1, dGeomID _o2)
 
   dContactGeom contact;
 
-  RaycastResult* result = static_cast<RaycastResult*>(_data);
+  ODERayCastData* data = static_cast<ODERayCastData*>(_data);
 
   auto setResult = [&](dart::collision::RayHit &rayHit)
   {
+
       auto geomData = dGeomGetData(other);
       rayHit.mNormal = Eigen::Vector3d(contact.normal);
       rayHit.mPoint = Eigen::Vector3d(contact.pos);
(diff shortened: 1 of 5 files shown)


azeey:
Thanks for the contribution! Could you please fill out the PR template?

jmachowinski:
codecheck` passed <- This throws pagewise unrelated warnings...

Note, for the tests, I needed to remove the unsupported 'banana' as this defaulted to ode, and was therefore supported.

dartsim/src/GzCollisionDetector.cc
```diff
@@ -124,6 +124,7 @@ bool GzCollisionDetector::BatchRaycast(
 GzOdeCollisionDetector::GzOdeCollisionDetector()
   : OdeCollisionDetector(), GzCollisionDetector()
 {
+
```
azeey: nit: Remove unneccesary diff.

dartsim/src/GzCollisionDetector.cc
```diff
@@ -170,6 +186,168 @@ bool GzOdeCollisionDetector::collide(
...
+bool GzOdeCollisionDetector::BatchRaycast(
+      CollisionGroup *_group,
+      const std::vector<GzRay> &_rays,
+      std::vector<GzRayResult> &_results) const
+{
+  auto odeGroup = static_cast<GzOdeCollisionGroup *>(_group);
+  const dSpaceID spaceId = odeGroup->getOdeSpaceId();
+
+  const dGeomID rayId = dCreateRay(spaceId, 1.0);
```
azeey: Gemini suggests that it's possible to create the ray outside of the existing space (in the null space) using `dCreateRay(0, 1.0)`. This would ensoure we're not mutating the existing space in a `const` function. I've tested it locally and it seems to work fine.
```suggestion
  const dGeomID rayId = dCreateRay(0, 1.0);
```

dartsim/src/GzCollisionDetector.cc
```diff
@@ -170,6 +186,168 @@ bool GzOdeCollisionDetector::collide(
...
+  {
+      gzwarn << "raycast multihit support is not"
+             << " implemented for ODE" << std::endl;
+      return false;
+  }
+
+  auto odeGroup = static_cast<GzOdeCollisionGroup *>(group);
+  const dSpaceID spaceId = odeGroup->getOdeSpaceId();
+
+  const dGeomID rayId = dCreateRay(spaceId, 1.0);
```
azeey:
```suggestion
  const dGeomID rayId = dCreateRay(0, 1.0);
```

dartsim/src/GzCollisionDetector.cc
```diff
@@ -170,6 +186,168 @@ bool GzOdeCollisionDetector::collide(
...
+  if(dCollide(ray, other, 1, &contact, sizeof(dContactGeom)) > 0)
+  {
+    if(result->mRayHits.empty())
+    {
+      setResult(result->mRayHits.emplace_back());
+    }
+    else
+    {
+      setResult(result->mRayHits.front());
+    }
```
azeey, line 259: From Gemini:
> `dSpaceCollide2` will call `NearCallbackODE` for every geometry whose bounding box intersects the ray. By unconditionally overwriting .front(), the raycast will simply return the last object visited by ODE's spatial traversal hierarchy, which is not guaranteed to be the closest object.

I have reproduced the issue by running this on gz-sim/examples/worlds/cpu_lidar_sensor.sdf with the `collision_detector` changed to ode. As you see in the video below, after moving the cube in front of the cylinder, there are still rays hitting the cylinder. Sometimes, it happens the other way around. I move the cube behind the cylinder and the rays now skip the cylinder and hit the cube.

https://github.com/user-attachments/assets/042965d4-c1ed-4ca0-8dd7-b6eb8234fe09


## Suggested solution:
> [!TIP]
> **Recommendation**: You need to keep track of the minimum distance and only overwrite if the new contact is closer. ODE conveniently populates `contact.depth` with the distance from the ray origin for ray intersections. You can pass a custom struct via `_data` instead of just `RaycastResult*`:
> 
> ```cpp
> struct OdeRaycastData {
>   RaycastResult* result;
>   double minDistance = std::numeric_limits<double>::infinity();
> };
> ```
> And in `NearCallbackODE`:
> ```cpp
>   OdeRaycastData* data = static_cast<OdeRaycastData*>(_data);
>   // ... inside dCollide block:
>   if (contact.depth < data->minDistance)
>   {
>     data->minDistance = contact.depth;
>     if(data->result->mRayHits.empty())
>       setResult(data->result->mRayHits.emplace_back());
>     else
>       setResult(data->result->mRayHits.front());
>   }
> ```

azeey (changes requested):
Overall, this looks good, but there a couple of issues that need to be addressed. It would also be great if we can update the test to use multiple objects in the line of sight of the ray.

dartsim/src/GzCollisionDetector.cc
jmachowinski: updated

dartsim/src/GzCollisionDetector.cc
jmachowinski: This one is odd, I would have expected to only get the closest hit as I set dGeomRaySetClosestHit.

I fixed it now, note I explicitly computed the sqr dist, as I tried using depth before for the fraction and it gave me not the distance.

jmachowinski:
@azeey ping

dartsim/src/GzCollisionDetector.cc
azeey: Looks like it's not quite fixed on my end. I believe you forgot to update `data->curContactDistSqr`.

Image

> This one is odd, I would have expected to only get the closest hit as I set dGeomRaySetClosestHit.

It is giving you the closest hit for each ray/object pair, but the `NearCallabackODE` will be called for each object in the scene. In other words, the ray may hit the box multiple times and you'll get the closest of those, but the collision of the ray with the cylinder will be treated as a separate collision and gets its own callback.

jmachowinski:
@azeey Added a test and fixed the issue.

test/common_test/simulation_features.cc
```diff
@@ -2347,6 +2347,52 @@ TYPED_TEST(SimulationFeaturesRayIntersectionTest, UnsupportedRayIntersections)
...
+          this->loader,
+          name,
+          common_test::worlds::kMultipleObjectsComplexSdf);
+      world->SetCollisionDetector(collisionDetector);
+      auto checkedOutput = StepWorld<FeaturesRayIntersections>(world, true, 1).first;
+      EXPECT_TRUE(checkedOutput);
+
+      // Test ray along X axis, should hit closest sphere at x=1
+      // Objects are at different Y positions to confuse spatial traversal
+      gzerr << "[TEST DEBUG] About to call GetRayIntersectionFromLastStep" << std::endl;
```
azeey: Remove

test/common_test/simulation_features.cc
```diff
@@ -2347,6 +2347,52 @@ TYPED_TEST(SimulationFeaturesRayIntersectionTest, UnsupportedRayIntersections)
...
+      auto result =
+        world->GetRayIntersectionFromLastStep(
+            Eigen::Vector3d(-2, 0, 0), Eigen::Vector3d(10, 0, 0));
+      gzerr << "[TEST DEBUG] GetRayIntersectionFromLastStep returned" << std::endl;
```
azeey: Remove debug prints

test/common_test/simulation_features.cc
```diff
@@ -2347,6 +2347,52 @@ TYPED_TEST(SimulationFeaturesRayIntersectionTest, UnsupportedRayIntersections)
   }
 }
 
+/////////////////////////////////////////////////
+TYPED_TEST(SimulationFeaturesRayIntersectionTest, ClosestRayIntersection)
+{
+  std::vector<std::string> supportedCollisionDetectors = {"ode"};
```
azeey: Any reason for not includding `bullet` here?

azeey (approved):
LGTM! Thanks again for the contribution. Just a few minor comments left. Also, would you mind merging from main and fixing the conflicts?

test/common_test/simulation_features.cc
jmachowinski: done

test/common_test/simulation_features.cc
jmachowinski: done

test/common_test/simulation_features.cc
jmachowinski: added bullet

azeey (approved):

PR #1004, opened 2026-06-11
mujoco: Implement joint velocity commands

azeey:
# 🎉 New feature

Toward #299 

Requires #1028 for all tests to pass. Please review that first.

## Summary
This PR implements joint velocity commands for the MuJoCo physics engine via native velocity actuators (`mjBIAS_AFFINE`).

To make velocity tracking configuration-independent and uniform, we dynamically compute the servo gain at each step as $k_v = J / \tau_s$, where $J$ is the un-factored joint inertia from `d->qM[m->dof_Madr[dofIndex]]` and $\tau_s = 0.05 \cdot \text{timestep}$ (5% of the timestep). This implements a proportional velocity controller using MuJoCo's native velocity actuators (`mjBIAS_AFFINE`).

~Additionally, instead of using joint-level effort limits (`actfrcrange`), which are solved as soft constraints in MuJoCo and introduce compliance and phase lag, we set force limits directly on the actuator (`forcerange`) using the SDF effort limits. This enforces limits as a hard input saturation clip, which prevents solver compliance and keeps the control loop stable.~

### ⚠️ Note on Joint Velocity Commands and Effort Limits (MuJoCo vs. DART)

When using joint velocity commands (`SetVelocityCommand`), imposing effort (force/torque) limits on joints impacts tracking accuracy across physics engines, but the minimum effort limit required for accurate tracking differs significantly between **MuJoCo** and **DART**.

#### Behavior Comparison:

* **DART (LCP Constraint Solver):** Enforces velocity commands via hard algebraic constraints ($M \Delta \dot{q} = J^T \lambda$) within its constraint solver. Because the solver accounts for the complete system dynamics (full mass matrix, Coriolis, and gravity) to determine the exact required impulse, DART can maintain accurate velocity tracking even with tighter effort limits. The video below shows a Scara robot being simulated in DART. The `JointPositionController` plugin is used with `use_velocity_commands=true`

https://github.com/user-attachments/assets/c607d2f2-fb3e-4076-b123-d2268e5d0dbd

*File: [scara_velocity_commands.sdf](https://gist.github.com/azeey/d58d0368c79eaddea05631ae2fa5fc7c)*

* **MuJoCo (Decoupled Velocity Servo):** Implements velocity commands via a proportional velocity servo (`mjBIAS_AFFINE`) scaled by the diagonal apparent joint inertia ($k_v = J_{\text{diag}} / \tau$). Because this decoupled servo relies on high transient gains to overcome dynamic coupling and gravity rather than solving for exact global forces, **MuJoCo requires a substantially larger effort limit** than DART to achieve accurate tracking on the same model. If effort limits are set to a moderate value that works fine in DART, MuJoCo's servo torque gets clamped prematurely, leading to noticeable tracking lag and sagging. The videos below show the Scara robot simulated in MuJoCo with and without effort limits. The controller is unable to track the commanded velocities accurately in the former.

https://github.com/user-attachments/assets/5f46e0ad-cab6-4151-9126-7f13dec1b272

*File: [scara_velocity_commands.sdf](https://gist.github.com/azeey/d58d0368c79eaddea05631ae2fa5fc7c)*

https://github.com/user-attachments/assets/61393d40-a2db-4b6a-bced-7ada3afe8549

*File: [scara_velocity_commands_no_effort_limits.sdf](https://gist.github.com/azeey/64c0222c28d135e6241d2187be4ad9ae)*

#### Recommendation / Workaround:
When migrating or running velocity-controlled models under the MuJoCo physics plugin, ensure that joint effort limits are set to a sufficiently large value (or removed) to prevent premature clamping of the velocity servo.

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Generated-by: Gemini 3.5 Flash

### DIFF
--- a/mujoco/src/JointFeatures.cc
+++ b/mujoco/src/JointFeatures.cc
@@ -19,6 +19,7 @@
 #include <cmath>
 #include <cstddef>
 #include <memory>
+#include <numeric>
 #include <string_view>
 #include <utility>
 #include <vector>
@@ -1061,9 +1062,9 @@ Identity JointFeatures::AttachFixedJoint(
     worldInfo->specDirty = true;
     if (worldInfo->mjModelObj)
     {
-      int b1 = mj_name2id(worldInfo->mjModelObj, mjOBJ_BODY, childBodyName);
-      int b2 = mj_name2id(worldInfo->mjModelObj, mjOBJ_BODY, parentBodyName);
-      if (b1 > 0 && b2 > 0 && b1 < worldInfo->mjModelObj->nbody &&
+      int b1 = mjs_getId(childLinkInfo->body->element);
+      int b2 = parentLinkInfo ? mjs_getId(parentLinkInfo->body->element) : 0;
+      if (b1 >= 0 && b2 >= 0 && b1 < worldInfo->mjModelObj->nbody &&
           b2 < worldInfo->mjModelObj->nbody)
       {
         int w1 = worldInfo->mjModelObj->body_weldid[b1];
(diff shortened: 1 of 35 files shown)


mujoco/src/JointFeatures.cc
```diff
@@ -300,8 +327,24 @@ double JointFeatures::GetJointForce(
...
+  auto *m = jointInfo->worldInfo->mjModelObj;
+  auto *d = jointInfo->worldInfo->mjDataObj;
+  if (jointInfo->actuator)
+  {
+    const int actuatorId =
+        mjs_getId(jointInfo->actuator->element) + static_cast<int>(_dof);
+    if (actuatorId >= 0 && m->actuator_biastype[actuatorId] == mjBIAS_NONE &&
+        d->ctrl[actuatorId] != 0.0)
+    {
+      return d->ctrl[actuatorId];
```
caguero, line 343: I see in this branch that we return `d->ctrl` but in the general case we return `d->qfrc_actuator`. Could this be an issue if we do a sequence like: `SetJointForce(5); Step(); SetJointForce(0); GetJointForce()`. Will this get `0`?

mujoco/src/JointFeatures.cc
```diff
@@ -862,11 +955,11 @@ Identity JointFeatures::AttachFixedJoint(
...
   }
-  if (modelInfo->joints.HasEntity(_name))
+  std::string uniqueName = _name;
+  std::size_t counter = 1;
+  while (modelInfo->joints.HasEntity(uniqueName))
   {
-    gzerr << "There's already a joint with the same name [ " << _name
-          << " ].\n";
-    return this->GenerateInvalidId();
+    uniqueName = _name + "_" + std::to_string(counter++);
```
caguero, line 962: My understanding is that this code now enables multiple detachable joints with the same default name, but at the same time we allow attaching the same pair twice. Should we avoid that or at least show a warning?

caguero (commented):
Nice! I left some comments.

mujoco/src/JointFeatures.cc
azeey: The issue is that the `AttachedFixedJoint` API has a name parameter which has a default value of `fixed`. Virtually all the uses of the API here and in gz-sim omit that parameter with the expectation that the joint will be created. So strictly enforcing uniqueness here will break all those use cases. Also note that both the dartsim and bullet-featherstone implementations support duplicated joint names.

mujoco/src/JointFeatures.cc
azeey: Good call. I added explicit tracking of whether joint force has been commanded. 2a9a785a75e7 shows a failing test based on the sequence you suggested. 5178c2dea794 adds the implementation.

caguero (approved):

PR #1005, opened 2026-06-15
Remove hash map allocation on every update step

luca-della-vedova:

# 🎉 New feature

Part of https://github.com/gazebosim/gz-sim/issues/3608
Improve performance by removing allocation + deallocation of a hash map for all link poses in each update step.

## Summary
Guided by the [profiling report](https://caguero.github.io/gz-profiling/2026-04-21/findings_report.pdf) by @caguero, Priority 2.
Building and destroying a hash map, where each node needs to call malloc / free, is an expensive operation.
I tried the quick recommendation to reserve the hash map but found out it didn't help much and I thought the "best" approach of not sending updates for static links might be a bit tricky (and still, it is not mutually exclusive, we can still do it in a followup).
Thus the approach is to remove the extra map and make sure we use the preexisting `prevLinkPoses` hash map instead.

Before there was an

```
declare empty map
if (pose changed)  -> add new pose to map
else -> add old pose to map
```

By caching the pose through iterations, we can get rid of the else statement (since it will be just set to the value in the previous iteration).
Finding out which links have been removed is a bit trickier. I thought since removing links is something that happens fairly rarely, we could do a fast check where, if at any point in time the size of our cache is larger than the size of our links map, we trigger a full traversal and delete stale links.
I believe this should be robust even to cases where we add and remove a link in the same iteration. Adding a link with a new ID will make the cache allocate a new element and thus it will become larger than the links array, triggering the reallocation.
At the same time, this approach will do virtually no work at all iterations, instead of having to do expensive lookups / traversal at every step for, again, something I believe happens very rarely (link deletion).

## Backport Policy
- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure
## Test it

Run gz-sim with and without this PR:

```
$ gz sim -v3 3k_shapes.sdf -r --physics-engine gz-physics-dartsim-plugin -s
$ gz sim -v3 3k_shapes.sdf -r --physics-engine gz-physics-bullet-featherstone-plugin -s
```

On my machine I actually don't observe a noticeable improvement under `dartsim` since it's already very slow (~1% RTF, I'm open to revert the commit that makes this change there https://github.com/gazebosim/gz-physics/pull/1005/commits/1a78c624df2c8b382e37b62ff64067b385d550de), but on bullet-featherstone the RTF goes from ~10-20% to 20-30%, so about 75% increase.

## Checklist
- [ ] Added tests

### DIFF
--- a/bullet-featherstone/src/SimulationFeatures.cc
+++ b/bullet-featherstone/src/SimulationFeatures.cc
@@ -370,8 +370,6 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
   _changedPoses.entries.clear();
   _changedPoses.entries.reserve(this->links.size());
 
-  std::unordered_map<std::size_t, math::Pose3d> newPoses;
-
   for (const auto &[id, info] : this->links)
   {
     const auto &model = this->ReferenceInterface<ModelInfo>(info->model);
@@ -385,16 +383,27 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
         !iter->second.Rot().Equal(wp.pose.Rot(), 1e-6))
     {
       _changedPoses.entries.push_back(wp);
-      newPoses[id] = wp.pose;
+      this->prevLinkPoses[id] = wp.pose;
     }
-    else
-      newPoses[id] = iter->second;
   }
 
-  // Save the new poses so that they can be used to check for updates in the
-  // next iteration. Re-setting this->prevLinkPoses with the contents of
-  // newPoses ensures that we aren't caching data for links that were removed
-  this->prevLinkPoses = std::move(newPoses);
+  if (this->prevLinkPoses.size() > this->links.size())
+  {
+    // Clear all the removed links
+    auto it = this->prevLinkPoses.begin();
+    while (it != this->prevLinkPoses.end())
+    {
+      if (this->links.find(prevPose.first) == this->links.end())
+      {
+        // Link was removed, clear from cache
+        it = this->prevLinkPoses.erase(it);
+      }
+      else
+      {
+        ++it;
+      }
+    }
+  }
 }
 }  // namespace bullet_featherstone
 }  // namespace physics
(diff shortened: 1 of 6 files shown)


iche033:
Did some quick plots of RTF. The top shows RTF when sim is initially launched with the 3k shapes. The bottom plot shows RTF after sim is running for a while and it went up presumably because bullet deactivation kicks in. Blue line is from `main` and red line is with the changes in this branch. I see a good improvement in RTF after the objects go to sleep (bottom plot). It looks like there's is some small overhead when objects are not put to sleep yet? The difference is small though

Screenshot 2026-06-18 at 3 15 16 PM

cheriehu:
These are the results I get on mac 
#### Bullet-featherstone 
- post-sleep: 140% before PR, 160% after PR
- pre-sleep: about 3% both with and without PR

#### Dart 
- 3.1% both with and without the PR

bullet-featherstone/src/SimulationFeatures.cc
```diff
@@ -385,16 +383,27 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
...
     }
-    else
-      newPoses[id] = iter->second;
   }
 
-  // Save the new poses so that they can be used to check for updates in the
-  // next iteration. Re-setting this->prevLinkPoses with the contents of
-  // newPoses ensures that we aren't caching data for links that were removed
-  this->prevLinkPoses = std::move(newPoses);
+  if (this->prevLinkPoses.size() > this->links.size())
```
azeey: Alternatively, we can handle this in the `RemoveModelImpl` functions. gz-physics only has APIs for removing models, not links by themselves.

luca-della-vedova: Interesting, that would require the data to be cached in a variable thas is accessible in `Base.hh` since that's where this function is implemented.
So I acknowledge and raise with https://github.com/gazebosim/gz-physics/pull/1005/commits/1e31498952044e5d71ce56a31680ba09bb65ece0 that actually makes this PR a negative LOC diff. If we just store the cached data in a variable in `LinkInfo` the cache will be automatically deleted once the LinkInfo goes out of scope. We can also get rid of a hash map lookup in the hot loop improving performance further (in my machine this adds a 5% or so RTF). The only cost is that the variable is now defined in Base.hh which is a bit of a leak since it's only necessary in a submodule.
What do you think?

azeey: Nice! I think this is a better outcome.

azeey (approved):

luca-della-vedova:
@mergify backport gz-physics7 gz-physics8 gz-physics9

PR #1006, opened 2026-06-16
mujoco: Support setting surface friction params to collisions

iche033:
# 🎉 New feature

Toward https://github.com/gazebosim/gz-physics/issues/299

## Summary

Set collision surface friction values from SDF. Note that SDF has engine-specific friction sdf elements for sliding and rolling friction. So I had to add some logic to parse them, i.e. use mu from `<ode>` if it exists otherwise, use the one from `<bullet>`. 

Currently if no friction params are specified, they won't be set and the plugin should continue to use the [default values from mujoco](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-geom-friction). The alternative is to use the [default values from SDF spec](https://sdformat.org/spec/1.12/collision/#surface_friction) (which has quite different torsional and rolling friction values from the defaults in mujoco).
 
## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure
- [x] Other (fill in yourself)
    * The whole mujoco plugin should be backported together.

## Checklist
- [x] Added tests

Generated-by: Gemini 3


mujoco/src/SDFFeatures.cc
```diff
@@ -693,12 +698,58 @@ struct ModelKinematicStructure
...
+            // Parse torsional friction value
+            if (auto torsionalElem = frictionElem->FindElement("torsional"))
+            {
+              if (auto coeffElem = torsionalElem->FindElement("coefficient"))
+                spinningFriction = coeffElem->Get<double>();
+            }
+          }
+        }
+
+       uint16_t contype = contypeOpt.value_or(conaffinity);
```
caguero: Nit: Add extra space.

mujoco/src/SDFFeatures.cc
```diff
@@ -693,12 +698,58 @@ struct ModelKinematicStructure
...
         geom->contype = static_cast<int>(contype);
         geom->conaffinity = static_cast<int>(conaffinity);
+
+        if (mu.has_value())
+          geom->friction[0] = mu.value();
+        if (spinningFriction.has_value())
+          geom->friction[1] = spinningFriction.value();
```
caguero, line 967: Apparently, Mujoco needs the contact dimmensionality parameter `condim` `>=4` to generate torsional friction and `== 6` to generate rolling friction. See [docs](https://mujoco.readthedocs.io/en/stable/computation/index.html#condim).

I think we should set `geom->condim = std::max(geom->condim, 4);` in this case.

mujoco/src/SDFFeatures.cc
```diff
@@ -693,12 +698,58 @@ struct ModelKinematicStructure
...
+        if (rollingFriction.has_value())
+          geom->friction[2] = rollingFriction.value();
```
caguero, line 973: And `geom->condim = 6;` in this case.

caguero (commented):
Looks good to me, I just left a few comments.

mujoco/src/SDFFeatures.cc
iche033: [21108c3](https://github.com/gazebosim/gz-physics/pull/1006/commits/21108c36249e94a47b08c90598a6827084064938)

mujoco/src/SDFFeatures.cc
iche033: oh good catch, updated in [21108c3](https://github.com/gazebosim/gz-physics/pull/1006/commits/21108c36249e94a47b08c90598a6827084064938)

mujoco/src/SDFFeatures.cc
iche033: done. [21108c3](https://github.com/gazebosim/gz-physics/pull/1006/commits/21108c36249e94a47b08c90598a6827084064938)

caguero:
~~CI seems stuck~~. It looks happier now.

caguero (approved):

caguero:
@iche033 There are some conflicts here.

iche033:
> @iche033 There are some conflicts here.

fixed conflicts in [d71a02b](https://github.com/gazebosim/gz-physics/pull/1006/commits/d71a02b545dc466e99da21e564fd9e78511da335)

azeey (commented):
We recently received a comment from a user (a power user) that they were surprised to find that dartsim was using the friction parameter from the `<ode>` tag instead of the `<dart>` tag. At the very least, we should print a message that mujoco is using these parameters. We could also use the `gz` xml namespace to define friction parameters for mujoco:
```xml
<gz:mujoco>
  <friction>1 0.005 0.0001</friction>
</gz:mujoco>
```

It would be nice in general to pass mujoco parameters from SDF files, but I haven't thought about the best way.

iche033:
> At the very least, we should print a message that mujoco is using these parameters.

I added a few msgs in [d96fcb1](https://github.com/gazebosim/gz-physics/pull/1006/commits/d96fcb1959e0473c1112e97bdf9ff5c729ae628b)

> It would be nice in general to pass mujoco parameters from SDF files, but I haven't thought about the best way.

yes agreed. I think we should discuss the best way to do this. I'll hold of on adding the `<gz:mujoco>` flag until we have a consensus on the best way forward.

iche033:
The [JointSetCommand](https://github.com/gazebosim/gz-physics/blob/b6dbf1276046edae0fe34cf8aadf3bfaa2274902/test/common_test/joint_features.cc#L171) test started failing after merging with `main` (because it is now run in mujoco). Turns out that having a high `mu` (100 from `<ode>`) in mujoco causes contacts to be very soft and objects start penetrating and sinking into the ground plane. 

@azeey, looks like this pushes the need for mujoco specific friction parameters. I capped it to `1` for the test to pass but open to adding the `gz:mujoco` tag. Any preference?

azeey:
pinging @scpeters since this subject keeps coming up.

PR #1008, opened 2026-06-19
Publish small DART pose deltas

jslee02:
# Bug fix

Fixes the root cause behind https://github.com/gazebosim/gz-sim/issues/3697.

## Summary

The DART engine plugin was treating link pose deltas smaller than 1e-6 as unchanged in ChangedWorldPoses. gz-sim diff-drive command tests can legitimately produce first-step motion below that threshold, so downstream pose consumers miss a real update.

This changes ChangedWorldPoses to publish every real pose delta and adds a DART-plugin regression that moves a free group by 5e-7 m and verifies the changed pose is emitted.

## Testing

- On the matching gz-physics8 backport branch, configured against source-built DART 6.19 and built COMMON_TEST_simulation_features plus gz-physics8-dartsim-plugin.
- `ctest --test-dir build/dart619 --output-on-failure --parallel 1 -R '^(COMMON_TEST_simulation_features_dartsim|check_COMMON_TEST_simulation_features_dartsim)$'`
- DART downstream validation: `DART_PARALLEL_JOBS=8 pixi run -e gazebo test-gz` passed, including gz-sim INTEGRATION_entity_system against the source-built DART plugin.

## Backport Policy

- [x] This is safe to backport to supported gz-physics branches used by affected Gazebo distributions.
- [ ] This should not be backported
- [ ] I am not sure

## Related Issues / PRs

- https://github.com/gazebosim/gz-sim/issues/3697
- DART validation PR: https://github.com/dartsim/dart/pull/3068

### DIFF
--- a/dartsim/src/SimulationFeatures.cc
+++ b/dartsim/src/SimulationFeatures.cc
@@ -185,8 +185,8 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
       // add it to the output poses. Otherwise, keep the existing link pose
       auto iter = this->prevLinkPoses.find(id);
       if ((iter == this->prevLinkPoses.end()) ||
-          !iter->second.Pos().Equal(wp.pose.Pos(), 1e-6) ||
-          !iter->second.Rot().Equal(wp.pose.Rot(), 1e-6))
+          !iter->second.Pos().Equal(wp.pose.Pos(), 0.0) ||
+          !iter->second.Rot().Equal(wp.pose.Rot(), 0.0))
       {
         _changedPoses.entries.push_back(wp);
         newPoses[id] = wp.pose;
--- a/test/common_test/simulation_features.cc
+++ b/test/common_test/simulation_features.cc
@@ -1284,6 +1284,53 @@ TYPED_TEST(SimulationFeaturesTestFreeGroup, FreeGroup)
   }
 }
 
+TYPED_TEST(SimulationFeaturesTestFreeGroup, ChangedWorldPosesPublishesSmallDelta)
+{
+  for (const std::string &name : this->pluginNames)
+  {
+    if (name.find("dartsim") == std::string::npos)
+      continue;
+
+    auto world = LoadPluginAndWorld<FreeGroupFeatures>(
+      this->loader,
+      name,
+      common_test::worlds::kSphereSdf);
+
+    auto model = world->GetModel("sphere");
+    ASSERT_NE(nullptr, model);
+    auto freeGroup = model->FindFreeGroup();
+    ASSERT_NE(nullptr, freeGroup);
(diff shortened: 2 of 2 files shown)


dartsim/src/SimulationFeatures.cc
```diff
@@ -185,8 +185,8 @@ void SimulationFeatures::Write(ChangedWorldPoses &_changedPoses) const
       // add it to the output poses. Otherwise, keep the existing link pose
       auto iter = this->prevLinkPoses.find(id);
       if ((iter == this->prevLinkPoses.end()) ||
-          !iter->second.Pos().Equal(wp.pose.Pos(), 1e-6) ||
-          !iter->second.Rot().Equal(wp.pose.Rot(), 1e-6))
+          !iter->second.Pos().Equal(wp.pose.Pos(), 0.0) ||
```
azeey, line 188: This code was originally added to improve performance, especially on large scenes like the 3k_shapes.sdf, so we'll need to evaluate this change on its performance impact. @luca-della-vedova can you take a look?

luca-della-vedova:
Thanks for the ping @azeey!
Indeed I ran this on my machine, with the standard dartsim distributed in Ubuntu 24.04 as part of a gazebo installation.
I noticed a noticeable RTF drop on 3k_shapes, from about 1.4% to 1.15% so about 17% performance drop.
I captured a flamegraph before and after the PR.

Before:

graph_main

After:

graph

What's interesting is that zooming into `gz::sim::v11::systems::PhysicsPrivate::Update`, you can see how `PhysicsPrivate::UpdateSim` now takes 11% of the app time while before the PR it took about 2.5% of the time. The vast majority of the time there is spent updating model's pose so indeed this impacts performance.
I wonder, if I understand correctly there is an upcoming "autosleep" feature in dartsim, would that feature return deltas that are exactly equal to 0 or would there still be some variance / noise that would make them go through this if statement?
The reason I ask is if going forward we can expect a lot of pose diffs to be _exactly_ 0 we could consider having such a change in versions of Gazebo that will use dartsim with autosleep and not versions that don't have it. Not sure how much of a good idea it is but it could help at least performance wise.

cheriehu:
> I noticed a noticeable RTF drop on 3k_shapes, from about 1.4% to 1.15% so about 17% performance drop.

I tried running a modified version of the benchmark in https://github.com/gazebosim/gz-sim/pull/3671 (with dartsim instead of bullet-featherstone) and got these results

before PR (on main)
```
---------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations
---------------------------------------------------------------------------------------
BM_RuntimeWorld/lengthy_dart_3k_shapes_sdf/1       24.4 ms         24.4 ms           29
```
after PR
```
BM_RuntimeWorld/lengthy_dart_3k_shapes_sdf/1       28.5 ms         28.4 ms           24
```

PR #1032, opened 2026-08-06
Use custom comparator and wrapper around typeid() instead of type_index

cheriehu:
# 🦟 Bug fix

## Summary

The issue was surfaced [here](https://github.com/gazebosim/gz-physics/pull/1019#issuecomment-5172729649). Copying the description of the bug and the proposed fix here. 

This PR was largely generated with help from Gemini. 

> While the performance concern that motivated #1019 was completely valid (allocating a new `std::string` for every contact property during high-frequency physics steps is very expensive), `std::type_index` is fundamentally unsafe to use across dynamically loaded shared libraries (DSOs) on macOS and often on Linux if plugins are loaded with `RTLD_LOCAL`. 
> 
> The macOS Clang compiler (and `libc++`) does not guarantee that `std::type_info` instances for implicitly instantiated templates or inline structs are coalesced across library boundaries. As a result, when the test binary queries `std::type_index(typeid(ExtraContactDataT))`, it gets a different pointer than the one the `mujoco` plugin registered it with, and the `find()` fails.
> 
> **A better solution for #1019:**
> We can get the best of both worlds (zero heap allocations + safe cross-plugin type matching) by reverting `std::type_index` and instead using a lightweight wrapper around `typeid(Data).name()` with a custom comparator. 
> 
> Because `typeid(Data).name()` is guaranteed to return a static `const char*` that lives for the lifetime of the program, we can simply store the pointer and use `std::strcmp` for map lookups:
> 
> ```cpp
> struct TypeName {
>     const char* name;
>     bool operator<(const TypeName& other) const {
>         return std::strcmp(name, other.name) < 0;
>     }
> };
> 
> // Inside CompositeData:
> using MapOfData = std::map<TypeName, DataEntry>;
> 
> // Usage:
> this->dataMap.insert(std::make_pair(TypeName{typeid(Data).name()}, DataEntry()));
> ```
> 
> This completely eliminates the `std::string` allocations that #1019 was trying to solve, but correctly uses string comparison to match types safely across macOS plugin boundaries. 

## Testing
1. Checkout 045e5d2d352cc155a0f15c83827aee42016378e4, verify it fails (only on macOS, on Ubuntu it seems to pass)
2. Checkout 054890aa535791b476e8c83b8cda33058a8e58ac or 6ba44e07ea45a02b26d881a90c57bf8da2515bb3 and verify the test passes

## Performance
Performance seems to be on par with the `type_index` implementation
Ran `./build/gz-sim/bin/BENCHMARK_server_run --benchmark_filter="BM_RuntimeWorldContacts/lengthy"`
```
on 1032
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       2.96 ms         2.96 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_dart/1/iterations:1000            29.3 ms         29.2 ms         1000

main
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       2.95 ms         2.95 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_dart/1/iterations:1000            28.8 ms         28.8 ms         1000

main + revert 1019
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       3.65 ms         3.64 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_dart/1/iterations:1000            29.5 ms         29.4 ms         1000
```

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [X] This **should not** be backported
- [ ] I am not sure

## Checklist
- [ ] Added tests
- [X] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Gemini 3.1 Pro

### DIFF
--- a/include/gz/physics/CompositeData.hh
+++ b/include/gz/physics/CompositeData.hh
@@ -21,7 +21,6 @@
 #include <string>
 #include <map>
 #include <set>
-#include <typeindex>
 
 #include <gz/utils/SuppressWarning.hh>
 
@@ -964,7 +963,7 @@ namespace gz
 
       // We make this typedef public so that helper functions can use it without
       // being friends of the class.
-      public: using MapOfData = std::map<std::type_index, DataEntry>;
+      public: using MapOfData = std::map<std::string, DataEntry>;
 
       GZ_UTILS_WARN_IGNORE__DLL_INTERFACE_MISSING
       /// \brief Map from the label of a data object type to its entry
--- a/include/gz/physics/CompositeData.hh
+++ b/include/gz/physics/CompositeData.hh
@@ -972,14 +972,10 @@ namespace gz
         const char *name = "";
         bool operator<(const TypeName &_other) const
         {
-          if (this->name == _other.name)
-            return false;
           return std::strcmp(this->name, _other.name) < 0;
         }
         bool operator==(const TypeName &_other) const
         {
-          if (this->name == _other.name)
-            return true;
           return std::strcmp(this->name, _other.name) == 0;
         }
         bool operator!=(const TypeName &_other) const
--- a/include/gz/physics/CompositeData.hh
+++ b/include/gz/physics/CompositeData.hh
@@ -18,9 +18,11 @@
 #ifndef GZ_PHYSICS_COMPOSITEDATA_HH_
 #define GZ_PHYSICS_COMPOSITEDATA_HH_
 
+#include <cstring>
 #include <string>
 #include <map>
 #include <set>
+#include <typeinfo>
 
 #include <gz/utils/SuppressWarning.hh>
 
@@ -961,9 +963,34 @@ namespace gz
       /// \private
       public: struct DataEntry;
 
+      /// \brief Lightweight wrapper around typeid(Data).name() with a custom
+      /// comparator to match types safely across dynamically loaded shared
+      /// library boundaries without heap allocations.
+      /// \private
+      public: struct TypeName
+      {
+        const char *name = "";
+        bool operator<(const TypeName &_other) const
+        {
+          if (this->name == _other.name)
+            return false;
+          return std::strcmp(this->name, _other.name) < 0;
+        }
+        bool operator==(const TypeName &_other) const
+        {
+          if (this->name == _other.name)
+            return true;
+          return std::strcmp(this->name, _other.name) == 0;
+        }
+        bool operator!=(const TypeName &_other) const
+        {
+          return !(*this == _other);
+        }
+      };
+
       // We make this typedef public so that helper functions can use it without
       // being friends of the class.
-      public: using MapOfData = std::map<std::string, DataEntry>;
+      public: using MapOfData = std::map<TypeName, DataEntry>;
 
       GZ_UTILS_WARN_IGNORE__DLL_INTERFACE_MISSING
       /// \brief Map from the label of a data object type to its entry
--- a/include/gz/physics/CompositeData.hh
+++ b/include/gz/physics/CompositeData.hh
@@ -967,21 +967,23 @@ namespace gz
       /// comparator to match types safely across dynamically loaded shared
       /// library boundaries without heap allocations.
       /// \private
-      public: struct TypeName
+      public: class TypeName
       {
-        const char *name = "";
-        bool operator<(const TypeName &_other) const
+        public: TypeName(const std::type_info& _info) : _name(_info.name()) {}
+        public: const char *name() const { return this->_name; }
+        public: bool operator<(const TypeName &_other) const
         {
-          return std::strcmp(this->name, _other.name) < 0;
+          return std::strcmp(this->_name, _other._name) < 0;
         }
-        bool operator==(const TypeName &_other) const
+        public: bool operator==(const TypeName &_other) const
         {
-          return std::strcmp(this->name, _other.name) == 0;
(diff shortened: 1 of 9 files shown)


include/gz/physics/CompositeData.hh
```diff
@@ -962,9 +963,34 @@ namespace gz
...
+      /// comparator to match types safely across dynamically loaded shared
+      /// library boundaries without heap allocations.
+      /// \private
+      public: struct TypeName
+      {
+        const char *name = "";
+        bool operator<(const TypeName &_other) const
+        {
+          if (this->name == _other.name)
+            return false;
```
luca-della-vedova: I _think_ I understand what this (and the one below) are doing and I'm not sure the optimization is worth the risk.
Since we are comparing `const char *`, these are actually pointer comparison functions, not string comparison functions, so they can offer a quick equality check that doesn't require traversing the whole string.
However, as unlikely as it is, I wonder if it could happen that two different libraries are compiled where the pointer value is the same but what it points to is different, hence these short circuits operations return a false positive / negative. It sounds like a similar symptom to what prompted this PR in the first place (using pointers instead of names)

luca-della-vedova (commented):
Minor feedback, style seems to fail, might require a force push if you want to keep the test in a separate commit.

I've mentioned privately that the issue and test is out of my knowledge and it introduces non trivial changes to the `CMakeLists` so I'm slightly leaning towards not committing it and just committing the change, but that would mean not having a regression test so I am happy to be convinced otherwise.

include/gz/physics/CompositeData.hh
cheriehu: ah ok removed in 614bf1eec11fbcf591023df5b3a454d14cc834ef

cheriehu:
> I've mentioned privately that the issue and test is out of my knowledge and it introduces non trivial changes to the `CMakeLists` so I'm slightly leaning towards not committing it and just committing the change, but that would mean not having a regression test so I am happy to be convinced otherwise.

After our discussions last week, we decided it might be a good idea to keep the test but maybe simplify it. I've tried to do that in [22912c2](https://github.com/gazebosim/gz-physics/pull/1032/commits/22912c2734951ff7b5ae5831461a3958af7700ac).

test/plugins/CMakeLists.txt
```diff
@@ -5,11 +5,16 @@ endif()
...
+set_target_properties(MockCompositeData PROPERTIES
+  CXX_VISIBILITY_PRESET hidden
+  VISIBILITY_INLINES_HIDDEN 1)
 
 list(APPEND plugins
     MockEntities
     MockFrames
-    MockJoints)
+    MockJoints
+    MockCompositeData)
```
scpeters: nit: alphabetize this list by adding `MockCompositeData` above `MockEntities`

src/CompositeData.cc
```diff
@@ -337,7 +336,7 @@ namespace gz
       for (const auto &entry : dataMap)
       {
         if (entry.second.data)
-          entries.insert(entries.end(), entry.first.name());
```
luca-della-vedova, line 340: I wonder if we should have a const getter function for `name` so we can both keep this API and make `name` immutable after the struct has been constructed (unless there is a case of changing it, but I suspect there shouldn't be?)

include/gz/physics/detail/SpecifyData.hh
```diff
@@ -35,7 +36,7 @@ namespace gz
       : CompositeData(),
         privateExpectData(
           this->dataMap.insert(
-            std::make_pair(std::type_index(typeid(Expected)),
```
luca-della-vedova, line 38: Nit, since we always construct `CompositeData::TypeName` with a `name` string parameter that is derived from `typeid`, how about instead making it a class, keeping the name private and adding a constructor that takes the `type_info` as a parameter, so we have strong typing, more "safety" with the name being immutable, and less code repetition, i.e:

```
// New constructor
TypeName(const std::type_info& _info) : name(_info.name()) {}

// Now construction goes from
CompositeData::TypeName{typeId(Expected).name()}
// To
CompositeData::TypeName(typeId(Expected))
```

luca-della-vedova (commented):
Some minor suggestions to the API that could be cleaned up a bit, the simplified test LGTM

include/gz/physics/detail/SpecifyData.hh
cheriehu: thanks for the suggestion, i've implemented this in f302e75

test/plugins/CMakeLists.txt
cheriehu: thanks for the note, done in [af0d3a6](https://github.com/gazebosim/gz-physics/pull/1032/commits/af0d3a624b0c6d88f36288f64913f5ce44cfc23d)

luca-della-vedova:
@osrf-jenkins run tests

luca-della-vedova (approved):
LGTM, I would propose trying to get this in sooner rather than later to fix the homebrew regression

PR #1029, opened 2026-07-29
Bump vendor Mujoco to latest 3.11.0

j-rivero:
# 🎉 New feature

Bump Mujoco version to use the latest 3.11.0

## Summary

The PR bumps Mujoco to the 3.11.0 release, agreed with @azeey in the last PMC. The required changes for building it are low, a couple of fixes detailed below in section A. I instructed the agent to detect in the Changelog the relevant aspects that have changed and affects our gz-physics plugin in the behaviour and there a good bunch of them that require an expert review, section B. 

#### A.1 `mjsJoint.stiffness` / `mjsJoint.damping` became polynomial arrays — [**3.7.0**](https://mujoco.readthedocs.io/en/stable/changelog.html#version-3-3-7-october-13-2025)

> *"The `mjs` layer fields `stiffness` and `damping` in `mjsJoint` and `mjsTendon` have been widened
> from `mjtNum` scalars to `mjtNum[mjNPOLY+1]` arrays. Element 0 is the linear coefficient."*

This is the single most invasive change for the plugin. `mjNPOLY` is 2, so both fields are now `double[3]`.

**Sites:** `mujoco/src/SDFFeatures.cc:140,143` (`copyStandardJointAxisProperties`) and four assertions in `mujoco/src/SDFFeatures_TEST.cc`.

**Fix applied:** write/read index `[0]`. I think that SDFormat's `<damping>` and `<spring_stiffness>` are linear coefficients, so element 0 is the correct target and the higher-order terms keep their zero defaults. The IA agent claims: "Behaviour is identical to 3.5.0."

#### A.2 `mjtByte`/`int` fields became strongly-typed enums — [**3.9.0**](https://mujoco.readthedocs.io/en/stable/changelog.html#version-3-9-0-may-27-2026)

> *"Added `mjtBool` to represent boolean variables, replacing `mjtByte` across all boolean fields."*

Alongside `mjtBool`, several `int` spec fields were retyped to their real enums. `mjsJoint.limited` and `mjsJoint.actfrclimited` are now `mjtLimited`, and C++ rejects implicit `int → enum`.

**Sites:** `SDFFeatures.cc:144`, `:150`, `:543`, `:556`.

[...]

PR #1033, opened 2026-08-07
mujoco: Narrow the post-step kinematics sync

j-rivero:
# 🦟 Bug fix

This is an experimental change that requires expert validation before we consider the merge of it.

While bumping Mujoco to 3.11.0 #1029 and trying to use the new parallel collision feature, one my agents complained how much performance the gz-physics harness was putting on the computation.  Asking the agent to do a performance investigation on the harness applied to Mujoco, ended up in this PR.

Trying to explain it in simple words I got this summary: 

## Summary

After each simulation step the plugin needed to update where things are, and it was calling a function that updates where things are and re-runs half the physics engine — all of which the next step immediately overwrote. Replacing that call with the three small pieces the plugin actually reads made stepping 23–28 % faster

So the plugin is paying for collision detection, for building and factorizing the mass matrix, for assembling the entire set of contact constraints, for tendons and motors and drag and gravity-compensation terms. Then it read three of the arrays and returned. Milliseconds later, mj_step ran again and recomputed every one of those things from scratch, because by then the world had moved.

## Technical description

After `mj_step` we could be doing too much work by calling the methods `mj_fwdPosition(m, d)` and `mj_fwdVelocity(m, d)` since they perform much more actions that what we need to "Synchronize Cartesian position and velocity kinematics for the new state". Our syncs comes from the fact of in "MuJoCo, the numerical integrator in mj_step (Stage 24) advances the joint space variables (qpos/qvel) to the new state, but does not recompute the corresponding Cartesian kinematic and frame variables (e.g. xpos, xipos, site_xpos, cvel). These are normally computed lazily at the start of the next step"

This is coming directly from the AI agent findings, take it with a grain of salt I can not validate it easily although the explanation sounds reasonable to me.:

""""
mj_step advances qpos/qvel but leaves the Cartesian arrays at the previous state, so WorldForwardStep re-synchronises them before any downstream query. It did so with mj_fwdPosition followed by mj_fwdVelocity, which is far more work than the sync needs. Replace them with the three stages that actually produce what the plugin reads: mj_kinematics for the frame arrays, mj_comPos for subtree_com and cdof, and mj_comVel for cvel, which mj_objectVelocity needs in Link::FrameDataRelativeToWorld.

Everything else those two functions do -- collision detection, building and factorising the mass matrix, constructing the constraint set and islands, transmission, passive forces, RNE -- is discarded and recomputed by the next mj_step, so it was pure waste.
""""

### Benchmarking

As a result of the change, the collision parallelization improved by reducing the overhead of gz-physics from 42% to 5%.  Numbers produced in a single world and in the parallel feature:

"""
Stepping 400 free boxes on a 16-thread machine, serial World::Step drops from 0.949s to 0.730s (-23%). Measured against a raw mj_step loop in the same run, the plugin's per-step overhead falls from about 42% to about 5%.
"""

### Side effect / Tradeoff

I think that this is interesting since contacts solving has changed. AI agent also pointed that we could be solving a hidden problem/issue. In simple worlds: 

""
When two objects touch, MuJoCo does two separate things: it detects the contact, and then it solves for the force needed to stop the objects passing through each other. Gazebo asks for both together: here is a contact, and here is how hard it pushed. 

The current extra call ran the detection half again, on the newer positions, but never the solving half — solving happens later in the step, and the plugin was not doing a step. So on the exact step where a contact first came into existence, the plugin overwrote the solved list with a freshly detected one that had no forces attached to it yet.

The change also fixes contact force reporting. mj_fwdPosition rebuilds the efc arrays through mj_makeConstraint, which allocates them on the arena and does not solve them, so GetContactsFromLastStep was pairing contacts with whatever the arena happened to hold. 

The observable consequence is that d->contact now describes the pose the solver used, at the start of the step, rather than the pose reached after integration. Contact sets therefore change one step later than before in both directions: later on touchdown, and later on separation.
"""

**This seems important** to my eyes: "contacts are now noticed one step later than before. And the contact list now belongs to the positions the solver worked from, which are the positions at the start of the step, not the ones being reported alongside it. That is the same convention the forces already followed, so the two now agree with each other instead of being one step apart.

## Testing and validation

Given the tradeoff , I was expecting the test suite to fail in a way that we can detect. It did not. 

Instructed the agent to maximize the validation it went through 39 SDF worlds, 400 steps, and claim that "every link pose, orientation, linear and angular velocity and every joint position and velocity is byte-identical; 17 worlds differ, and only in contact records". So the contact change seems to be the behaviour change here.

Assisted-by: claude-opus-5

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Assisted-by: Opus 5.0

### DIFF
--- a/mujoco/src/SimulationFeatures.cc
+++ b/mujoco/src/SimulationFeatures.cc
@@ -66,12 +66,32 @@ void SimulationFeatures::WorldForwardStep(const Identity &_worldID,
   // In MuJoCo, the numerical integrator in mj_step (Stage 24) advances the
   // joint space variables (qpos/qvel) to the new state, but does not recompute
   // the corresponding Cartesian kinematic and frame variables (e.g. xpos,
-  // xipos, site_xpos, cvel, xvel). These are normally computed lazily at the
-  // start of the next step. Synchronizing them here ensures that immediate
-  // downstream state queries (like Link::FrameDataRelativeToWorld) return
-  // accurate, lag-free results for the current timestep.
-  mj_fwdPosition(m, d);
-  mj_fwdVelocity(m, d);
+  // xipos, site_xpos, cvel). These are normally computed lazily at the start of
+  // the next step. Synchronizing them here ensures that immediate downstream
+  // state queries (like Link::FrameDataRelativeToWorld) return accurate,
+  // lag-free results for the current timestep.
+  //
+  // Only these three stages are needed, and calling them directly rather than
+  // mj_fwdPosition/mj_fwdVelocity matters for two reasons:
+  //
+  // 1. Cost. mj_fwdPosition also runs collision detection, builds and
+  //    factorizes the mass matrix, and constructs the constraint set; all of
+  //    that is discarded and recomputed by the next mj_step, so it is pure
+  //    waste. Skipping it removes most of the plugin's per-step overhead.
+  //
+  // 2. Contact consistency. mj_fwdPosition rebuilds the efc arrays through
+  //    mj_makeConstraint, which allocates them on the arena without solving
+  //    them, so GetContactsFromLastStep ended up pairing contacts with
+  //    whatever the arena happened to hold. That surfaces as reported normal
+  //    forces that are negative, which a solved contact can never be.
+  //    Leaving d->contact as mj_step solved it keeps every reported contact
+  //    paired with the force that produced it. The trade-off is that contacts
+  //    now describe the pose the solver used, at the start of the step,
+  //    rather than the pose reached after integration, so contact sets change
+  //    one step later than before both on touchdown and on separation.
+  mj_kinematics(m, d);  // xpos, xquat, xipos, ximat, geom_xpos, site_xpos/xmat
+  mj_comPos(m, d);      // subtree_com, cdof, cinert
+  mj_comVel(m, d);      // cvel, needed by mj_objectVelocity
 
   // Clear joint control forces so that they are not applied in the next
   // timestep, which is the expected behavior in Gazebo.


azeey (commented):
I have dug around and found that this is the right thing to do. The one step offset of contact observation is actually more consistent with what we do with dartsim and the API `GetContactsFromLastStep` is aptly named to indicate that the contacts are from the last step.

There are other places in the code where we call `mj_forward` when it's not necessary. With the help of Gemini, I've identified the following:

1. `mujoco/src/JointFeatures.cc`
- In `JointFeatures::SetJointPosition`:
  - We need to do forward kinematics and update velocities. `mj_comPos` is needed for `mj_comVel`, so replace `mj_forward(m, d)` with:
    ```cpp
    mj_kinematics(m, d);
    mj_comPos(m, d);
    mj_comVel(m, d);
    ```
 
- In `JointFeatures::SetJointVelocity`:
  - Only `qvel` changes, so no need to do forward kinematics or compute the COM poses, so replace `mj_forward(m, d)` with:
    ```cpp
    mj_comVel(m, d);
    ```

 2. `mujoco/src/FreeGroupFeatures.cc`
- In `FreeGroupFeatures::SetFreeGroupWorldPose`:
  - We're setting the pose of the root link of the kinematic tree, so we'll need forward kinematics and updated velociited. So, replace `mj_forward(m, d)` with:
    ```cpp
    mj_kinematics(m, d);
    mj_comPos(m, d);
    mj_comVel(m, d);
    ```

- In `FreeGroupFeatures::SetFreeGroupWorldLinearVelocity` & `SetFreeGroupWorldAngularVelocity`:
  - We're updating the velocity of the root link, but since the pose doesn't change, we only need to update velocities. So, replace `mj_forward(m, d)` with:
    ```cpp
    mj_comVel(m, d);
    ```

j-rivero:
> I have dug around and found that this is the right thing to do. The one step offset of contact observation is actually more consistent with what we do with dartsim and the API `GetContactsFromLastStep` is aptly named to indicate that the contacts are from the last step.

Great news. I have changed the status of Draft to an active PR. Thanks for taking the time to check this,  I think it is promising for the Mujoco plugin in terms of performance and opens the door to use own Mujoco physics parallelization.

> There are other places in the code where we call `mj_forward` when it's not necessary. With the help of Gemini, I've identified the following:

good catch, I think I modified the places you found in [eafddd6](https://github.com/gazebosim/gz-physics/pull/1033/commits/eafddd6731d35be0ed9a9544bf7c2ae6f2404de7)

j-rivero:
Asked my agent to rebuild the benchmarking for this PR after the last review and patch:

  - **base**: fffd0105 (`main` before this PR)
  - **PR head**: 8310a574 (post-step sync narrowed)
  - **review**: eafddd67 (setters narrowed as suggested)

  ### Setup

  - 400 free 1 m boxes (20×20 grid) resting on a plane, ~1600 active contacts, 1 ms step, vendored MuJoCo 3.11.0, Release build.
  - Serial `World::Step`, pinned to one core of a Ryzen 7 5800X.
  - 200 warm-up steps + 1000 timed steps per run, 5 runs per build with builds interleaved. Medians are reported; run-to-run spread was 1–2%.
  - **raw** is a plain `mj_step` loop in the same process, on the exact model compiled by the plugin (dumped with `mj_saveXML`).

  ### 1. Stepping only (the benchmark in the PR description)

  | Build | ms/step | vs base | raw `mj_step` ms | Plugin overhead over raw |
  |---|---:|---:|---:|---:|
  | base | 2.249 | — | 1.586 | +41.8% |
  | PR head | 1.645 | −26.9% | 1.587 | +3.6% |
  | review | 1.638 | −27.2% | 1.596 | +3.4% |

  This confirms the numbers in the description (overhead 42% → ~4%). The review changes don't touch the step path, so as expected the review build matches PR head within noise.

  ### 2. Stepping with a setter call before every step

  To measure the setters changed in the review, one of the 400 models receives a command before each `World::Step`:

  - **velocity command**: `SetWorldLinearVelocity` + `SetWorldAngularVelocity`
  - **pose command**: `SetWorldPose`

  | Build | velocity command ms/step | vs base | pose command ms/step | vs base |
  |---|---:|---:|---:|---:|
  | base | 4.987 | — | 3.728 | — |
  | PR head | 4.462 | −10.5% | 3.126 | −16.2% |
  | review | **1.634** | **−67.2%** | **1.693** | **−54.6%** |

Each of these setters used to run a full `mj_forward`, about 1.4–1.5 ms per call in this world, which is almost the cost of a whole step. On PR head, commanding the velocity of a single model made each step 2.7× more expensive than stepping alone. With the review changes the velocity command has no measurable cost, and the pose command adds ~0.0

mujoco/src/SimulationFeatures.cc
```diff
@@ -68,12 +68,32 @@ void SimulationFeatures::WorldForwardStep(const Identity &_worldID,
...
+  //
+  // 1. Cost. mj_fwdPosition also runs collision detection, builds and
+  //    factorizes the mass matrix, and constructs the constraint set; all of
+  //    that is discarded and recomputed by the next mj_step, so it is pure
+  //    waste. Skipping it removes most of the plugin's per-step overhead.
+  //
+  // 2. Contact consistency. mj_fwdPosition rebuilds the efc arrays through
+  //    mj_makeConstraint, which allocates them on the arena without solving
+  //    them, so GetContactsFromLastStep ended up pairing contacts with
+  //    whatever the arena happened to hold. That surfaces as reported normal
```
azeey: Since readers will not have the historical context, consider rephrasing this as "`GetContactsFromLastStep" would end up pairing ..."

mujoco/src/SimulationFeatures.cc
```diff
@@ -68,12 +68,32 @@ void SimulationFeatures::WorldForwardStep(const Identity &_worldID,
...
+  //    forces that are negative, which a solved contact can never be.
+  //    Leaving d->contact as mj_step solved it keeps every reported contact
+  //    paired with the force that produced it. The trade-off is that contacts
+  //    now describe the pose the solver used, at the start of the step,
+  //    rather than the pose reached after integration, so contact sets change
+  //    one step later than before both on touchdown and on separation.
```
azeey: The trade-off makes this sound like a problem. This is the same behavior as dart, so our users expect it.

azeey (approved):
LGTM! Just a couple of minor comments

PR #1059, opened 2026-09-09
mujoco:Implement `GetJointTransmittedWrench`

azeey:
# 🎉 New feature

## Summary
Implement `GetJointTransmittedWrenchInJointFrame` for the MuJoCo physics engine plugin.

This feature uses native MuJoCo force and torque sensors (`mjSENS_FORCE` and `mjSENS_TORQUE`) attached to each joint's site in `mjSpec` (skipping free joints). This enables MuJoCo's internal simulation pipeline to automatically evaluate and project the reaction wrench into the joint frame natively, avoiding custom spatial algebra calculations or previous-step snapshot buffers.

## Backport Policy
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Test it
Run the joint transmitted wrench common tests:
```bash
./bin/COMMON_TEST_joint_transmitted_wrench_features ./lib/libgz-physics-mujoco-plugin.dylib
```
All 5 tests in `JointTransmittedWrenchFixture` pass.

## Checklist
- [x] Added tests

Assisted-by: Gemini 3.8 Flash

### DIFF
--- a/mujoco/src/JointFeatures.cc
+++ b/mujoco/src/JointFeatures.cc
@@ -1122,6 +1122,8 @@ Wrench3d JointFeatures::GetJointTransmittedWrenchInJointFrame(
     {
       gzerr << "MuJoCo currently does not support wrench queries on detachable "
                "joints\n";
+      // TODO(azeey): Use constraint forces exposed by MuJoCo to compute
+      // wrenches on weld constraints.
     }
     else
     {
--- a/mujoco/src/JointFeatures.cc
+++ b/mujoco/src/JointFeatures.cc
@@ -1109,11 +1109,25 @@ Wrench3d JointFeatures::GetJointTransmittedWrenchInJointFrame(
   auto *worldInfo = jointInfo->worldInfo;
   if (worldInfo->specDirty)
   {
-    this->RecompileSpec(*worldInfo);
+    if (!this->RecompileSpec(*worldInfo))
+    {
+      gzerr << "Failed to recompile spec.\n";
+      return {};
+    }
   }
 
   if (!jointInfo->forceSensorId || !jointInfo->torqueSensorId)
   {
+    if (jointInfo->weldEqIndex)
+    {
+      gzerr << "MuJoCo currently does not support wrench queries on detachable "
+               "joints\n";
+    }
+    else
+    {
+      gzerr << "Unkown error: Force and torque sensor IDs have not been "
+               "resolved\n";
+    }
     return {};
   }
 
--- a/mujoco/src/JointFeatures.cc
+++ b/mujoco/src/JointFeatures.cc
@@ -1095,6 +1095,39 @@ Identity JointFeatures::AttachFixedJoint(
   return this->GenerateIdentity(jointInfo->entityId, jointInfo);
 }
 
+/////////////////////////////////////////////////
+Wrench3d JointFeatures::GetJointTransmittedWrenchInJointFrame(
+    const Identity &_id) const
+{
+  auto jointInfo = this->ReferenceInterface<JointInfo>(_id);
+  if (!jointInfo || !jointInfo->worldInfo)
+  {
+    gzerr << "Invalid joint info when querying transmitted wrench.\n";
+    return {};
+  }
+
+  auto *worldInfo = jointInfo->worldInfo;
+  if (worldInfo->specDirty)
+  {
+    this->RecompileSpec(*worldInfo);
+  }
+
+  if (!jointInfo->forceSensorId || !jointInfo->torqueSensorId)
+  {
+    return {};
+  }
+
+  const auto *m = worldInfo->mjModelObj;
+  const auto *d = worldInfo->mjDataObj;
+  const int fAdr = m->sensor_adr[*jointInfo->forceSensorId];
+  const int tAdr = m->sensor_adr[*jointInfo->torqueSensorId];
+
+  Wrench3d wrench;
+  wrench.force = convertPos(&d->sensordata[fAdr]);
+  wrench.torque = convertPos(&d->sensordata[tAdr]);
+  return wrench;
+}
+
 namespace {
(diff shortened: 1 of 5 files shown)


mujoco/src/JointFeatures.cc
```diff
@@ -1095,6 +1095,39 @@ Identity JointFeatures::AttachFixedJoint(
...
+    return {};
+  }
+
+  auto *worldInfo = jointInfo->worldInfo;
+  if (worldInfo->specDirty)
+  {
+    this->RecompileSpec(*worldInfo);
+  }
+
+  if (!jointInfo->forceSensorId || !jointInfo->torqueSensorId)
```
caguero, line 1115: Joints created by `AttachFixedJoint` don't get these sensors, so this would return a zero wrench for them. I think DART reports a real wrench in this case. Could we add the sensors in `AttachFixedJoint` too, or alternatively add a warning here?

mujoco/src/SDFFeatures.cc
```diff
@@ -678,11 +678,26 @@ struct ModelKinematicStructure
...
 
+      auto addSiteSensor = [&](mjtSensor _type, const std::string &_suffix) {
+        auto *sensor = mjs_addSensor(_spec);
+        sensor->type = _type;
+        sensor->objtype = mjOBJ_SITE;
+        mjs_setString(sensor->objname, jointSiteName.c_str());
+        mjs_setName(sensor->element, (scopedJointName + _suffix).c_str());
+        return sensor;
+      };
+      jointInfo->forceSensorSpec = addSiteSensor(mjSENS_FORCE, "_force");
```
caguero, line 698: These sensors make MuJoCo run `mj_rnePostConstraint` every step for every joint, even if nobody asks for the wrench. Gazebo only queries joints with a ForceTorque sensor. Would it make sense to add the sensors lazily on the first query? The getter already recompiles the spec when it's dirty.

mujoco/src/JointFeatures.cc
```diff
@@ -1095,6 +1095,39 @@ Identity JointFeatures::AttachFixedJoint(
...
+  if (!jointInfo || !jointInfo->worldInfo)
+  {
+    gzerr << "Invalid joint info when querying transmitted wrench.\n";
+    return {};
+  }
+
+  auto *worldInfo = jointInfo->worldInfo;
+  if (worldInfo->specDirty)
+  {
+    this->RecompileSpec(*worldInfo);
```
caguero, line 1112: Should we check the return value here?

mujoco/src/SDFFeatures.cc
```diff
@@ -678,11 +678,26 @@ struct ModelKinematicStructure
...
+      const std::string scopedJointName =
+          ::sdf::JoinName(_modelInfo->name, sdfJoint->Name());
+      const std::string jointSiteName = scopedJointName + "_site";
+      mjs_setName(jointSite->element, jointSiteName.c_str());
       copyPos(jointPose.Pos(), jointSite->pos);
       copyQuat(jointPose.Rot(), jointSite->quat);
       _base.frames[jointInfo->entityId] =
           std::make_shared<FrameInfo>(jointSite, worldInfo);
 
+      auto addSiteSensor = [&](mjtSensor _type, const std::string &_suffix) {
```
caguero, line 690: Nit: Move `{` to the next line

mujoco/src/SDFFeatures.cc
azeey: I had several versions of this code that would call `mj_rnePostConstraint` only if there's a request to get the wrench. But in all situations, the first call would return an invalid or 0 wrench and the code was much more complicated. In the end, I opted for the simplest approach where we use the MuJoCo's own sensor. With your suggestion, this would be the sequence of events in Gazebo

- System creates the `JointTransmittedWrench` component on an joint entity
- Step
  - Mujoco `step`
  - GetJointTransmittedWrench
    - No sensor, so add a new one and recompile
    - Return 0 or invalid wrench because the sensor was not ther when `step` was called. The sensor values will only be valid after the next `step`.

The other thing to consider is that `mj_rnePostConstraint`computes the forces and torques for all joints in the world, not just the ones that have the sensors, so there's no difference between having just once force/torque sensor and having many (when only considering `mj_rnePostConstraint`). Also, even though the most common use case in Gazebo is with a ForceTorque sensor, it is possible to query the wrench without a sensor by just creating the `JointTransmittedWrench` component on a joint entity. See how it's used in [gz_ros2_control](https://github.com/ros-controls/gz_ros2_control/blob/9ca4c190c77a09be60518d4b87ff957984e2b33d/gz_ros2_control/src/gz_system.cpp#L359-L365) for example without a ForceTorque sensor.

mujoco/src/JointFeatures.cc
azeey: The current implementation for detachable joints uses weld constraints, which don't support force/torque sensors. MuJoCo exposes the constraint forces and we can report that, but I haven't seen many use cases of this and would rather print an error for now. https://github.com/gazebosim/gz-physics/pull/1059/commits/d75bc6d254bdacee6c02d07f4ab69d19e79b63a0)adds the error and https://github.com/gazebosim/gz-physics/pull/1059/commits/d720277ec123e0e1596bc0d0829112e183514684 adds a TODO

mujoco/src/JointFeatures.cc
azeey: Done in https://github.com/gazebosim/gz-physics/pull/1059/commits/d75bc6d254bdacee6c02d07f4ab69d19e79b63a0

mujoco/src/SDFFeatures.cc
azeey: Done in https://github.com/gazebosim/gz-physics/pull/1059/commits/d75bc6d254bdacee6c02d07f4ab69d19e79b63a0

PR #1091, opened 2026-09-15
Fix dangling references when removing models in tpe plugin

### LINKED ISSUE #1090: TPE parent removal skips alternating nested models
## Environment

* OS Version: Ubuntu 24.04.5 LTS, x86_64
* Source or binary build? Source build of `main` at `b6dbf1276046edae0fe34cf8aadf3bfaa2274902`, tested with Bazel 9.1.1
* Physics plugin: `gz::physics::tpeplugin::Plugin`

This also reproduces on tag `gz-physics9_9.5.1` (`d1611113878d7fcba08212b0a03509f5b2d966ac`).

## Description

* Expected behavior: removing a parent model removes all nested models and every retained child proxy subsequently reports `Removed() == true`.
* Actual behavior: with six immediate nested models, parent removal returns true but children 1, 3, and 5 still report that they have not been removed.

Reproduction files: 

[tpe-nested-model-removal-poc.zip](https://github.com/user-attachments/files/32242780/tpe-nested-model-removal-poc.zip)

## Steps to reproduce

1. Clone `https://github.com/gazebosim/gz-physics.git` and check out the tested commit.
2. Copy the accompanying `poc` directory into the clean repository root as `tpe-nested-model-removal-repro`.
3. Run:

   ~~~bash
   git clone https://github.com/gazebosim/gz-physics.git
   cd gz-physics
   git checkout b6dbf1276046edae0fe34cf8aadf3bfaa2274902
   cp -R /absolute/path/to/tpe-nested-model-removal/poc ./tpe-nested-model-removal-repro
   bazelisk test //tpe-nested-model-removal-repro:repro --test_output=all
   ~~~

The reproducer creates one parent with six immediate nested models, retains every child proxy, removes the parent, and checks all children.

## Output

~~~text
parent->Remove() = true

child_0->Removed() = true
child_1->Removed() = false
child_2->Removed() = true
child_3->Removed() = false
child_4->Removed() = true
child_5->Removed() = false
~~~

## Additional context

`RemoveModelImpl` iterates the live child collection by increasing index and recursively removes each nested model:

[`Base.hh`](https://github.com/gazebosim/gz-physics/blob/b6dbf1276046edae0fe34cf8aadf3bfaa2274902/tpe/plugin/src/Base.hh#L176-L216)

Every recursive removal erases that child from the same collection:

[`Entity.cc`](https://github.com/gazebosim/gz-physics/blob/b6dbf1276046edae0fe34cf8aadf3bfaa2274902/tpe/lib/src/Entity.cc#L222-L232)

After erasure, the next child shifts into the just-consumed index, but the loop increments the index and skips it.

Copying the nested-model IDs before recursion, or repeatedly removing an appropriate live child until none remain, would avoid iterating a mutating indexed collection.

PR #230 introduced cascade removal and tests a single nested child, but does not cover multiple siblings.

j-rivero:
# 🦟 Bug fix

Fixes #1090 

## Summary

Removing a model from the TPE plugin left stale bookkeeping records behind, holding raw pointers into `tpelib` memory that is freed together with the model. Those dangling records were then dereferenced by the per-step pose loop and by the by-name `GetModel`/`GetLink`/`GetShape` lookups. This PR makes model removal clean up completely, so no record outlives the object it points at.

## Details

Two independent gaps in the removal path:

1. **Nested models were skipped.** `Base::RemoveModelImpl` walked a model's children with an increasing index (`GetChildByIndex(i)`) while the recursive removal erased children from that same container. After each erasure the successor shifted into the just-consumed slot, so every other immediate nested model was skipped. Its `ModelInfo` record stayed in `models`/`childIdToParentId` even though the object was destroyed with the parent subtree. For a parent with N nested models, the odd-indexed ones (`child_1`, `child_3`, `child_5`, ...) reported `Removed() == false` and left dangling records.

2. **Links and collisions were never removed.** `Base::RemoveModelFromParent` erased only the model's own `models`/`childIdToParentId` entries. The removed model's `links` and `collisions` records were never erased, so every model removal (nested or flat) left dangling `LinkInfo`/`CollisionInfo` entries pointing at freed `tpelib::Link`/`tpelib::Collision` objects.

Because `SimulationFeatures::Write` iterates `links`/`models` on every step, and `GetModel`/`GetLink`/`GetShape` iterate the whole map (calling `GetName()` on each entry before comparing ids), a stale record was read as freed memory — a use-after-free, confirmed under AddressSanitizer.

## Fix

- `RemoveModelImpl`: snapshot the immediate nested-model ids before removing any, then recurse over that snapshot. The index walk no longer races the container mutation, so no child is skipped.
- `RemoveModelFromParent`: before freeing the model's backend subtree, erase the model's `links` and each link's `collisions` records (and their `childIdToParentId` entries).
- `RemoveNestedModelByIndex`/`RemoveNestedModelByName`: route through `RemoveModelImpl` so they recurse into deeper nested models and clean up links/collisions as well (previously they did neither).

## Backport Policy
- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Assisted-by: Claude Opus 4.8


azeey (approved):

azeey:
@mergifyio backport gz-physics9 gz-physics8 gz-physics7 ign-physics5

I think it's okay to backport to Fortress as well

PR #1092, opened 2026-09-17
Mujoco: recompile the dirty spec before accessing the compiled model

caguero:
# 🦟 Bug fix

## Summary

The MuJoCo engine can crash when a controller drives a joint on a model that was just added to a running world. MuJoCo turns the world description into a compact internal form that holds a numbered slot for every moving part, and a joint is reached by its slot number. Gazebo builds a world one model at a time and defers rebuilding that internal form until the first simulation step. A command applied to a freshly added joint before that first step refers to a slot that does not exist yet, so it writes past the end of an array. That stray write can corrupt memory and take the process down.

The fix performs the pending rebuild before a value is read from or written to that form, so it uses the real slot instead of one that is not there yet. A safety check stays in place for the rare joint that genuinely has nothing to move.

## Backport Policy
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Assisted-by: Claude Fable 5.1


mujoco/src/JointFeatures.cc
```diff
@@ -499,6 +499,14 @@ void JointFeatures::SetJointForce(
     gzerr << "No actuator set up for this joint\n";
     return;
   }
+
+  // A joint whose body is fixed to the world (or is otherwise absent from the
+  // compiled model) has no degree of freedom, so nv_index is left unset (-1).
+  // There is no force to apply; guard as the other joint-state accessors do to
+  // avoid an out-of-bounds write into ctrl / jointForceCmdReceived.
+  if (jointInfo->nv_index < 0)
```
azeey, line 527: I think it's okay to keep this, but I don't know this is the underlying cause. The joint in the test added in the PR is trying to set the force on a revolute joint, which should work. I think the issue is that the mujoco model hasn't been recompiled after the model was created. The dirty flag is set, but we're not checking that in `SetJoint*` functions.

azeey (commented):
Can we do the same for SetJointPosition and SetJointVelocity and all the GetJoint* counterparts?

mujoco/src/JointFeatures.cc
caguero: You're right, thanks for the suggestion. Code updated.

caguero:
> Can we do the same for SetJointPosition and SetJointVelocity and all the GetJoint* counterparts?

Added there and also in link and free group setters. [d2d8f8a](https://github.com/gazebosim/gz-physics/pull/1092/commits/d2d8f8acf2cd4cc650642916c4e50942d7de866a)

azeey (approved):

PR #1098, opened 2026-09-18
mujoco: Store body pointer and constant offset in FrameInfo

azeey:
# 🦟 Bug fix

## Summary

In MuJoCo, each call to `mjs_add*` triggers a full-tree walk (`MakeTreeLists()`) and string serialization of the entire kinematic tree to recompute the spec signature (`Signature()`) (see https://github.com/google-deepmind/mujoco/issues/3397).

Previously, the MuJoCo plugin added four `mjSite` instances per single-link model (for joints, links, model frames, and collision shapes) solely so that `KinematicsFeatures::FrameDataRelativeToWorld` and joint frame query methods could read `d->site_xpos` and `d->site_xmat`. For worlds with thousands of bodies, this created quadratic element-addition overhead during world loading.

This PR eliminates all `mjSite` instances by storing the body pointer (`mjsBody *`) associated with the frame and a constant offset (`Eigen::Isometry3d`) directly in `FrameInfo`.

In `KinematicsFeatures::FrameDataRelativeToWorld`:
- `data.pose` is computed directly by transforming the frame offset with the body's world pose (mapped from `d->xpos` and `d->xmat`).
- 6D spatial velocities are computed directly using `mju_transformSpatial` with the frame's world translation, producing bit-equivalent results without requiring an `mjSite`.

### Benchmark

Loading `3k_shapes.sdf` (`gz sim -s -r --iterations 1`):
- **Before**: 7.72s user CPU / 10.18s wall
- **After**: **4.16s user CPU** / 6.62s wall (~46% reduction in CPU time)

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure
- [x] Backport with all MuJoCo changes

## Checklist
- [ ] Added tests

Assisted-by: Gemini 3.8 Flash


mujoco/src/KinematicsFeatures.cc
```diff
@@ -41,21 +41,52 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
+  }
+
+  const int bodyId = mjs_getId(frame.body->element);
+  if (bodyId < 0 || bodyId >= m->nbody)
+  {
+    gzerr << "Frame [" << _id.ID() << "] has an invalid body id\n";
+    return data;
+  }
+
+  // World pose of the owning body, then apply the constant frame offset.
```
caguero: My understanding is that there's a function doing something similar in `Base.hh` but slower than this approach using `xmat` instead of `xquat`. Does it make sense to either (1) replace the original implementation in `Base.hh` with this faster approach or; (2) create a new helper function in `Base.hh` with this faster approach to be used by other pieces of the code?

mujoco/src/Base.hh
```diff
@@ -280,12 +286,19 @@ struct ModelInfo
...
   {
   }
-  mjsSite * site{nullptr};
-  WorldInfo* worldInfo;
+
+  /// \brief Body this frame is rigidly attached to.
+  mjsBody *body{nullptr};
+
+  /// \brief Constant pose of this frame expressed in the body frame.
+  Eigen::Isometry3d offset{Eigen::Isometry3d::Identity()};
```
caguero: Should we make it `const` as the comment says?

mujoco/src/Base.hh
```diff
@@ -280,12 +286,19 @@ struct ModelInfo
...
+
+  WorldInfo *worldInfo{nullptr};
```
caguero, line 327: Nit: Missing doxygen.

mujoco/src/Base.hh
```diff
@@ -280,12 +286,19 @@ struct ModelInfo
 
 struct FrameInfo
 {
-  FrameInfo(mjsSite *_site, WorldInfo* _worldInfo)
-      : site(_site), worldInfo(_worldInfo)
+  FrameInfo(mjsBody *_body, const Eigen::Isometry3d &_offset,
```
caguero: Nit: Missing doxygen.

mujoco/src/Base.hh
```diff
@@ -128,6 +129,11 @@ inline Eigen::Isometry3d convertPose(const mjtNum *_pos, const mjtNum *_quat)
   return Eigen::Translation3d(convertPos(_pos)) * convertQuat(_quat);
 }
 
+inline Eigen::Isometry3d convertPose(const gz::math::Pose3d &_pose)
```
caguero, line 140: Nit: Missing doxygen.

mujoco/src/KinematicsFeatures.cc
```diff
@@ -41,21 +41,52 @@ FrameData3d KinematicsFeatures::FrameDataRelativeToWorld(
...
-  data.pose.linear() = rotMatMap;
+  auto *m = worldInfo->mjModelObj;
+  auto *d = worldInfo->mjDataObj;
+  const auto &frame = *it->second;
 
-  mjtNum velocity[6];
-  mj_objectVelocity(worldInfo->mjModelObj, d, mjOBJ_SITE, siteId, velocity, 0);
-  mju_copy3(data.angularVelocity.data(), velocity);
-  mju_copy3(data.linearVelocity.data(), velocity+3);
+  if (!frame.body)
```
caguero, line 49: Just checking, is this possible to happen?

azeey: Maybe a little to defensive, but it could happen if the FrameInfo constructor was handed a body pointer returned from mjs_addBody that failed.

mujoco/src/KinematicsFeatures.cc
azeey: Yup, makes sense. https://github.com/gazebosim/gz-physics/pull/1098/commits/f55ad8365c2ad59e18b2843112c1a736d36c3de5

mujoco/src/Base.hh
azeey: Done in https://github.com/gazebosim/gz-physics/pull/1098/commits/dcb455fb22e46a78cc1b9329db8fa7a04f486521

mujoco/src/Base.hh
azeey: https://github.com/gazebosim/gz-physics/pull/1098/commits/dcb455fb22e46a78cc1b9329db8fa7a04f486521

mujoco/src/Base.hh
azeey: https://github.com/gazebosim/gz-physics/pull/1098/commits/dcb455fb22e46a78cc1b9329db8fa7a04f486521 and https://github.com/gazebosim/gz-physics/pull/1098/commits/c5c2fb28355a828d1d475cac34dc4235fe000ea9

mujoco/src/Base.hh
azeey: https://github.com/gazebosim/gz-physics/pull/1098/commits/dcb455fb22e46a78cc1b9329db8fa7a04f486521

mujoco/src/JointFeatures.cc
```diff
@@ -676,8 +676,7 @@ Pose3d JointFeatures::GetJointTransformFromParent(const Identity &_id) const
   {
     Pose3d childInParent =
         convertPose(jointInfo->childBody->pos, jointInfo->childBody->quat);
-    Pose3d jointInChild =
-        convertPose(it->second->site->pos, it->second->site->quat);
+    Pose3d jointInChild = it->second->offset;
```
iche033, line 679: nit: avoid a copy or just use `it->second->offset` directly?
```suggestion
    const Pose3d &jointInChild = it->second->offset;
```

same for the other changes below

iche033 (approved):
very minor nit. Looks good.

caguero (approved):

